Issue #184, opened 2021-07-13
s3 custom treatment of GetBucketLocation response


wooj2:
https://github.com/awslabs/smithy-kotlin/issues/413

There is a fix in progress: https://github.com/awslabs/smithy/pull/839/files

*internal id*: 178574430

PR #385, opened 2021-09-01
feat: support S3UnwrappedXmlOutputTrait

wooj2:
## Issue \#
https://github.com/awslabs/aws-sdk-swift/issues/184

## Corresponding:
https://github.com/awslabs/smithy-swift/pull/335

## Description of changes
Adding support for S3UnwrappedXmlOutputTrait

## Scope

- [x] New feature (non-breaking change which adds functionality)

### DIFF
--- a/codegen/protocol-test-codegen/build.gradle.kts
+++ b/codegen/protocol-test-codegen/build.gradle.kts
@@ -32,12 +32,12 @@ val enabledProtocols = listOf(
     ProtocolTest("aws-restjson", "aws.protocoltests.restjson#RestJson", "AWSRestJsonTestSDK"),
     ProtocolTest("rest-xml", "aws.protocoltests.restxml#RestXml", "RestXmlTestSDK"),
     ProtocolTest("rest-xml-xmlns", "aws.protocoltests.restxml.xmlns#RestXmlWithNamespace", "RestXmlWithNamespaceTestSDK"),
-    ProtocolTest("aws-query", "aws.protocoltests.query#AwsQuery", "AWSQueryTestSDK")
+    ProtocolTest("aws-query", "aws.protocoltests.query#AwsQuery", "AWSQueryTestSDK"),
 
     // service specific tests
     //ProtocolTest("apigateway", "com.amazonaws.apigateway#BackplaneControlService"),
     //ProtocolTest("glacier", "com.amazonaws.glacier#Glacier", "GlacierTestSDK")
-    //ProtocolTest("s3", "com.amazonaws.s3#AmazonS3", "S3TestSDK")
+    ProtocolTest("s3", "com.amazonaws.s3#AmazonS3", "S3TestSDK")
 )
 
 // This project doesn't produce a JAR.
--- a/codegen/smithy-aws-swift-codegen/src/main/kotlin/software/amazon/smithy/aws/swift/codegen/restxml/RestXmlStructDecodeXMLGenerator.kt
+++ b/codegen/smithy-aws-swift-codegen/src/main/kotlin/software/amazon/smithy/aws/swift/codegen/restxml/RestXmlStructDecodeXMLGenerator.kt
@@ -1,0 +1,68 @@
+package software.amazon.smithy.aws.swift.codegen.restxml
+
+import software.amazon.smithy.aws.traits.customizations.S3UnwrappedXmlOutputTrait
+import software.amazon.smithy.model.shapes.BlobShape
+import software.amazon.smithy.model.shapes.CollectionShape
+import software.amazon.smithy.model.shapes.MapShape
+import software.amazon.smithy.model.shapes.MemberShape
+import software.amazon.smithy.model.shapes.OperationShape
+import software.amazon.smithy.model.shapes.TimestampShape
+import software.amazon.smithy.model.traits.TimestampFormatTrait
+import software.amazon.smithy.swift.codegen.SwiftTypes
+import software.amazon.smithy.swift.codegen.SwiftWriter
+import software.amazon.smithy.swift.codegen.integration.ProtocolGenerator
+import software.amazon.smithy.swift.codegen.integration.serde.xml.StructDecodeXMLGenerator
+import software.amazon.smithy.swift.codegen.model.ShapeMetadata
+import software.amazon.smithy.swift.codegen.model.hasTrait
+
+class RestXmlStructDecodeXMLGenerator(
+    val ctx: ProtocolGenerator.GenerationContext,
+    private val members: List<MemberShape>,
+    private val metadata: Map<ShapeMetadata, Any>,
+    private val writer: SwiftWriter,
+    defaultTimestampFormat: TimestampFormatTrait.Format
+) : StructDecodeXMLGenerator(ctx, members, metadata, writer, defaultTimestampFormat) {
+    override fun render() {
+        writer.openBlock("public init (from decoder: \$N) throws {", "}", SwiftTypes.Decoder) {
+            if (members.isNotEmpty()) {
+                renderDecodeBody()
+            }
+        }
+    }
+
+    private fun renderDecodeBody() {
+        val containerName = "containerValues"
+        if (responseBodyIsNotWrapped()) {
+            writer.write("var $containerName = try decoder.unkeyedContainer()")
+            members.forEach { member ->
+                renderSingleMemberUnkeyed(member, containerName)
+            }
+        } else {
+            writer.write("let $containerName = try decoder.container(keyedBy: CodingKeys.self)")
+            members.forEach { member ->
+                renderSingleMember(member, containerName)
+            }
+        }
+    }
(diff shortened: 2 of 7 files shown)


codegen/protocol-test-codegen/build.gradle.kts
```diff
@@ -32,12 +32,12 @@ val enabledProtocols = listOf(
...
     ProtocolTest("rest-xml", "aws.protocoltests.restxml#RestXml", "RestXmlTestSDK"),
     ProtocolTest("rest-xml-xmlns", "aws.protocoltests.restxml.xmlns#RestXmlWithNamespace", "RestXmlWithNamespaceTestSDK"),
-    ProtocolTest("aws-query", "aws.protocoltests.query#AwsQuery", "AWSQueryTestSDK")
+    ProtocolTest("aws-query", "aws.protocoltests.query#AwsQuery", "AWSQueryTestSDK"),
 
     // service specific tests
     //ProtocolTest("apigateway", "com.amazonaws.apigateway#BackplaneControlService"),
     //ProtocolTest("glacier", "com.amazonaws.glacier#Glacier", "GlacierTestSDK")
-    //ProtocolTest("s3", "com.amazonaws.s3#AmazonS3", "S3TestSDK")
+    ProtocolTest("s3", "com.amazonaws.s3#AmazonS3", "S3TestSDK")
```
kneekey23, line 40: how exciting!

codegen/smithy-aws-swift-codegen/src/main/kotlin/software/amazon/smithy/aws/swift/codegen/restxml/RestXmlStructDecodeXMLGenerator.kt
```diff
@@ -0,0 +1,68 @@
...
+            writer.write("let $containerName = try decoder.container(keyedBy: CodingKeys.self)")
+            members.forEach { member ->
+                renderSingleMember(member, containerName)
+            }
+        }
+    }
+    private fun responseBodyIsNotWrapped(): Boolean {
+        if (metadata.containsKey(ShapeMetadata.OPERATION_SHAPE)) {
+            val operationShape = metadata[ShapeMetadata.OPERATION_SHAPE] as OperationShape
+            if (operationShape.hasTrait<S3UnwrappedXmlOutputTrait>()) {
```
kneekey23, line 50: since this only applies to s3, should this be a customization to the decode xml generator or in here?

codegen/smithy-aws-swift-codegen/src/main/kotlin/software/amazon/smithy/aws/swift/codegen/restxml/RestXmlStructDecodeXMLGenerator.kt
```diff
@@ -0,0 +1,68 @@
...
+            }
+        }
+        return false
+    }
+
+    fun renderSingleMemberUnkeyed(member: MemberShape, containerName: String) {
+        val memberTarget = ctx.model.expectShape(member.target)
+        when (memberTarget) {
+            is CollectionShape, is MapShape, is TimestampShape, is BlobShape -> {
+                assert(false) { "Not implemented: renderSingleMemberUnkeyed - Collection, Map, Timestamp, Blob" }
```
kneekey23, line 61: can that trait even be applied to these types?

wooj2: The trait is applied to an operation... so technically -- yes, it can apply to any and all of these types of shapes -- that being said, I feel hesitant to update all of our logic with unkeyed members, if we don't don't have an example of how this works (e.g. protocol codegen tests to back it up)

codegen/smithy-aws-swift-codegen/src/main/kotlin/software/amazon/smithy/aws/swift/codegen/restxml/RestXmlStructDecodeXMLGenerator.kt
wooj2: Are you suggesting a customization via an integration?  If so, that's an interesting thought.. but didn't consider it yet because it doesn't seem like there is much risk in other restxml protocols using the `S3UnwrappedOutputTrait`

codegen/protocol-test-codegen/build.gradle.kts
wooj2: yes, exciting, but less exciting:
```
        val testsToIgnore = setOf(
            "S3DefaultAddressing",
            "S3VirtualHostAddressing",
            "S3PathAddressing",
            "S3VirtualHostDualstackAddressing",
            "S3VirtualHostAccelerateAddressing",
            "S3VirtualHostDualstackAccelerateAddressing",
            "S3OperationAddressingPreferred"
        )
```
:)

codegen/smithy-aws-swift-codegen/src/main/kotlin/software/amazon/smithy/aws/swift/codegen/restxml/RestXmlStructDecodeXMLGenerator.kt
kneekey23: That makes sense. Okay this seems like a reasonable approach for now until we have more info or evidence of other types in other services

codegen/smithy-aws-swift-codegen/src/main/kotlin/software/amazon/smithy/aws/swift/codegen/restxml/RestXmlStructDecodeXMLGenerator.kt
kneekey23: Ya it just seems to weird to call out an s3 specific trait here in a class where it's more general to all rest xml services. Even though it's not affecting them. Just feels a little off but I also trust your judgment if you wanted to leave it here

kneekey23 (approved):
Trusting your judgment in regards to customization vs leaving it here

codegen/smithy-aws-swift-codegen/src/main/kotlin/software/amazon/smithy/aws/swift/codegen/restxml/RestXmlStructDecodeXMLGenerator.kt
wooj2: Hmmm -- Let me try and implement it as a section writer -- if it is cleaner, then we'll go with that, but at least we can have something to compare it with.

codegen/protocol-test-codegen/build.gradle.kts
kneekey23: ya I saw this next haha but still it means we are making progress and chipping away at s3 customizations so still exciting

codegen/smithy-aws-swift-codegen/src/main/kotlin/software/amazon/smithy/aws/swift/codegen/restxml/RestXmlStructDecodeXMLGenerator.kt
wooj2: Looked into implementing this with an integration/section writer, and does not seem to be the right tool for the job.  Why? because section writers are really good at writing/generating code (not passing back a value) to the original context.  For example, if we wanted to put the following code in an integration only enabled for S3:
```
    private fun responseBodyIsNotWrapped(): Boolean {
        if (metadata.containsKey(ShapeMetadata.OPERATION_SHAPE)) {
            val operationShape = metadata[ShapeMetadata.OPERATION_SHAPE] as OperationShape
            if (operationShape.hasTrait<S3UnwrappedXmlOutputTrait>()) {
                return true
            }
        }
        return false
    }
```

Then a section writer is not the right tool for this -- we'd have to do something like modify the `SwiftIntegration` protocol in a way that we could call into this customization.

So, in that regard -- if we wanted to make `RestXmlStructDecodeXMLGenerator` generic, we could use the SwiftIntegration's `preprocessModel()` in order to detect whether or not an operation has `S3UnwrappedXmlOutputTrait` applied to it, and then replace it with a newly created trait like `UnwrappedXmlOutputTrait` -- so that way the code inside of `RestXmlStructDecodeXMLGenerator` becomes:
```
    private fun responseBodyIsNotWrapped(): Boolean {
        if (metadata.containsKey(ShapeMetadata.OPERATION_SHAPE)) {
            val operationShape = metadata[ShapeMetadata.OPERATION_SHAPE] as OperationShape
            if (operationShape.hasTrait<UnwrappedXmlOutputTrait>()) {
                return true
            }
        }
        return false
    }
```

This doesn't seem like a huge win, and seems to overcomplicate the code generation, so seems like it is reasonable just to leave it until we have a stronger case for changing it.

Issue #391, opened 2021-09-14
EC2, SSM and Kinesis do not compile


adam-fowler:
## Description
EC2, SSM and Kinesis all have compilation errors

## Environment

* AWS Swift SDK version used: v0.0.8
* Operating System and version (iOS/macOS/Linux): macOS
* Xcode version (if applicable): v12.5.1
* Swift compiler version (if applicable): Swift 5.4

PR #399, opened 2021-09-20
fix: compiler error in ec2

kneekey23:
## Issue \#

This fixes a compile issue in EC2. It was not compiling due to an incorrect type being placed in the error binding. Thus producing two compiler errors: 1 that doesnt conform to the protocol and 2. the init was expecting a diff type. This fixes 1/3 of #391 

## Conventional Commits


wooj2 (approved):
lgtm!

PR #400, opened 2021-09-20
fix: use correct content type

wooj2:
corresponding:
https://github.com/awslabs/smithy-swift/pull/348

Uses the correct content type instead of just the default one


kneekey23 (approved):
LGTM

PR #407, opened 2021-09-24
feat: Implement GET presigner for synthesizespeech

wooj2:
<DRAFT>

Heading in the right direction, but need to implement a piece of middleware which is able to take the input object and put it into the query string.

Corresponding:
https://github.com/awslabs/smithy-swift/pull/354


codegen/smithy-aws-swift-codegen/src/main/kotlin/software/amazon/smithy/aws/swift/codegen/customization/polly/PollyGetPresignerIntegration.kt
```diff
@@ -0,0 +1,164 @@
...
+import software.amazon.smithy.swift.codegen.middleware.OperationMiddleware
+import software.amazon.smithy.swift.codegen.model.expectShape
+
+internal val PRESIGNABLE_GET_OPERATIONS: Map<String, Set<String>> = mapOf(
+    "com.amazonaws.polly#Parrot_v1" to setOf(
+        "com.amazonaws.polly#SynthesizeSpeech"
+    )
+)
+
+class PollyGetPresignerIntegration(private val presignedOperations: Map<String, Set<String>> = PRESIGNABLE_GET_OPERATIONS) : SwiftIntegration {
```
kneekey23, line 36: does this override the previous integration which would have generated all these files somehow? don't we need order on this integration to ensure that?

kneekey23 (approved):
LGTM

Issue #388, opened 2021-09-09
Expose sigv4 signer


wooj2:
## Describe the Feature
Given a URL, method (potentially more inputs/config parameters pending investigation), expose a function in the AWSClientRuntime to allow users to use the Sigv4 signer, to generate a presigned request (or perhaps a presigned URL, pending design/discussion/investigation)

Additional requirements:
* Give option to double URI encode and additional, tunable configuration parameters

Essentially, we need a replacement and improvements on providing functionality for:
https://github.com/aws-amplify/aws-sdk-ios/blob/e5c21d5b8b042e55e51f8f184a4e06915648d285/AWSCore/Authentication/AWSSignature.m#L407

## Proposed Solution
* Pending investigation/examples

- [x] I may be able to implement this feature request

ameter:
Example use-cases...

As part of a fix to improve SigV4Signer support for non-s3 use cases in aws-sdk-ios, I wrote a couple of additional [test cases](https://github.com/aws-amplify/aws-sdk-ios/blob/fix/signatureV4Encoding/AWSCoreUnitTests/SigV4Tests/SigV4TestCases.swift) to validate the signature with special characters (single encoding the path before signing for s3 and double encoding for everything else).

Specifically, that last two test cases ("get-s3-specialchars" and "get-specialchars") address handling of special characters.

PR #414, opened 2021-09-30
feat: Expose sigv4 signer for SdkHttpRequestBuilder

wooj2:
No corresponding PRs

## Issue \#
https://github.com/awslabs/aws-sdk-swift/issues/388

## Description of changes
Need the ability to instrument the signer for various URLs

### DIFF
--- a/AWSClientRuntime/Sources/Signing/AWSSigV4Signer.swift
+++ b/AWSClientRuntime/Sources/Signing/AWSSigV4Signer.swift
@@ -1,0 +1,68 @@
+//
+// Copyright Amazon.com Inc. or its affiliates.
+// All Rights Reserved.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import ClientRuntime
+import AwsCommonRuntimeKit
+
+public class AWSSigV4Signer {
+    static let logger: SwiftLogger = SwiftLogger(label: "AWSSigV4Signer")
+    
+    public static func sigV4SignedURL(requestBuilder: SdkHttpRequestBuilder,
+                                      credentialsProvider: CredentialsProvider,
+                                      signingName: Swift.String,
+                                      signingRegion: Swift.String,
+                                      date: ClientRuntime.Date,
+                                      expiration: Int64) -> ClientRuntime.URL? {
+        do {
+            let credentials = try credentialsProvider.getCredentials()
+            
+            let flags = SigningFlags(useDoubleURIEncode: true,
+                                     shouldNormalizeURIPath: true,
+                                     omitSessionToken: false)
+            let signedBodyHeader: AWSSignedBodyHeader = .none
+            let signedBodyValue: AWSSignedBodyValue = .empty
+            let signingConfig = AWSSigningConfig(credentials: credentials,
+                                                 expiration: expiration,
+                                                 signedBodyHeader: signedBodyHeader,
+                                                 signedBodyValue: signedBodyValue,
+                                                 flags: flags,
+                                                 date: date,
+                                                 service: signingName,
+                                                 region: signingRegion,
+                                                 signatureType: .requestQueryParams)
+            let builtRequest = sigV4SignedRequest(requestBuilder: requestBuilder, signingConfig: signingConfig)
+            guard let presignedURL = builtRequest?.endpoint.url else {
+                logger.error("Failed to generate presigend url")
+                return nil
+            }
+            return presignedURL
+        } catch let err {
+            logger.error("Failed to generate presigned url: \(err)")
+            return nil
+        }
+    }
+    
+    public static func sigV4SignedRequest(requestBuilder: SdkHttpRequestBuilder,
+                                          signingConfig: AWSSigningConfig) -> SdkHttpRequest? {
+        let originalRequest = requestBuilder.build()
+        let crtUnsignedRequest = originalRequest.toHttpRequest()
+        let signer = SigV4HttpRequestSigner()
+        do {
+            let result = try signer.signRequest(request: crtUnsignedRequest,
+                                                config: signingConfig.toCRTType())
+            let crtSignedRequest = try result.get()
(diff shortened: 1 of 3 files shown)


AWSClientRuntime/Sources/Signing/AWSSigV4Signer.swift
```diff
@@ -0,0 +1,68 @@
...
+
+public class AWSSigV4Signer {
+    static let logger: SwiftLogger = SwiftLogger(label: "AWSSigV4Signer")
+    
+    public static func sigV4SignedURL(requestBuilder: SdkHttpRequestBuilder,
+                                      credentialsProvider: CredentialsProvider,
+                                      signingName: Swift.String,
+                                      signingRegion: Swift.String,
+                                      date: ClientRuntime.Date,
+                                      expiration: Int64) -> ClientRuntime.URL? {
```
kneekey23, line 19: Question. Since this is the convenience function does it make sense to ever default the expiration or no?

AWSClientRuntime/Sources/Signing/AWSSigV4Signer.swift
```diff
@@ -0,0 +1,68 @@
...
+            }
+            return presignedURL
+        } catch let err {
+            logger.error("Failed to generate presigned url: \(err)")
+            return nil
+        }
+    }
+    
+    public static func sigV4SignedRequest(requestBuilder: SdkHttpRequestBuilder,
+                                          signingConfig: AWSSigningConfig) -> SdkHttpRequest? {
```
kneekey23, line 50: Is there ever a world in which we would want these apis to just throw instead of returning nil or no?

AWSClientRuntime/Tests/Sigv4/SigV4SigningTests.swift
```diff
@@ -0,0 +1,67 @@
...
+import AwsCommonRuntimeKit
+import ClientRuntime
+import XCTest
+
+@testable import AWSClientRuntime
+
+extension Formatter {
+    // Inspired from:
+    // https://stackoverflow.com/questions/28016578/how-can-i-parse-create-a-date-time-stamp-formatted-with-fractional-seconds-utc
+    static let iso8601withFractionalSeconds: DateFormatter = {
```
kneekey23, line 17: I believe we have one of these already in the ClientRuntime?

kneekey23: Actually looks like this isn't being used? Remove?

kneekey23 (approved):
Looks like a great starting point to me! No blockers just some questions and one fix-n-ship!

AWSClientRuntime/Tests/Sigv4/SigV4SigningTests.swift
wooj2: It's being used in the unit test for that specific date format, however, the only reason this exists is because i was copying a unit test from the iOS sdk repo.  I agree though, let me update the test so that we don't need this.

kneekey23: k I believe we do have this exact formatter in ClientRuntime if you do need it.

AWSClientRuntime/Sources/Signing/AWSSigV4Signer.swift
wooj2: I guess this is more of a philosophical consideration -- when should we throw vs when we should just return `nil`.  I've typically thrown exceptions as a way to convey additional information of some error happening so that the caller can optionally write some code that accounts for that error at the time that the code is running.  For example, if a caller receives a "Busy exception", they might write some code to retry calling the function.

In this case, I could imagine us throwing some sort of "Invalid argument exception" -- however, in this context, we aren't the owners of what is valid/invalid -- we are largely plumbing through values to our downstream dependencies, and I don't think we want to be exposing the exceptions from our lower level dependencies (crt, etc).  We could have a mapping of the CRT errors to exceptions we own, but it seems a bit overkill right now.  What do you think?

A second argument against an exception is that I believe, most of the time, this should just work, so I don't see a strong argument for updating the signature of the function to throw.

AWSClientRuntime/Sources/Signing/AWSSigV4Signer.swift
wooj2: Good question -- we could have one, but then what would the default be?  5 minutes? 10 minutes? 1 hour? 1 day?  I don't have a good answer for this.

Also, it seems that some services probably have a maximum for how long a presigned url is valid (or there is something incredibly wrong in our code).  For example, when creating a presigned url w/ polly, I can seemingly only generate presigned urls with a maximum of 5 minutes, despite passing in a higher value like 86400 (24 hours).

AWSClientRuntime/Tests/Sigv4/SigV4SigningTests.swift
wooj2: It was slightly different, but i updated the test to avoid having to use it.

AWSClientRuntime/Sources/Signing/AWSSigV4Signer.swift
kneekey23: ya this is interesting, I guess for now we dont have one until we ever figure out what it should be or we get customer requests. Agree with your analysis here, thanks for providing it.

AWSClientRuntime/Sources/Signing/AWSSigV4Signer.swift
kneekey23: k this is a valid argument, right now ya we don't have a way to map those errors but we just return them as ClientRuntime.crtError and usually the message in the crt error explains the issue but still not a great experience and I agree it's probably overkill right now to try and improve this because it is still working and functional. Let's leave this for now, I think this is justified.

PR #417, opened 2021-09-30
fix: change credential provider to be async with future

kneekey23:
## Description of changes

This PR allows the exposed protocol for a custom credentials provider to be async using a Future instead of a closure.
## Conventional Commits

### DIFF
--- a/AWSClientRuntime/Tests/Auth/AWSCredentialProviderTests.swift
+++ b/AWSClientRuntime/Tests/Auth/AWSCredentialProviderTests.swift
@@ -4,7 +4,7 @@
 //
 // SPDX-License-Identifier: Apache-2.0
 //
-        
+
 import ClientRuntime
 import SmithyTestUtil
 import XCTest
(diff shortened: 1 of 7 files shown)


AWSClientRuntime/Tests/Auth/AWSCredentialProviderTests.swift
```diff
@@ -14,14 +14,46 @@ class AWSCredentialProviderTests: XCTestCase {
...
+        future.fulfill(AWSCredentials(accessKey: "MYACCESSKEY", secret: "sekrit", expirationTimeout: 30))
+        return future
+    }
+}
+
+struct MyCustomThreadTestCredentialsProvider: CredentialsProvider {
+    func getCredentials() throws -> SdkFuture<AWSCredentials> {
+        let group = DispatchGroup()
+        let future = SdkFuture<AWSCredentials>()
+        let sleepVal = Int.random(in: 1...4)
```
wooj2, line 44: Please remove randomization in unit tests -  If we need to sleep, 100ms seems like enough.

AWSClientRuntime/Tests/Auth/AWSCredentialProviderTests.swift
```diff
@@ -14,14 +14,46 @@ class AWSCredentialProviderTests: XCTestCase {
...
-    func getCredentials() throws -> AWSCredentials {
-        return AWSCredentials(accessKey: "MYACCESSKEY", secret: "sekrit", expirationTimeout: 30)
+    func getCredentials() throws -> SdkFuture<AWSCredentials> {
+        let future = SdkFuture<AWSCredentials>()
+        future.fulfill(AWSCredentials(accessKey: "MYACCESSKEY", secret: "sekrit", expirationTimeout: 30))
+        return future
+    }
+}
+
+struct MyCustomThreadTestCredentialsProvider: CredentialsProvider {
```
wooj2, line 40: Thanks for adding unit tests, but i think these tests should focus on correctness.  We should be able to verify that the call is async in nature, such that it does not block the caller -- if we can write a test that does this too, then great, but if not, i think it's fine to not have a test that does this.

AWSClientRuntime/Tests/Auth/AWSCredentialProviderTests.swift
kneekey23: will do

AWSClientRuntime/Tests/Auth/AWSCredentialProviderTests.swift
kneekey23: k tried to address this in a third unit test I think

kneekey23: removing this as not an accurate test

wooj2 (approved):
lgtm assuming you remove unit tests -- lets ship, and have royjit@ test our threading model

PR #421, opened 2021-10-01
feat: Glacier Checksums Customization

### LINKED ISSUE #208: Glacier Request Body Checksum
Compute X-Amz-Sha256-Tree-Hash, and X-Amz-Content-Sha256 checksum for all request bodies

*internal id*: 176722164

kneekey23:
## Issue \#

Closes #208 

Corresponding PRs: https://github.com/awslabs/smithy-swift/pull/362 and https://github.com/awslabs/aws-crt-swift/pull/46

## Conventional Commits

### DIFF
--- a/AWSClientRuntime/Sources/Middlewares/Sha256TreeHashMiddleware.swift
+++ b/AWSClientRuntime/Sources/Middlewares/Sha256TreeHashMiddleware.swift
@@ -1,0 +1,51 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0.
+
+import AwsCommonRuntimeKit
+import ClientRuntime
+
+public struct Sha256TreeHashMiddleware<OperationStackOutput: HttpResponseBinding,
+                                      OperationStackError: HttpResponseBinding>: Middleware {
+    public let id: String = "Sha256TreeHash"
+    
+    private let sha256TreeHashHeaderName = "X-Amz-Sha256-Tree-Hash"
+    
+    public init() {}
+    
+    public func handle<H>(context: Context,
+                          input: MInput,
+                          next: H) -> Result<MOutput, MError>
+    where H: Handler,
+    Self.MInput == H.Input,
+    Self.MOutput == H.Output,
+    Self.Context == H.Context,
+    Self.MError == H.MiddlewareError {
+        let request = input.build()
+        switch request.body {
+        case .data(let data):
+            guard let data = data,
+                  let bodyString = String(data: data, encoding: .utf8) else {
+                return next.handle(context: context, input: input)
+            }
+            let base64Encoded = bodyString.base64EncodedSha256()
+            input.withHeader(name: sha256TreeHashHeaderName, value: base64Encoded)
+        case .stream(let stream):
+            guard let logger = context.getLogger() else {
+                return next.handle(context: context, input: input)
+            }
+            logger.error("TODO: Content-MD5 to stream buffer/reader")
+        default:
+            guard let logger = context.getLogger() else {
+                return next.handle(context: context, input: input)
+            }
+            logger.error("Unhandled case for Content-MD5")
+        }
+        
+        return next.handle(context: context, input: input)
+    }
+
+    public typealias MInput = SdkHttpRequestBuilder
+    public typealias MOutput = OperationOutput<OperationStackOutput>
+    public typealias Context = HttpContext
+    public typealias MError = SdkError<OperationStackError>
+}
--- a/AWSClientRuntime/Sources/Middlewares/Sha256TreeHashMiddleware.swift
+++ b/AWSClientRuntime/Sources/Middlewares/Sha256TreeHashMiddleware.swift
@@ -3,12 +3,15 @@
 
 import AwsCommonRuntimeKit
 import ClientRuntime
+import AwsCCal
 
 public struct Sha256TreeHashMiddleware<OperationStackOutput: HttpResponseBinding,
-                                      OperationStackError: HttpResponseBinding>: Middleware {
+                                       OperationStackError: HttpResponseBinding>: Middleware {
     public let id: String = "Sha256TreeHash"
     
     private let sha256TreeHashHeaderName = "X-Amz-Sha256-Tree-Hash"
+    
+    private let sha256HashHeaderName = "X-Amz-Content-Sha256"
     
     public init() {}
     
@@ -16,36 +19,99 @@ public struct Sha256TreeHashMiddleware<OperationStackOutput: HttpResponseBinding
                           input: MInput,
                           next: H) -> Result<MOutput, MError>
     where H: Handler,
-    Self.MInput == H.Input,
-    Self.MOutput == H.Output,
-    Self.Context == H.Context,
-    Self.MError == H.MiddlewareError {
-        let request = input.build()
-        switch request.body {
-        case .data(let data):
-            guard let data = data,
-                  let bodyString = String(data: data, encoding: .utf8) else {
-                return next.handle(context: context, input: input)
-            }
-            let base64Encoded = bodyString.base64EncodedSha256()
-            input.withHeader(name: sha256TreeHashHeaderName, value: base64Encoded)
-        case .stream(let stream):
-            guard let logger = context.getLogger() else {
-                return next.handle(context: context, input: input)
-            }
-            logger.error("TODO: Content-MD5 to stream buffer/reader")
-        default:
-            guard let logger = context.getLogger() else {
-                return next.handle(context: context, input: input)
-            }
-            logger.error("Unhandled case for Content-MD5")
-        }
+          Self.MInput == H.Input,
+          Self.MOutput == H.Output,
+          Self.Context == H.Context,
+          Self.MError == H.MiddlewareError {
+              let request = input.build()
+              
+              switch request.body {
+              case .data(let data):
+                  guard let data = data else {
+                            return next.handle(context: context, input: input)
+                        }
+                  if !input.currentHeaders.exists(name: sha256HashHeaderName) {
+                      let base64Encoded = ByteBuffer(data: data).base64EncodedSha256()
+                      input.withHeader(name: sha256TreeHashHeaderName, value: base64Encoded)
+                  }
+              case .stream(let stream):
+                  let streamBytes = stream.toBytes()
+                  guard streamBytes.length > 0 else {
+                      return next.handle(context: context, input: input)
+                  }
+                  let (linearHash, treeHash) = computeHashes(bytes: streamBytes)
+                  if let treeHash = treeHash {
+                      input.withHeader(name: sha256TreeHashHeaderName, value: treeHash)
+                  }
+                  if let linearHash = linearHash {
+                      input.withHeader(name: sha256HashHeaderName, value: linearHash)
+                  }
+                  
+                  return next.handle(context: context, input: input)
+                  
+              case .empty, .none:
+                  return next.handle(context: context, input: input)
+              }
+              
+              return next.handle(context: context, input: input)
+          }
+    
+    /// Computes the tree-hash and linear hash of a `ByteBuffer`.
+    /// See http://docs.aws.amazon.com/amazonglacier/latest/dev/checksum-calculations.html for more information.
+    private func computeHashes(bytes: ByteBuffer) -> (String?, String?) {
+        let bufferSize = 1024 * 1024
+        var hashes = [[UInt8]]()
         
-        return next.handle(context: context, input: input)
-    }
+        while true {
+            var oneMbTempBuffer = ByteBuffer(size: bufferSize)
+            let bytesRead = bytes.readIntoBuffer(buffer: &oneMbTempBuffer)
+            if bytesRead == 0 {
+                break
+            }
+            let hash: ByteBuffer = oneMbTempBuffer.sha256()
+            hashes.append(hash.toByteArray())
+        }
 
+        return (bytes.base64EncodedSha256(), computeTreeHash(hashes: hashes))
+    }
+    
+    /// Builds a tree hash root node given a slice of hashes. Glacier tree hash to be derived from SHA256 hashes of 1MBm chunks of the data.
+    /// See http://docs.aws.amazon.com/amazonglacier/latest/dev/checksum-calculations.html for more information.
+    private func computeTreeHash(hashes: [[UInt8]]) -> String? {
+        guard !hashes.isEmpty else {
+            return nil
+        }
+        let hashCount = hashes.count
+        
+        if hashCount == 1 {
+            let byteArray = hashes[0]
+            let byteBuffer = ByteBuffer(bytes: byteArray)
+            return byteBuffer.toData().base64EncodedString()
+        }
+        var tempHashes = [[UInt8]]()
+        for index in stride(from: 0, to: hashCount, by: 2) {
+            if index + 1 <= hashCount - 1 {
+                var tempHashArray = [UInt8]()
+                tempHashArray.append(contentsOf: hashes[index])
+                tempHashArray.append(contentsOf: hashes[index + 1])
+                let tempByteBuffer = ByteBuffer(bytes: tempHashArray)
+                
+                let tempSha256: ByteBuffer = tempByteBuffer.sha256()
+                tempHashes.append(tempSha256.toByteArray())
+                
+            } else {
+                tempHashes.append(hashes[index])
+            }
+        }
+        let byteBuf = ByteBuffer(bytes: tempHashes[0])
+       
+        return byteBuf.sha256().toData().base64EncodedString()
+    }
+    
     public typealias MInput = SdkHttpRequestBuilder
     public typealias MOutput = OperationOutput<OperationStackOutput>
     public typealias Context = HttpContext
     public typealias MError = SdkError<OperationStackError>
 }
+
+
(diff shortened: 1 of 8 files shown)


AWSClientRuntime/Sources/Middlewares/Sha256TreeHashMiddleware.swift
```diff
@@ -0,0 +1,118 @@
...
+import ClientRuntime
+import AwsCCal
+
+public struct Sha256TreeHashMiddleware<OperationStackOutput: HttpResponseBinding,
+                                       OperationStackError: HttpResponseBinding>: Middleware {
+    public let id: String = "Sha256TreeHash"
+    
+    private let sha256TreeHashHeaderName = "X-Amz-Sha256-Tree-Hash"
+    
+    private let sha256HashHeaderName = "X-Amz-Content-Sha256"
```
wooj2, line 14: I think the code would read better if the variables were renamed to:
X_AMZ_SHA256_TREE_HASH_HEADER_NAME
X_AMZ_CONTENT_SHA256_HEADER_NAME

AWSClientRuntime/Sources/Middlewares/Sha256TreeHashMiddleware.swift
```diff
@@ -0,0 +1,118 @@
...
+          Self.Context == H.Context,
+          Self.MError == H.MiddlewareError {
+              let request = input.build()
+              
+              switch request.body {
+              case .data(let data):
+                  guard let data = data else {
+                            return next.handle(context: context, input: input)
+                        }
+                  if !request.headers.exists(name: sha256HashHeaderName) {
```
wooj2, line 33: Is there a case where this header already exists as part of the request?

AWSClientRuntime/Sources/Middlewares/Sha256TreeHashMiddleware.swift
```diff
@@ -0,0 +1,118 @@
...
+                  guard streamBytes.length > 0 else {
+                      return next.handle(context: context, input: input)
+                  }
+                  let (linearHash, treeHash) = computeHashes(bytes: streamBytes)
+                  if let treeHash = treeHash {
+                      input.withHeader(name: sha256TreeHashHeaderName, value: treeHash)
+                  }
+                  if let linearHash = linearHash {
+                      input.withHeader(name: sha256HashHeaderName, value: linearHash)
+                  }
```
wooj2, line 48: It seems like we want to only want to add both of these headers or none, right?  If so, does the following work?
```
guard let (linearHash, treeHash) = computeHashes(...) else{
   return next.handle(..)
}
input.withHeader(name: sha256TreeHashHeaderName, treeHash)
input.withHeader(name: sha256HashHeaderName, linearHash)

```

AWSClientRuntime/Sources/Middlewares/Sha256TreeHashMiddleware.swift
```diff
@@ -0,0 +1,118 @@
...
+                  
+                  return next.handle(context: context, input: input)
+                  
+              case .empty, .none:
+                  return next.handle(context: context, input: input)
+              }
+              
+              return next.handle(context: context, input: input)
```
wooj2, line 56: We seem to be repeating next.handle() quite a bit -- is there any way to reduce the number of times we have to write this?

AWSClientRuntime/Sources/Middlewares/Sha256TreeHashMiddleware.swift
```diff
@@ -0,0 +1,118 @@
...
+            let hash = oneMbTempBuffer.sha256()
+            hashes.append(hash.toByteArray())
+        }
+
+        return (bytes.base64EncodedSha256(), computeTreeHash(hashes: hashes))
+    }
+    
+    /// Builds a tree hash root node given a slice of hashes. Glacier tree hash to be derived from SHA256 hashes of 1MBm chunks of the data.
+    /// See http://docs.aws.amazon.com/amazonglacier/latest/dev/checksum-calculations.html for more information.
+    private func computeTreeHash(hashes: [[UInt8]]) -> String? {
```
wooj2, line 80: I'm surprised that this isn't offered by the CRT?

AWSClientRuntime/Sources/Middlewares/Sha256TreeHashMiddleware.swift
```diff
@@ -0,0 +1,118 @@
...
+                let tempByteBuffer = ByteBuffer(bytes: tempHashArray)
+                
+                let tempSha256 = tempByteBuffer.sha256()
+                tempHashes.append(tempSha256.toByteArray())
+                
+            } else {
+                tempHashes.append(hashes[index])
+            }
+        }
+         let byteBuf = ByteBuffer(bytes: tempHashes[0])
```
wooj2, line 106: 1.  spacing seems off here
2.  This seems off, because in the last for loop we did a bunch of work to append a bunch of data to tempHashes, and in this line, we are only using the first byte.  Something seems off.

AWSClientRuntime/Sources/Middlewares/Sha256TreeHashMiddleware.swift
```diff
@@ -0,0 +1,118 @@
...
+                tempHashArray.append(contentsOf: hashes[index])
+                tempHashArray.append(contentsOf: hashes[index + 1])
+                let tempByteBuffer = ByteBuffer(bytes: tempHashArray)
+                
+                let tempSha256 = tempByteBuffer.sha256()
+                tempHashes.append(tempSha256.toByteArray())
+                
+            } else {
+                tempHashes.append(hashes[index])
+            }
```
wooj2, line 104: hmmm if i read this correctly, we don't have to call `.sha256` on the last byte?

AWSClientRuntime/Tests/Middlewares/Sha256TreeHashMiddlewareTests.swift
```diff
@@ -0,0 +1,80 @@
...
+import XCTest
+import ClientRuntime
+import AwsCommonRuntimeKit
+@testable import AWSClientRuntime
+
+class Sha256TreeHashMiddlewareTests: XCTestCase {
+    func testTreeHashAllZeroes() {
+        let context = HttpContextBuilder().build()
+        let byteArray: [UInt8] = Array(repeating: 0, count: 5767168)
+        let byteBuffer = ByteBuffer(bytes: byteArray) // 5.5MB buffer
```
wooj2, line 17: is there something special about a 5.5MB buffer?  if not, plz remove this comment.

AWSClientRuntime/Tests/Middlewares/Sha256TreeHashMiddlewareTests.swift
```diff
@@ -0,0 +1,80 @@
...
+        let streamInput = StreamInput(body: ByteStream.buffer(byteBuffer))
+        var stack = OperationStack<StreamInput, StreamOutput, StreamError>(id: "TreeHashMiddlewareTestStack")
+        stack.serializeStep.intercept(position: .before, middleware: MockBodyMiddleware())
+        stack.finalizeStep.intercept(position: .after, middleware: Sha256TreeHashMiddleware())
+        _ = stack.handleMiddleware(context: context, input: streamInput, next: MockHandler(callback: { context, input in
+            let linear = input.headers.value(for: "X-Amz-Content-Sha256")
+            XCTAssertEqual(linear, "czz1E0SM5rIK0bxeUOsnwGrvrgwyBxOl3Zn05RvBymA=")
+            let treeHash = input.headers.value(for: "X-Amz-Sha256-Tree-Hash")
+            XCTAssertEqual(treeHash, "x7O1UHyNIXvn/g19LDbAILBPDnqTAz6FwI+MsNCADq0=")
+        }))
```
wooj2, line 27: If this callback doesn't get called, then we will pass the test -- One possible solution is that we use an `expectation()` to ensure that this callback gets called.

AWSClientRuntime/Tests/Middlewares/Sha256TreeHashMiddlewareTests.swift
```diff
@@ -0,0 +1,80 @@
...
+    typealias MInput = SerializeStepInput<StreamInput>
+    
+    typealias MOutput = OperationOutput<StreamOutput>
+    
+    typealias Context = HttpContext
+    
+    typealias MError = SdkError<StreamError>
+    
+    
+}
```
wooj2, line 79: 1.  Ideally this would be in SmithyTestUtil (or re-use the mock classes we already have in there)
2. If not in SmithyTestUtil, then perhaps somewhere inside of the test portion of AWSClientRuntime?

codegen/smithy-aws-swift-codegen/src/main/kotlin/software/amazon/smithy/aws/swift/codegen/customization/glacier/GlacierChecksum.kt
```diff
@@ -0,0 +1,41 @@
...
+import software.amazon.smithy.swift.codegen.model.hasTrait
+
+/**
+ * Adds a middleware for Glacier to autofill accountId when not set
+ * See: https://github.com/awslabs/aws-sdk-swift/issues/208
+ * See also: https://docs.aws.amazon.com/amazonglacier/latest/dev/checksum-calculations.html
+ */
+class GlacierChecksum : SwiftIntegration {
+    override fun enabledForService(model: Model, settings: SwiftSettings) =
+        model.expectShape<ServiceShape>(settings.service).sdkId.equals("Glacier", ignoreCase = true)
```
wooj2, line 25: We should avoid using `ignoreCase` -- can we just explicitly list out all the services this Integration applies to?

AWSClientRuntime/Sources/Middlewares/Sha256TreeHashMiddleware.swift
```diff
@@ -0,0 +1,118 @@
...
+          Self.MOutput == H.Output,
+          Self.Context == H.Context,
+          Self.MError == H.MiddlewareError {
+              let request = input.build()
+              
+              switch request.body {
+              case .data(let data):
+                  guard let data = data else {
+                            return next.handle(context: context, input: input)
+                        }
```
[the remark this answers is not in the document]
kneekey23: corrected in the entire file

AWSClientRuntime/Tests/Middlewares/Sha256TreeHashMiddlewareTests.swift
kneekey23: corrected

AWSClientRuntime/Sources/Middlewares/Sha256TreeHashMiddleware.swift
kneekey23: 1. corrected
2. we only need the root of the tree we have created. the root is at the zeroth index of tempHashes which is an array of hashes

AWSClientRuntime/Sources/Middlewares/Sha256TreeHashMiddleware.swift
kneekey23: removed it from one of the cases and put. a break in the switch for .empty and .none so we only calll it at the end

codegen/smithy-aws-swift-codegen/src/main/kotlin/software/amazon/smithy/aws/swift/codegen/customization/glacier/GlacierChecksum.kt
kneekey23: I guess why should we avoid using it?

AWSClientRuntime/Sources/Middlewares/Sha256TreeHashMiddleware.swift
kneekey23: this desnt compile. cant do a guard let for a tuple of optionals for some reason but yes you are correct, ill combine them into one if let

AWSClientRuntime/Tests/Middlewares/Sha256TreeHashMiddlewareTests.swift
wooj2: I see, we could have a variable called
```
let bytesIn5_5MB: Int = 1024*1024*5.5
```

codegen/smithy-aws-swift-codegen/src/main/kotlin/software/amazon/smithy/aws/swift/codegen/customization/glacier/GlacierChecksum.kt
wooj2: Other than a relatively negligible performance hit, I think using ignoreCase seems less direct.  Reading it with `ignoreCase = true` makes me think there we are targeting two services: one named `GLACIER` and one named `Glacier`.

kneekey23: addressed

AWSClientRuntime/Sources/Middlewares/Sha256TreeHashMiddleware.swift
```diff
@@ -0,0 +1,110 @@
...
+                break
+            }
+            let hash = oneMbTempBuffer.sha256()
+            hashes.append(hash.toByteArray())
+        }
+        
+        return (bytes.sha256().encodeToHexString(), computeTreeHash(hashes: hashes))
+    }
+    
+    /// Builds a tree hash root node given a slice of hashes. Glacier tree hash to be derived from SHA256 hashes of 1MBm chunks of the data.
```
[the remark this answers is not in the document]
kneekey23: fixed

AWSClientRuntime/Sources/Middlewares/Sha256TreeHashMiddleware.swift
```diff
@@ -0,0 +1,107 @@
...
+    /// See http://docs.aws.amazon.com/amazonglacier/latest/dev/checksum-calculations.html for more information.
+    private func computeTreeHash(hashes: inout [[UInt8]]) -> String? {
+        guard !hashes.isEmpty else {
+            return nil
+        }
+        
+        while hashes.count > 1 {
+            var tempHashes = [[UInt8]]()
+            for index in stride(from: 0, to: hashes.count, by: 2) {
+                if index + 1 <= hashes.count - 1 {
```
wooj2, line 83: is it possible to do `hashes.count - index > 1` so that we mirror the example?

AWSClientRuntime/Sources/Middlewares/Sha256TreeHashMiddleware.swift
```diff
@@ -0,0 +1,107 @@
...
+    
+    /// Builds a tree hash root node given a slice of hashes. Glacier tree hash to be derived from SHA256 hashes of 1MB chunks of the data.
+    /// See http://docs.aws.amazon.com/amazonglacier/latest/dev/checksum-calculations.html for more information.
+    private func computeTreeHash(hashes: inout [[UInt8]]) -> String? {
+        guard !hashes.isEmpty else {
+            return nil
+        }
+        
+        while hashes.count > 1 {
+            var tempHashes = [[UInt8]]()
```
wooj2, line 81: Can we rename `tempHashes` to `currLvlHashes` to map to the example code?

AWSClientRuntime/Sources/Middlewares/Sha256TreeHashMiddleware.swift
```diff
@@ -0,0 +1,107 @@
...
+            for index in stride(from: 0, to: hashes.count, by: 2) {
+                if index + 1 <= hashes.count - 1 {
+                    var tempHashArray = [UInt8]()
```
wooj2, line 84: rename `tempHashArray` to `md`
rename `tempByteBuffer` to `mdByteBuffer`
rename `tempSha256` to `messageDigest`

AWSClientRuntime/Sources/Middlewares/Sha256TreeHashMiddleware.swift
```diff
@@ -0,0 +1,107 @@
...
+        return (bytes.sha256().encodeToHexString(), computeTreeHash(hashes: &hashes))
+    }
+    
+    /// Builds a tree hash root node given a slice of hashes. Glacier tree hash to be derived from SHA256 hashes of 1MB chunks of the data.
+    /// See http://docs.aws.amazon.com/amazonglacier/latest/dev/checksum-calculations.html for more information.
+    private func computeTreeHash(hashes: inout [[UInt8]]) -> String? {
+        guard !hashes.isEmpty else {
+            return nil
+        }
+
```
wooj2, line 79: It's typically not a good idea to directly modify an input parameter, can we also do something like
```
var prevLvlHashes = hashes
```

wooj2 (approved):
I think it would be better to update naming/ and ifstatement to make it look more similar to the example, other than that, looks good to me!

AWSClientRuntime/Sources/Middlewares/Sha256TreeHashMiddleware.swift
kneekey23: updated

AWSClientRuntime/Sources/Middlewares/Sha256TreeHashMiddleware.swift
kneekey23: renamed

AWSClientRuntime/Sources/Middlewares/Sha256TreeHashMiddleware.swift
kneekey23: updated

AWSClientRuntime/Sources/Middlewares/Sha256TreeHashMiddleware.swift
kneekey23: updated

PR #435, opened 2021-10-11
chore: Update doc generation to not include Models

wooj2:
- Generating documentation in html format is 1.3GB is too large and will adversely impact our repo
- We moved to markdown, which resulted in generation of the github pages site to timeout.
- Pairing down documentation to avoid generating documentation for all models and only include top level client calls.  If customers want to view documentation for specific input/output/error types, it seems sufficient that they can view this directly in XCode using DocC.


kneekey23 (approved):
LGTM, thanks for this change

PR #452, opened 2021-10-19
fix: Remove DEVELOPER_DIR from CI

kneekey23:
## Description of changes

This PR fixes CI where the DEVELOPER_DIR is no longer needed on this github actions host and because it now defaults to swift 5.5, the package.swift also needed to be updated to handle the new api.

### DIFF
--- a/codegen/Package.swift
+++ b/codegen/Package.swift
@@ -79,7 +79,7 @@ func appendLibTarget(name: String, path: String) {
 func appendTstTarget(name: String, path: String, dependency: String) {
     package.targets.append(.testTarget(name: name,
                                        dependencies:  [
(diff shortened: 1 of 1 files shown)


.github/workflows/continuous-integration.yml
```diff
@@ -22,8 +22,6 @@ env:
 jobs:
   ios-compat:
     runs-on: macos-11
-    env:
-      DEVELOPER_DIR: /Applications/Xcode_12.5.app
```
wooj2, line 26: Seems like we should update this in generate-docs.yml and release.yml ?

wooj2 (approved):
lgtm!

PR #478, opened 2021-11-03
chore: improve developer experience with region resolver

wooj2:
**Use case 1:**
* assume customer has logging disabled
* no region provider is properly configured, then customers will see:
```
<No output, SDK will be silent>
```

**Use case 2:**
* Assume customer has logging enabled to `debug` level: `SDKLoggingSystem.initialize(logLevel: .debug)`
* No region provider
Result is that customers will see:
```
2021-11-03T14:55:13-0700 debug DefaultRegionProvider : Attempting to resolve region with: EnvironmentRegionProvider
2021-11-03T14:55:13-0700 debug DefaultRegionProvider : Attempting to resolve region with: ProfileRegionProvider
2021-11-03T14:55:13-0700 debug ProfileRegionResolver : No default profile collection was found at the path of ~/.aws/config
2021-11-03T14:55:13-0700 debug DefaultRegionProvider : Attempting to resolve region with: IMDSRegionProvider
2021-11-03T14:55:14-0700 debug DefaultRegionProvider : Unable to resolve region
```

**Use case 3 (happy path, with debug)**
* Assume customer has logging enabled to `debug` level: `SDKLoggingSystem.initialize(logLevel: .debug)`
* credentials/config file is available
Users will see:
```
2021-11-03T14:55:56-0700 debug DefaultRegionProvider : Attempting to resolve region with: EnvironmentRegionProvider
2021-11-03T14:55:56-0700 debug DefaultRegionProvider : Attempting to resolve region with: ProfileRegionProvider
2021-11-03T14:55:56-0700 debug DefaultRegionProvider : Resolved region with: ProfileRegionProvider
```

**Use case 4 (happy path, without debug)**
* Assume customer has logging disabled
* credentials/config file is available
Users will see:
```
<No output, SDK will be silent>
```

### DIFF
--- a/AWSClientRuntime/Sources/Regions/DefaultRegionResolver.swift
+++ b/AWSClientRuntime/Sources/Regions/DefaultRegionResolver.swift
@@ -5,26 +5,32 @@
 // SPDX-License-Identifier: Apache-2.0
 //
 import AwsCommonRuntimeKit
+import ClientRuntime
 
 public struct DefaultRegionResolver: RegionResolver {
     public let providers: [RegionProvider]
-    
+    let logger: SwiftLogger
+
     public init(providers: [RegionProvider] = [EnvironmentRegionProvider(), ProfileRegionProvider(), IMDSRegionProvider()]) {
         // TODO: add more region resolvers i.e. System Properties, etc
         self.providers = providers
+        self.logger = SwiftLogger(label: "DefaultRegionProvider")
     }
     
     public func resolveRegion() -> String? {
         for provider in providers {
             do {
                 // TODO: Verify that this does not block calling thread. If it does, we need to solve this similar to the way we solve this for the credentials provider
+                logger.debug("Attempting to resolve region with: \(String(describing: type(of: provider)))")
                 if let region = try provider.resolveRegion().get() {
+                    logger.debug("Resolved region with: \(String(describing: type(of: provider)))")
                     return region
                 }
             } catch {
                 return nil
             }
         }
+        logger.error("Unable to resolve region")
         return nil
     }
 }
--- a/AWSClientRuntime/Sources/Regions/DefaultRegionResolver.swift
+++ b/AWSClientRuntime/Sources/Regions/DefaultRegionResolver.swift
@@ -30,7 +30,7 @@ public struct DefaultRegionResolver: RegionResolver {
                 return nil
             }
         }
-        logger.error("Unable to resolve region")
+        logger.debug("Unable to resolve region")
         return nil
     }
 }
--- a/AWSClientRuntime/Sources/Regions/ProfileRegionResolver.swift
+++ b/AWSClientRuntime/Sources/Regions/ProfileRegionResolver.swift
@@ -30,17 +30,18 @@ public struct ProfileRegionProvider: RegionProvider {
 
         let profileCollection = profileCollection ?? CRTAWSProfileCollection(fromFile: path, source: .config)
         guard let profileCollection = profileCollection else {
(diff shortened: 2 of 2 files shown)


AWSClientRuntime/Sources/Regions/DefaultRegionResolver.swift
```diff
@@ -5,26 +5,32 @@
...
+                logger.debug("Attempting to resolve region with: \(String(describing: type(of: provider)))")
                 if let region = try provider.resolveRegion().get() {
+                    logger.debug("Resolved region with: \(String(describing: type(of: provider)))")
                     return region
                 }
             } catch {
                 return nil
             }
         }
+        logger.error("Unable to resolve region")
```
wooj2, line 33: @kneekey23  -- do you know of any cases where returning nil isn't an error?

kneekey23: returning nil is never an error. user could provide a region either at the client level or later on at the operation level. this default chain is basically to get a region if they didnt provide it. but they should be able to even make a request without providing a region and then see an error

kneekey23: love all these changes minus this one. can we just do logger.debug here or logger.info instead of error? it's not an error if this region resolver didn't find a region, this is a nicety the sdk provides so they dont have to give us a region. and in certain cases like lambda or ec2, they shouldnt have to at all.

wooj2: Updated to `.debug`

wooj2: FYI -- this means that case1 and case4 (see above) output nothing from the SDK -- (i think this is the behavior we want)

kneekey23: ya I agree, its confusing as .info at least some customers keep getting confused thinking it's an error when it's not. looks good to me.

AWSClientRuntime/Sources/Regions/DefaultRegionResolver.swift
```diff
@@ -5,26 +5,32 @@
...
+                logger.debug("Attempting to resolve region with: \(String(describing: type(of: provider)))")
                 if let region = try provider.resolveRegion().get() {
+                    logger.debug("Resolved region with: \(String(describing: type(of: provider)))")
                     return region
                 }
             } catch {
                 return nil
             }
         }
+        logger.debug("Unable to resolve region")
```
kneekey23, line 33: can we update this to "Unable to resolve region using the DefaultRegionResolver chain" possibly?

kneekey23 (approved):
LGTM, just one ask

AWSClientRuntime/Sources/Regions/DefaultRegionResolver.swift
wooj2: We are already outputting:

```
2021-11-03T14:55:14-0700 debug DefaultRegionProvider : Unable to resolve region
```

If we change it, it will look like:
```
2021-11-03T14:55:14-0700 debug DefaultRegionProvider : Unable to resolve region using the DefaultRegionResolver chain
```

This seems like duplicated information.  I'm assuming that we do not want to change this, but if we do, we can change it in a subsequent PR.

kneekey23: Ah yes sorry I didn't realize that's how it looked. Glad you kept it the same

PR #479, opened 2021-11-04
feat: region resolver via bundle

wooj2:
To make use of `BundleRegionProvider` pulling the region from your bundle:
1.  Create an iOS project (or some sort of project which uses a Bundle)
2.  Add `<debug,production>.xcconfig` to each one of your targets (and assign appropriately)
3. So that the values of these variables are available during execution time of the code, you MUST, inside each of your .xcconfig files, define a variable like:

For Debug.xcconfig
```
AWS_REGION=us-west-2
```

For Production.xcconfig
```
AWS_REGION=us-east-1
```
4. In your Info.plist, add a **Key** with `AWS_REGION` of **Type** `String`, with value: `$(AWS_REGION)`.  For example:
```
	<key>AWS_REGION</key>
	<string>$(AWS_REGION)</string>
```

5.  Attempt to invoke the `BundleRegionProvider`.

### DIFF
--- a/AWSClientRuntime/Sources/Config/BundleConfiguration.swift
+++ b/AWSClientRuntime/Sources/Config/BundleConfiguration.swift
@@ -5,7 +5,7 @@
 // SPDX-License-Identifier: Apache-2.0
 //
 
-import Foundation
+import Foundation.NSBundle
 
 // Adapted from: https://nshipster.com/xcconfig/
 struct BundleConfiguration {
--- a/AWSClientRuntime/Sources/Config/BundleConfiguration.swift
+++ b/AWSClientRuntime/Sources/Config/BundleConfiguration.swift
@@ -1,0 +1,30 @@
+//
+// Copyright Amazon.com Inc. or its affiliates.
+// All Rights Reserved.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import Foundation
+
+// Adapted from: https://nshipster.com/xcconfig/
+struct BundleConfiguration {
+    enum Error: Swift.Error {
+        case missingKey, invalidValue
+    }
+
+    static func value<T>(bundle: Foundation.Bundle, for key: String) throws -> T where T: LosslessStringConvertible {
+        guard let object = bundle.object(forInfoDictionaryKey: key) else {
+            throw Error.missingKey
+        }
+        switch object {
+        case let value as T:
+            return value
+        case let string as String:
+            guard let value = T(string) else { fallthrough }
+            return value
+        default:
+            throw Error.invalidValue
+        }
+    }
+}
--- a/AWSClientRuntime/Sources/Config/BundleConfiguration.swift
+++ b/AWSClientRuntime/Sources/Config/BundleConfiguration.swift
@@ -5,15 +5,13 @@
 // SPDX-License-Identifier: Apache-2.0
 //
 
-import Foundation.NSBundle
-
 // Adapted from: https://nshipster.com/xcconfig/
 struct BundleConfiguration {
     enum Error: Swift.Error {
         case missingKey, invalidValue
     }
 
-    static func value<T>(bundle: Foundation.Bundle, for key: String) throws -> T where T: LosslessStringConvertible {
+    static func value<T>(bundle: Bundle, for key: String) throws -> T where T: LosslessStringConvertible {
         guard let object = bundle.object(forInfoDictionaryKey: key) else {
             throw Error.missingKey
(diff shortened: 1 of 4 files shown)


AWSClientRuntime/Sources/Regions/BundleRegionProvider.swift
```diff
@@ -0,0 +1,40 @@
...
+    public init(bundle: Foundation.Bundle = Bundle.main, regionKey: String = "AWS_REGION") {
+        self.logger = SwiftLogger(label: "BundleRegionProvider")
+        self.bundle = bundle
+        self.regionKey = regionKey
+    }
+    
+    public func resolveRegion() -> Future<String?> {
+        let future = Future<String?>()
+        #if os(iOS) || os(watchOS) || os(tvOS)
+        future.fulfill(region())
```
wooj2, line 25: From a security perspective, if we allow customers to enter "anything" into a plist -- this can be bad -- we probably want to, at the very minimum, limit the size of string we can read from a plist -- e.g. 20 bytes maximum.

AWSClientRuntime/Sources/Config/BundleConfiguration.swift
```diff
@@ -0,0 +1,30 @@
+//
+// Copyright Amazon.com Inc. or its affiliates.
+// All Rights Reserved.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import Foundation
```
kneekey23, line 8: question. can we be specific here with our import? like `import struct Foundation.Bundle`?

AWSClientRuntime/Sources/Regions/BundleRegionProvider.swift
kneekey23: how can we do that?

AWSClientRuntime/Sources/Regions/DefaultRegionResolver.swift
```diff
@@ -11,7 +11,7 @@ public struct DefaultRegionResolver: RegionResolver {
     public let providers: [RegionProvider]
     let logger: SwiftLogger
 
-    public init(providers: [RegionProvider] = [EnvironmentRegionProvider(), ProfileRegionProvider(), IMDSRegionProvider()]) {
+    public init(providers: [RegionProvider] = [EnvironmentRegionProvider(), ProfileRegionProvider(), IMDSRegionProvider(), BundleRegionProvider()]) {
```
kneekey23, line 14: can you actually move this one to be first in the list possibly? for iOS, all the other ones won't even work so I think it should be first. For linux it just will quickly pass over it.

kneekey23 (approved):
this honestly looks great to me! if we can solve the security concern, I think this is good to ship!

AWSClientRuntime/Sources/Config/BundleConfiguration.swift
wooj2: I tried that, and it didn't work.  I can try again.

wooj2: works:
```
import Foundation.NSBundle
```

AWSClientRuntime/Sources/Regions/BundleRegionProvider.swift
wooj2: Done. used `prefix()`

AWSClientRuntime/Sources/Regions/DefaultRegionResolver.swift
```diff
@@ -11,7 +11,7 @@ public struct DefaultRegionResolver: RegionResolver {
     public let providers: [RegionProvider]
     let logger: SwiftLogger
 
-    public init(providers: [RegionProvider] = [EnvironmentRegionProvider(), ProfileRegionProvider(), IMDSRegionProvider()]) {
+    public init(providers: [RegionProvider] = [BundleRegionProvider(), EnvironmentRegionProvider(), ProfileRegionProvider(), IMDSRegionProvider()]) {
         // TODO: add more region resolvers i.e. System Properties, etc
```
kneekey23, line 15: Can remove this todo now :). This is the last one!

kneekey23 (approved):
Love this now, looks perfect to me

AWSClientRuntime/Sources/Regions/DefaultRegionResolver.swift
wooj2: removed.

PR #491, opened 2021-11-11
feat: add s3 get object presigner

wooj2:
Adding a presigner for S3's Get Object -- kind of quick and dirty in that we aren't supporting a the input type's that are of type `Date` nor `Int` -- added a TODO for this.

Tested with a quick and dirty hello world app:
```
struct MySpecialEndpointResolver: EndpointResolver {
    func resolve(serviceId: String, region: String) throws -> AWSEndpoint {
        return AWSEndpoint(endpoint: Endpoint(host: "s3.amazonaws.com"))
    }
}
func genPresignedUrlForS3GetObject() throws{
    let s3Config = try S3Client.S3ClientConfiguration( endpointResolver: MySpecialEndpointResolver(), region: "us-west-2")
    let input = GetObjectInput(bucket: "<INSERT YOUR BUCKET>", key: "<key>")
    guard let url = input.presignURL(config: s3Config, expiration: 10000) else {
        print("exiting")
        exit(1)
    }
    print(url)
}
try genPresignedUrlForS3GetObject()
```

(note you may not need `MySpecialEndpointResolver` if you are executing in us-east-1)

### DIFF
--- a/codegen/smithy-aws-swift-codegen/src/main/kotlin/software/amazon/smithy/aws/swift/codegen/customization/presignable/PresignableGetIntegration.kt
+++ b/codegen/smithy-aws-swift-codegen/src/main/kotlin/software/amazon/smithy/aws/swift/codegen/customization/presignable/PresignableGetIntegration.kt
@@ -1,9 +1,9 @@
-package software.amazon.smithy.aws.swift.codegen.customization.polly
+package software.amazon.smithy.aws.swift.codegen.customization.presignable
 
 import software.amazon.smithy.aws.swift.codegen.AWSClientRuntimeTypes
 import software.amazon.smithy.aws.swift.codegen.PresignableOperation
 import software.amazon.smithy.aws.swift.codegen.middleware.AWSSigningMiddleware
-import software.amazon.smithy.aws.swift.codegen.middleware.SynthesizeSpeechInputGETQueryItemMiddleware
+import software.amazon.smithy.aws.swift.codegen.middleware.InputTypeGETQueryItemMiddlewareRenderable
 import software.amazon.smithy.aws.traits.auth.UnsignedPayloadTrait
 import software.amazon.smithy.codegen.core.Symbol
 import software.amazon.smithy.model.Model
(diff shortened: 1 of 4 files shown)


codegen/smithy-aws-swift-codegen/src/main/kotlin/software/amazon/smithy/aws/swift/codegen/customization/presignable/PresignableGetIntegration.kt
```diff
@@ -166,7 +178,15 @@ class PollyGetPresignerIntegration(private val presignedOperations: Map<String,
             .build()
         delegator.useShapeWriter(headerMiddlewareSymbol) { writer ->
             writer.addImport(SwiftDependency.CLIENT_RUNTIME.target)
-            val queryItemMiddleware = PollySynthesizeSpeechGETQueryItemMiddleware(ctx, inputSymbol, outputSymbol, outputErrorSymbol, inputShape, writer)
+            val queryItemMiddleware =
+                software.amazon.smithy.aws.swift.codegen.customization.InputTypeGETQueryItemMiddleware(
```
kneekey23, line 182: Nit: Can we add the import here to the top of the file instead of having this inline?

kneekey23 (approved):
LGTM!! Nice refactor!

codegen/smithy-aws-swift-codegen/src/main/kotlin/software/amazon/smithy/aws/swift/codegen/customization/presignable/PresignableGetIntegration.kt
wooj2: are you referring to the line that says: `writer.addImport(SwiftDependency.CLIENT_RUNTIME.target)` -- where else would we put this?

wooj2: OH! i see.. wow.. i wonder why it didn't add an import statement.

PR #492, opened 2021-11-11
fix: Language metadata to render extras properly

wooj2:
Prior to doing this, we were improperly generating user agent strings because `.map{}` generates an `[String]` rather than `String` -- so, we need this update to properly generate a user agent string.  Also, I added some unit tests.


kneekey23 (approved):
LGTM! Sorry for this mistake!

PR #493, opened 2021-11-11
chore: add frameworkmetadata and surrounding unit tests

wooj2:
Prior to figuring out how to expose this for our customers, it seems like this is heading in the right direction, so i wanted to:
* Gain consensus that this is the correct direction
* add surrounding unit tests so we make sure we aren't breaking anything


AWSClientRuntime/Tests/Http/AWSUserAgentMetadataTests.swift
```diff
@@ -0,0 +1,84 @@
...
+// Copyright Amazon.com Inc. or its affiliates.
+// All Rights Reserved.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import XCTest
+@testable import AWSClientRuntime
+
+class AWSUseragentMetadataTests: XCTestCase {
```
kneekey23, line 11: were you not able to find the existing user agent tests?

AWSClientRuntime/Tests/Http/FrameworkMetadataTests.swift
```diff
@@ -0,0 +1,50 @@
...
+                      "test2": "SomethingOtherThanANumber"]
+
+        let sut = FrameworkMetadata(name: "aws-amplify", version: "2.0.0", extras: extras)
+        
+        let option1 = "lib/aws-amplify/2.0.0 md/test1/1.0 md/test2/SomethingOtherThanANumber" == sut.description
+        let option2 = "lib/aws-amplify/2.0.0 md/test2/SomethingOtherThanANumber md/test1/1.0" == sut.description
+        XCTAssert(option1 || option2)
+    }
+
+    func testMetadataIsSanitized() {
```
kneekey23, line 43: nice test!

kneekey23 (approved):
LGTM, just had one question about testing but definitely in the right direction!

AWSClientRuntime/Tests/Http/AWSUserAgentMetadataTests.swift
wooj2: Ah, I didn't notice there were already UserAgentTests inside of a folder called "UserAgent" -- considering the source files were located in a folder called `Http` I was expecting the accompanying unit tests would be there as well -- i'll move the test over. to be consistent

PR #494, opened 2021-11-11
feat: expose ability to set framework in useragent

wooj2:
* This pr exposes setting the framework meta data via the Configuration struct as a quick way for Amplify to set this value in the short term.  It will be eventually removed/refactored once we provide interceptor capability.  Kotlin is doing this via an environment variable, but this does not seem applicable since amplify won't have the ability to guarantee the environment variable is set when a customer adopts their framework.

Reference:
https://github.com/awslabs/aws-sdk-kotlin/pull/372


codegen/smithy-aws-swift-codegen/src/main/kotlin/software/amazon/smithy/aws/swift/codegen/middleware/UserAgentMiddleware.kt
```diff
@@ -22,10 +22,11 @@ class UserAgentMiddleware(val settings: SwiftSettings) : MiddlewareRenderable {
     override val position = MiddlewarePosition.BEFORE
 
     override fun render(writer: SwiftWriter, op: OperationShape, operationStackName: String) {
+        writer.write("let apiMetadata = ${AWSClientRuntimeTypes.Core.APIMetadata}(serviceId: serviceName, version: \"${settings.moduleVersion}\")")
```
kneekey23, line 25: Nit: Since this is a type can we use $N here?

codegen/smithy-aws-swift-codegen/src/main/kotlin/software/amazon/smithy/aws/swift/codegen/middleware/UserAgentMiddleware.kt
```diff
@@ -22,10 +22,11 @@ class UserAgentMiddleware(val settings: SwiftSettings) : MiddlewareRenderable {
...
         writer.write("$operationStackName.${middlewareStep.stringValue()}.intercept(position: ${position.stringValue()}, middleware: \$N(${middlewareParamsString()}))", AWSClientRuntimeTypes.Core.UserAgentMiddleware)
     }
 
     private fun middlewareParamsString(): String {
-        return "metadata: ${AWSClientRuntimeTypes.Core.AWSUserAgentMetadata}.fromEnv(apiMetadata: ${AWSClientRuntimeTypes.Core.APIMetadata}(serviceId: serviceName, version: \"${settings.moduleVersion}\"))"
+        return "metadata: ${AWSClientRuntimeTypes.Core.AWSUserAgentMetadata}.fromEnv(apiMetadata: apiMetadata, frameworkMetadata: config.frameworkMetadata)"
```
kneekey23, line 30: Same nit as above. Can we use $N here for consistency?

kneekey23 (approved):
2 small nits. Fix n ship!

codegen/smithy-aws-swift-codegen/src/main/kotlin/software/amazon/smithy/aws/swift/codegen/middleware/UserAgentMiddleware.kt
wooj2: Hmm i don't think it is possible to do this here because this is returning a string rather than a call to `writer.write()`

kneekey23: Fair point. Leave this one alone. My bad for not seeing that

codegen/smithy-aws-swift-codegen/src/main/kotlin/software/amazon/smithy/aws/swift/codegen/middleware/UserAgentMiddleware.kt
wooj2: updated.

PR #499, opened 2021-11-18
chore: adding level of indirection for Bundle

wooj2:
Working around swiftlint instantiating `Bundle()` directly in this commit: https://github.com/awslabs/aws-sdk-swift/commit/4649f0c352f5110a92054de4066252101429578d#
- this eliminates the false positive around directly instantiating `Bundle()`
- Because it is a false positive, pushing this code in is not required, but if we feel it adds to readability, we should push this in.


kneekey23 (approved):
LGTM

PR #501, opened 2021-12-02
fix: changes to fix lifetime issues when multiple sdks are in use

kneekey23:
## Description of changes

Corresponding PR: https://github.com/awslabs/smithy-swift/pull/390
## Conventional Commits


justinboswell (approved):
I would strongly recommend putting this into the release, otherwise anyone who tries to use multiple services in the same application will get a crash or hang.

AWSClientRuntime/Sources/IMDS/IMDSClient.swift
```diff
@@ -13,8 +13,9 @@ import ClientRuntime
 /// This client supports fetching tokens, retrying failures, and token caching according to the specified TTL.
 /// NOTE: This client ONLY supports IMDSv2. It will not fallback to IMDSv1.
 /// See https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/configuring-instance-metadata-service.html#instance-metadata-transition-to-version-2 for more information.
-public class IMDSClient {
+public class IMDSClient{
```
wooj2, line 16: seems like this should fail lint.  can we run `bss lint` ?

wooj2 (approved):
assuming lint passes lgtm!

PR #508, opened 2021-12-15
fix: eliminate latency of instantiation of client on iOS with explicit region

wooj2:
Updated the way we determine region such that if a user passes in an explicit region, we do not instantiate the DefaultRegionResolver, and therefore avoid instantiating the IMDSRegionResolver.
```
                        if let region = region {
                            self.region = region
                            self.regionResolver = nil
                            self.signingRegion = signingRegion ?? region
                        } else {
                            let resolvedRegionResolver = regionResolver ?? DefaultRegionResolver()
                            let region = resolvedRegionResolver.resolveRegion()
                            self.region = region
                            self.regionResolver = resolvedRegionResolver
                            self.signingRegion = signingRegion ?? region
                        }
```                        


kneekey23 (approved):
LGTM

PR #509, opened 2021-12-15
chore: disable imds region resolver for ios, watchos and tvos

wooj2:

### DIFF
--- a/AWSClientRuntime/Sources/Regions/DefaultRegionResolver.swift
+++ b/AWSClientRuntime/Sources/Regions/DefaultRegionResolver.swift
@@ -13,7 +13,7 @@ public struct DefaultRegionResolver: RegionResolver {
 
     public init(providers: [RegionProvider]? = nil) {
         #if os(iOS) || os(watchOS) || os(tvOS)
-        self.providers = providers ?? [BundleRegionProvider(), EnvironmentRegionProvider(), ProfileRegionProvider()]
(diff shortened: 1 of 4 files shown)


AWSClientRuntime/Sources/Regions/DefaultRegionResolver.swift
```diff
@@ -11,8 +11,12 @@ public struct DefaultRegionResolver: RegionResolver {
     public let providers: [RegionProvider]
     let logger: SwiftLogger
 
-    public init(providers: [RegionProvider] = [BundleRegionProvider(), EnvironmentRegionProvider(), ProfileRegionProvider(), IMDSRegionProvider()]) {
-        self.providers = providers
+    public init(providers: [RegionProvider]? = nil) {
+        #if os(iOS) || os(watchOS) || os(tvOS)
+        self.providers = providers ?? [BundleRegionProvider(), EnvironmentRegionProvider(), ProfileRegionProvider()]
```
kneekey23, line 16: Don't think profile provider should be in here either? Don't think these devices will have anything saved at \.aws 

Thoughts?

kneekey23 (approved):
One question but otherwise it LGTM

AWSClientRuntime/Sources/Regions/DefaultRegionResolver.swift
wooj2: ah, yeah that makes sense

PR #523, opened 2022-01-11
chore: add script to facilitate local dev of generated SDK

wooj2:

### DIFF
--- a/scripts/updateReleaseToLocalDev.sh
+++ b/scripts/updateReleaseToLocalDev.sh
@@ -1,0 +1,84 @@
+#!/bin/bash
+
+# Use this script after pulling down a local copy of the AWS Swift SDK to your local machine.
+# Doing so will allow you to build (and modify) a local copy of the AWS Swift SDK.
+#
+# After setting this up, you can update your executable's Package.swift to depend on a local version of the SDK via:
+#
+#    dependencies: [
+#        .package(name: "AWSSwiftSDK", path:"~/Projects/SwiftSDK_Releases/aws-sdk-swift"),
+#    ],
+
+#
+# Setup Instructions
+#  Note: Versions listed below should be updated to your preference (ideally, the latest)
+#
+# mkdir -p ~/Projects/SwiftSDK_Releases
+# cd ~/Projects/SwiftSDK_Releases
+#
+# git clone git@github.com:awslabs/aws-sdk-swift.git
+# cd aws-sdk-swift
+# git checkout -b 0.1.1
+# cd ..
+#
+# git clone git@github.com:awslabs/smithy-swift.git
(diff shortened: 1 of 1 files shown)


scripts/updateReleaseToLocalDev.sh
```diff
@@ -0,0 +1,84 @@
...
+#    ],
+
+#
+# Setup Instructions
+#  Note: Versions listed below should be updated to your preference (ideally, the latest)
+#
+# mkdir -p ~/Projects/SwiftSDK_Releases
+# cd ~/Projects/SwiftSDK_Releases
+#
+# git clone git@github.com:awslabs/aws-sdk-swift.git
```
kneekey23, line 19: in a future version of this script couldn't we also run all these commands for you so we clone all 3 repos perhaps to a folder for you?

scripts/updateReleaseToLocalDev.sh
```diff
@@ -0,0 +1,84 @@
...
+    pushd ${SCRIPTDIR} > /dev/null
+    cd ../../
+    AWS_SDK_SWIFT=$PWD/aws-sdk-swift
+    SMITHY_SWIFT=$PWD/smithy-swift
+    AWS_CRT_SWIFT=$PWD/aws-crt-swift
+}
+
+determineAbsolutePath $0
+
+if [ -z ${AWS_SDK_SWIFT} ] || [ ! -d ${AWS_SDK_SWIFT} ]; then
```
kneekey23, line 54: is this checking for null or empty path here?

kneekey23 (approved):
LGTM, 1 comment, 1 question

scripts/updateReleaseToLocalDev.sh
wooj2: We are checking if the string is empty -- if it is empty, we bail.. if it isn't, then we check to make sure that the directory is there.  If the directory is not there, then we bail.

scripts/updateReleaseToLocalDev.sh
wooj2: The best we could do is clone 2 out of the 3 because we'd need to clone at least one repo (where the script lives)... that being said --  once a dev clones the repo, we need them to specify what version to check out.  So instead of creating an interface to abstract this -- I decided to simply, not. I feel developers need to have control over their sources/versions (and may change their versions in the future), so i figure, the setup process is part of their understanding to owning their dev env.

kneekey23: Got it, im thinking in the future when people just want to reproduce something with the sdk and dont necessarily want to set up a dev env, it could be useful to have something like this. but it's down the road, I was just brainstorming

PR #544, opened 2022-02-17
fix: test sdk function in order to fix streaming bug

kneekey23:
## Description of changes

This PR fixes our test code in order to debug issues with the new async code.

## Conventional Commits


TestLambdaSdk/Sources/TestSdk/TestSdk.swift
```diff
@@ -0,0 +1,30 @@
...
+            print("why is path empty")
+            fatalError()
+        }
+        guard let data = FileManager.default.contents(atPath: path) else {
+            print("why is data empty")
+            fatalError()
+        }
+       
+        let input = PutObjectInput(body: ByteStream.from(data: data),
+                                   bucket: "nickijustinsite", key: "4kimage.jpg")
```
Velfi, line 25: This test will break for anyone that isn't @kneekey23 right?

Velfi (approved):
I see there are linter errors, but I'll approve anyways. Fix them if you think it's important to do so.

rcoh (approved):
LGTM—this appears to only change the examples? Just wanted to make sure that was the intention

TestLambdaSdk/Sources/TestSdk/TestSdk.swift
kneekey23: lol yes but the idea is to just replace it with your own bucket and key for testing purposes. I'll add a comment.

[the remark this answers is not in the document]
kneekey23: this package is purely for sdk developers to test the local sdk as they are building and not for customers really. adding a note here to replace with bucket and key for next developer to test.

PR #554, opened 2022-03-01
fix: add pre sign urls for s3 put object

kneekey23:
## Description of changes

This adds pre sign urls for s3 put object.
Corresponding PR: https://github.com/awslabs/smithy-swift/pull/411

## Conventional Commits

### DIFF
--- a/codegen/smithy-aws-swift-codegen/src/main/kotlin/software/amazon/smithy/aws/swift/codegen/customization/presignable/PresignableUrlIntegration.kt
+++ b/codegen/smithy-aws-swift-codegen/src/main/kotlin/software/amazon/smithy/aws/swift/codegen/customization/presignable/PresignableUrlIntegration.kt
@@ -29,20 +29,22 @@ import software.amazon.smithy.swift.codegen.middleware.OperationMiddleware
 import software.amazon.smithy.swift.codegen.model.expectShape
 import software.amazon.smithy.swift.codegen.model.hasTrait
 
-internal val PRESIGNABLE_GET_OPERATIONS: Map<String, Set<String>> = mapOf(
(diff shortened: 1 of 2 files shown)


codegen/smithy-aws-swift-codegen/src/main/kotlin/software/amazon/smithy/aws/swift/codegen/customization/presignable/PresignableUrlIntegration.kt
```diff
@@ -191,7 +198,10 @@ class PresignableGetIntegration(private val presignedOperations: Map<String, Set
...
     }
 
-    private fun overrideHttpMethodWithGet(): String {
-        return "get"
+    private fun overrideHttpMethod(operation: OperationShape): String {
+        return when (operation.id.toString()) {
+            "com.amazonaws.s3#PutObject" -> "put"
+            else -> "get"
+        }
     }
```
rcoh, line 206: is the real edge case here polly? The HTTP method for all S3 operations is preserved AFAIK?

rcoh (approved):

codegen/smithy-aws-swift-codegen/src/main/kotlin/software/amazon/smithy/aws/swift/codegen/customization/presignable/PresignableUrlIntegration.kt
kneekey23: yes edge case is polly so this is basically just for polly I should change this to reflect that, good call out

kneekey23: actually requires changing a lot more code. leaving for now but its a good call out for the future.

PR #570, opened 2022-04-14
Revises generation script for Package.swift

brennanMKE:
## Issue \#

N/A

## Description of changes

The current `Package.swift` is quite large with each of the client SDKs listed as products and targets. These can be read from the `release` directory. The version details can also be placed into the new template with the generate script. This update makes it much simpler.

This update also moves the source files for `AWSClientRuntime` to the default location so that the `path` does not have to be set. Xcode was also not showing this code because that directory also contains it's own `Package.swift` and it causes it to be hidden in Xcode's file navigator panel. With these changes these source files are now accessible via search and the file navigator.

There is also support for reading in `filter.json` which is optional. If it is defined it will limit which clients are defined as targets and products in `Package.swift` which prevents Xcode from indexing any code which is not a part of a target. The result is an improved developer experience. Previously nearly 5 million lines of code would be indexed which slows down the Mac which has opened this package.

## Conventional Commits


brennanMKE:
A new branch from `main` is coming which will be pushed by @atierian so that the CI/CD tasks can run successfully.

PR #574, opened 2022-04-22
chore: package manifest local improvements

atierian:
## Description of changes

PR opened on behalf of @brennanMKE for changes made to the Package manifest generation that allow to define which targets are consumed when using the AWS SDK Swift locally.

## Conventional Commits


brennanMKE:
## Why?

While working in Xcode the files under AWSClientRuntime do not show up in the file navigator even though the folder and files are in this package. It was due to that folder also having a `Package.swift` in it. In the root folder the main `Package.swift` referenced the path where the code for AWSClientRuntime has been located.

For targets for Sources and Tests there is a default path where source files are normally placed. Since these files were not at the default path is was necessary to set `path` to point to the folder where Sources and Tests could be found. By moving files into the default folder structure it was possible to remove the `path` override in `./AWSClientRuntime/Package.swift`. Then symlinks were added to the root folder to point to `./AWSClientRuntime/Sources` and `./AWSClientRuntime/Tests` so the root package also did not need to use the `path` override. And this way these files now appear in Xcode's file navigator so they can be accessed like usual.

## Changes to root Package.swift

The script which generates the release version of Package.swift has been updated to use a template which is valid Swift source file which can be edited with Xcode with syntax coloring and code completion. It has placeholders where version numbers are inserted by the script. The generated Package.swift also does not list every target and product explicitly anymore since the `.release` folder can be read to get that list and then dynamically define every target and product. An optional `filter.json` can be placed in the root folder to filter which packages will be defined dynamically. This is done to limit what Xcode will index for use with code completion. Since there is nearly 5 million lines of code across all AWS clients it can put a considerable load on Xcode which will slow down the developer's computer. By defining a small set of targets it is possible to prevent Xcode from doing more work than is necessary.

aajtodd (approved):

PR #621, opened 2022-09-22
fix: Delete the `Package.swift` in `AWSClientRuntime/`

jbelkins:
## Description of changes
Delete the `Package.swift` in the `AWSClientRuntime/` subdirectory & move its function to the main `Package.swift`.

The `AWSClientRuntime` library is now available only as a product published by this package's main `Package.swift`.

This will simplify the package definitions in this project, and allow for editing of `AWSClientRuntime` source files from within Xcode.


jbelkins:
As a retest, I re-generated the `releases/` folder from this branch, then integrated the SDK into a CLI application.  It built and ran as expected.

codegen/Package.swift
```diff
@@ -70,7 +70,7 @@ func appendLibTarget(name: String, path: String) {
                     ),
                     .product(
                         name: "AWSClientRuntime",
-                        package: "AWSClientRuntime"
+                        package: "AWSSwiftSDK"
```
ganeshnj, line 73: this seems incorrect. Why AWSSwiftSDK is not defined as package.

codegen/Package.swift
```diff
@@ -99,11 +99,11 @@ if let smithySwiftDir = ProcessInfo.processInfo.environment["SMITHY_SWIFT_CI_DIR
    let sdkDir = ProcessInfo.processInfo.environment["AWS_SDK_SWIFT_CI_DIR"] {
     package.dependencies += [
         .package(name: "ClientRuntime", path: smithySwiftDir),
-        .package(name: "AWSClientRuntime", path: "\(sdkDir)/AWSClientRuntime"),
+        .package(name: "AWSSwiftSDK", path: sdkDir),
```
ganeshnj, line 102: same as above.

ganeshnj (changes requested):

codegen/Package.swift
jbelkins: This is indeed correct.

Instead of being a product published from its own package, i.e. package=`AWSClientRuntime` & product=`AWSClientRuntime`, `AWSClientRuntime` is now a product published from the `AWSSwiftSDK` package, i.e. package=`AWSSwiftSDK` and product=`AWSClientRuntime`.

This is similar to how the service clients are products published by the `AWSSwiftSDK` project.  The only difference here is that `AWSClientRuntime` is not auto-generated.

As proof that this configuration is correct, if you generate files and run `swift test` from within the `codegen/` directory, it resolves dependencies & runs tests successfully.

codegen/Package.swift
jbelkins: See my comment above, this package dependency is correct.

jbelkins:
@ganeshnj Please re-review in light of my comments

codegen/Package.swift
ganeshnj: awesome, thanks for verifying it.

ganeshnj (approved):

PR #625, opened 2022-09-27
fix: Use the IDLv2 ClientOptional trait instead of Box

sugmanue:
## Description of changes

Changed the way we convert all the members of a service to boxed. The way this was done will not work properly with IDLv2 which does no longer relay on the `Box` trait to determine whether or not a member is nullable. This change aligns with the pull request submitted to the swift-sdk [#444](https://github.com/awslabs/smithy-swift/pull/444) that makes the corresponding changes in the SDK itself.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -90,6 +90,9 @@ private extension Package {
         ]
 
         let sdksToIncludeInTargets = try! FileManager.default.contentsOfDirectory(atPath: localReleaseSwiftSDKDir.path)
+            .filter { target in
+                !target.hasPrefix(".DS_Store")
+            }
         includeTargets(package: self, releasedSDKs: sdksToIncludeInTargets)
         return self
     }
--- a/Package.swift
+++ b/Package.swift
@@ -91,7 +91,7 @@ private extension Package {
 
         let sdksToIncludeInTargets = try! FileManager.default.contentsOfDirectory(atPath: localReleaseSwiftSDKDir.path)
             .filter { target in
-                !target.hasPrefix(".DS_Store")
+                !target.hasPrefix(".")
             }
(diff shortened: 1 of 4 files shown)


Package.swift
```diff
@@ -90,6 +90,9 @@ private extension Package {
         ]
 
         let sdksToIncludeInTargets = try! FileManager.default.contentsOfDirectory(atPath: localReleaseSwiftSDKDir.path)
+            .filter { target in
+                !target.hasPrefix(".DS_Store")
```
jbelkins, line 94: Should this maybe be a helper method?  Also, more generally, we would likely want to exclude any dot file, not just `.DS_Store`

gradle.properties
```diff
@@ -1,7 +1,7 @@
 kotlin.code.style=official
 
 # codegen
-smithyVersion=1.22.0
+smithyVersion=[1.25.0,1.26.0[
```
jbelkins, line 4: Is this correct to have two opening brackets?

ganeshnj: Yes, this is correct.

ganeshnj: https://docs.gradle.org/current/userguide/single_versions.html

Package.swift
ganeshnj: Although not need but fair to include all files with `.` prefix.

ganeshnj: in a Package.swift file another helper methods seem overkill.

ganeshnj (approved):

gradle.properties
jbelkins: lol I need to get smart on Gradle

jbelkins (approved):

Issue #624, opened 2022-09-26
endpoints 2.0


ganeshnj:
## Issues

Fixes https://github.com/awslabs/aws-sdk-swift/issues/533
Fixes https://github.com/awslabs/aws-sdk-swift/issues/346

## Tasks

codegen

- [x] https://github.com/awslabs/aws-sdk-swift/pull/607
- [x] https://github.com/awslabs/smithy-swift/pull/433

CRT integration

- [ ] https://github.com/awslabs/smithy-swift/pull/446
- [ ] https://github.com/awslabs/aws-crt-swift/pull/79
- [ ] https://github.com/awslabs/aws-sdk-swift/pull/634

- [x] I may be able to implement this feature request

PR #634, opened 2022-10-05
feat: implement DefaultEndpointResolver using CRT

ganeshnj:
## Issue \#
https://github.com/awslabs/aws-sdk-swift/issues/624

## Description of changes

This changes takes https://github.com/awslabs/aws-crt-swift/pull/79 Swift bindings and create wrappers for it in AWS Client Runtime. These wrappers are used in the service code which is codegen.

## Conventional Commits

### DIFF
--- a/codegen/smithy-aws-swift-codegen/src/main/kotlin/software/amazon/smithy/aws/swift/codegen/AWSServiceConfig.kt
+++ b/codegen/smithy-aws-swift-codegen/src/main/kotlin/software/amazon/smithy/aws/swift/codegen/AWSServiceConfig.kt
@@ -79,7 +79,11 @@ class AWSServiceConfig(writer: SwiftWriter, val ctx: ProtocolGenerator.Generatio
                 }
 
                 ENDPOINT_RESOLVER -> {
-                    writer.write("self.${it.memberName} = ${it.memberName} ?? DefaultEndpointResolver()")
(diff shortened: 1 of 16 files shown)


codegen/smithy-aws-swift-codegen/src/main/kotlin/software/amazon/smithy/aws/swift/codegen/AWSServiceConfig.kt
```diff
@@ -174,7 +178,11 @@ class AWSServiceConfig(writer: SwiftWriter, val ctx: ProtocolGenerator.Generatio
                 }
 
                 ENDPOINT_RESOLVER -> {
-                    writer.write("self.${it.memberName} = ${it.memberName} ?? DefaultEndpointResolver()")
+                    writer.openBlock("if let endpointResolver = endpointResolver {", "} else {") {
+                        writer.write("self.endpointResolver = endpointResolver")
+                    }
+                    writer.indent().write("self.${it.memberName} = try \$L()", AWSServiceTypes.DefaultEndpointResolver)
+                    writer.dedent().write("}")
```
jbelkins, line 185: No need to address it here, but using indent then dedent is an awkward, potentially error-prone way of writing an if-else statement... wonder if there should be a method on writer for a two-block statement

jbelkins (approved):
Just one general question

PR #635, opened 2022-10-06
fix: pin smithy and crt versions in 0.2.6

atierian:
## Issue \#

N/A

## Description of changes

The SDK uses the `from` dependency version requirement for `ClientRuntime` and `CRT`.  This means that any consumer Package or application that depends on a specific version of the AWS SDK for Swift will have their transitive dependencies (`ClientRuntime` and `CRT`) updated with any new release. This is not ideal - it can lead to the improvements in those dependencies causing breaking behavioral changes for consumers. 

One such [improvement](https://github.com/awslabs/smithy-swift/pull/442) ended up breaking workarounds we had in place. While these changes are obviously very welcome, they shouldn't be forced on consumers that pin to a specific version of the AWS SDK for Swift.

We looked into fixing this ourselves here https://github.com/aws-amplify/amplify-ios/pull/2397. 
TL;DR pinning in a consuming package works when adding locally, but not remotely.

This PR changes `from` to `exact` for `release/0.2.6`; this unblocks us because we're pinning to it. Once merged, we'd need to update the `0.2.6` tag to point at the new commit. A separate PR should follow up with the necessary changes to `generatePackageSwift.swift` to ensure dependency versions are pinned in all releases going forward.

N/A

## Conventional Commits

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -320,8 +320,8 @@ let package = Package(
         .library(name: "AWSXRay", targets: ["AWSXRay"]),
     ],
     dependencies: [
-        .package(name: "AwsCrt", url: "https://github.com/awslabs/aws-crt-swift.git", from: "0.2.2"),
-        .package(name: "ClientRuntime", url: "https://github.com/awslabs/smithy-swift.git", from: "0.2.5")
+      .package(name: "AwsCrt", url: "https://github.com/awslabs/aws-crt-swift.git", .exact("0.2.2")),
+      .package(name: "ClientRuntime", url: "https://github.com/awslabs/smithy-swift.git", .exact("0.2.5"))
     ],
     targets: [
         .target(


ganeshnj:
When new features are added to Service specific packages, sometimes we need to add support for these features in Client Runtimes (AWS Client Runtime or Smithy Client Runtime). Hence, from is recommended way to go then pinning to exact version. I understand, we don't follow semver and introducing breaking change using minor version bump and this will be case until we go GA.

https://docs.swift.org/package-manager/PackageDescription/PackageDescription.html
>/// Specifying exact version requirements are not recommended as
/// they can cause conflicts in your dependency graph when multiple other packages depend on a package.
/// As Swift packages follow the semantic versioning convention,
/// think about specifying a version range instead.

Specially this is not going to work when we go GA and single service versions can be updated.

ganeshnj (approved):

PR #637, opened 2022-10-07
fix: pin smithy and crt versions in 0.3.0

ganeshnj:
## Issue \#

Followup: https://github.com/awslabs/aws-sdk-swift/pull/635 

## Conventional Commits


jbelkins (approved):

ganeshnj:
Cancelled the workflow as it adds the generated  changes again which we don't want to do. Because Models.swift file exists already.

Ran the local build, all looks good. Merging.

PR #647, opened 2022-10-13
fix: Update package names in `Package.swift` to match modern Swift convention

jbelkins:
## Description of changes
In [SE-0226](https://github.com/apple/swift-evolution/blob/main/proposals/0226-package-manager-target-based-dep-resolution.md), declaring the name of Swift packages in a package's manifest was deprecated.  Instead, the name of a package is inferred from the last path component of the package's URL.  (The purpose of this was to speed & simplify the resolution of dependencies by allowing the resolution of a target's dependencies without having to clone all its sub-dependencies to see what package contains what name.)

The latest version of Xcode relies on this naming convention for certain features, including package edit mode.  Specifically, the IDE will not properly resolve dependencies when packages are referred to by specific names that do not match the inferred names.

This PR changes the package manifest to no longer use explicit package names, relying instead on the last component of the URL or local path where the package resides.  This allows Xcode to successfully resolve all dependencies with packages in edit mode.

Also:
- Update `Package.swift` to adopt `swift-tools-version:5.5` (from 5.4).  5.5 is already required by `aws-crt-swift` so it is required to build the AWS SDK anyway.
- `Package.swift` code cleanup, including renaming some variables to be more descriptive of their contents and extracting package names to variables.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -1,4 +1,4 @@
-// swift-tools-version:5.4
+// swift-tools-version:5.5
 
 /*
  * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
--- a/Package.swift
+++ b/Package.swift
@@ -85,8 +85,8 @@ private extension Package {
 
     func setupDependencies() -> Package {
         dependencies += [
-            .package(name: "AwsCrt", path: awsCRTSwiftDir.path),
-            .package(name: "ClientRuntime", path: smithySwiftDir.path)
+            .package(name: "aws-crt-swift", path: awsCRTSwiftDir.path),
+            .package(name: "smithy-swift", path: smithySwiftDir.path)
         ]
 
         let sdksToIncludeInTargets = try! FileManager.default.contentsOfDirectory(atPath: localReleaseSwiftSDKDir.path)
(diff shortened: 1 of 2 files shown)


jbelkins:
This seems to build OK on newer toolchains but not on older ones.  Going to move this back to a draft while I figure out what's going on.

jbelkins:
The problem that was failing CI is fixed.

(Issue was that we were using explicit names in dependency declarations that did not match inferred names.  This seems not to matter on latest Swift, but earlier versions would fail dependency resolution on this.  Removing all explicit naming of package dependencies from this manifest corrects the problem on all Swift versions.)

ganeshnj:
Do you need to update https://github.com/awslabs/smithy-swift/blob/main/smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/PackageManifestGenerator.kt ?

How does the generated package manifest look like? We want handwritten and generated manifest to be same. Could you add to the PR description for posterity.

jbelkins:
> Do you need to update https://github.com/awslabs/smithy-swift/blob/main/smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/PackageManifestGenerator.kt ?

Considering that package manifests written by the Kotlin you cite are written with absolute file paths on the development machine, they are of very limited use anyway.  I would submit that if we want to keep those manifests around, they should be written so that they are portable, and I'd consider that to be outside the scope of this PR.

In any case, they should continue to work even with changes in this PR; other packages are free to continue referring to this package by its `name` property.

jbelkins:
Closing this PR since it will be rolled into #650.

PR #660, opened 2022-10-31
fix: prefer EndpointResolver Bucket for S3 in http label, AccountId for S3Control host label

ganeshnj:
### S3 Bucket deduplication

`Bucket` in the HTTP URI is now a `contextParam` hence rules engine takes care of adding the bucket to the URI to either host name or path depending on the bucket location `forceStylePath` configuration. Therefore, SDK must not add the bucket to the URI outside of the rules engine.

```json
"com.amazonaws.s3#GetObject": {
    "type": "operation",
    "input": {
        "target": "com.amazonaws.s3#GetObjectRequest"
    },
    "output": {
        "target": "com.amazonaws.s3#GetObjectOutput"
    },
    "traits": {
        "aws.protocols#httpChecksum": {
            "requestValidationModeMember": "ChecksumMode",
            "responseAlgorithms": [
                "CRC32",
                "CRC32C",
                "SHA256",
                "SHA1"
            ]
        },
        "smithy.api#http": {
            "method": "GET",
            "uri": "/{Bucket}/{Key+}?x-id=GetObject",
            "code": 200
        }
    }
},
```

```json
"com.amazonaws.s3#GetObjectRequest": {
    "type": "structure",
    "members": {
        "Bucket": {
            "target": "com.amazonaws.s3#BucketName",
            "traits": {
                "smithy.api#httpLabel": {},
                "smithy.api#required": {},
                "smithy.rules#contextParam": {
                    "name": "Bucket"
                }
            }
        },
```

### S3 Control AccountId deduplication

`AccountId` hostPrefix is now a `staticContextParams` hence rules engine takes care of adding `AccountId` to the host name. This is enabled by setting `RequiresAccountId` param in staticContextParams to `true`. Therefore, SDK must not add the `AccountId` to the host name outside of the rules engine.

```json
"com.amazonaws.s3control#CreateAccessPoint": {
    "type": "operation",
    "input": {
        "target": "com.amazonaws.s3control#CreateAccessPointRequest"
    },
    "output": {
        "target": "com.amazonaws.s3control#CreateAccessPointResult"
    },
    "traits": {
        "smithy.api#endpoint": {
            "hostPrefix": "{AccountId}."
        },
        "smithy.api#http": {
            "method": "PUT",
            "uri": "/v20180820/accesspoint/{Name}",
            "code": 200
        },
        "smithy.rules#staticContextParams": {
            "RequiresAccountId": {
                "value": true
            }
        }
    }
}
```

## Conventional Commits

### DIFF
--- a/codegen/smithy-aws-swift-codegen/src/main/kotlin/software/amazon/smithy/aws/swift/codegen/model/AWSEndpointTraitTransformer.kt
+++ b/codegen/smithy-aws-swift-codegen/src/main/kotlin/software/amazon/smithy/aws/swift/codegen/model/AWSEndpointTraitTransformer.kt
@@ -36,10 +36,11 @@ class AWSEndpointTraitTransformer : SwiftIntegration {
                                     val inputShape = model.expectShape(input)
                                     val members = inputShape.members() ?: emptyList()
                                     members.forEach { member ->
-                                        if (member.hasTrait<ContextParamTrait>()
-                                            && member.hasTrait<HostLabelTrait>()
-                                            && member.memberName == "AccountId"
-                                            && hostPrefix == "{AccountId}.") {
+                                        if (member.hasTrait<ContextParamTrait>() &&
+                                            member.hasTrait<HostLabelTrait>() &&
+                                            member.memberName == "AccountId" &&
+                                            hostPrefix == "{AccountId}."
+                                        ) {
                                             shapeBuilder.removeTrait(endpointTrait.toShapeId())
                                         }
                                     }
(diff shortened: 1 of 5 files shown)


codegen/smithy-aws-swift-codegen/src/main/kotlin/software/amazon/smithy/aws/swift/codegen/model/AWSEndpointTraitTransformer.kt
```diff
@@ -0,0 +1,56 @@
...
+import software.amazon.smithy.swift.codegen.SwiftSettings
+import software.amazon.smithy.swift.codegen.integration.SwiftIntegration
+import software.amazon.smithy.swift.codegen.model.getTrait
+
+/**
+ * This integration is responsible for removing the EndpointTrait from the operation
+ * - For S3 Control, if hostPrefix is {AccountId}., then remove the Endpoint Trait because it is already handled
+ *      within the EndpointResolver
+ */
+class AWSEndpointTraitTransformer : SwiftIntegration {
```
jbelkins, line 22: This and the other trait transformer have a lot of pretty deeply nested logic... should we break down the logic a bit more?

jbelkins (approved):
Approved, but with a question

codegen/smithy-aws-swift-codegen/src/main/kotlin/software/amazon/smithy/aws/swift/codegen/model/AWSEndpointTraitTransformer.kt
ganeshnj: I wouldn't because it is very specific to case, breaking it would make understanding it harder than at just one place.

PR #678, opened 2022-11-03
fix!: Rename package to match name inferred from git URL

jbelkins:
## Issue \#
Fixes https://github.com/awslabs/aws-sdk-swift/issues/680

## Description of changes

Since the introduction of SPM, a package manifest has included a field for providing a "declared" name (i.e. set in the `name` property of the `Package` structure) that is used in other manifests to identify a package as a dependency.

In [SE-0226], Swift package naming was altered to derive the name from the Git URL for the package.  While the preferred way to identify a dependency is by the inferred name, `Package` retained the declared name field to provide backward compatibility with dependencies declared before this improvement.  (Note that referring to a package by its declared name is now deprecated, but use of deprecated APIs in a package manifest does not throw warnings.)

The rationale for this change is that it allows the package manager to resolve what dependencies it needs for a specific target without resorting to downloading every dependency referenced in a manifest to see what name is declared in each dependency's own manifest.  In our case, the name that is inferred from this package's URL is `aws-sdk-swift` (essentially the last path component with extensions dropped).

Most other Swift packages have moved to (or been created with) a declared package name that matches the inferred name.  While latest Swift tools allow a manifest to identify a package by either name, consistent naming brings the following advantages:
  - Consumers who move to the inferred name will get the advantage of faster dependency resolution as promised by [SE-0226].
  - Certain common tools, such as `swift package edit` with the `--path` extension (which is used by Swift-VSCode's GUI package management tools), break when the declared name and the inferred name do not match.

**This will cause a breaking change** but the update to requiring package manifests will be trivial; package manifests need only require this package by the new/inferred name.  I do not expect our Git URL to change, so this name will be permanent unless SPM changes naming conventions again.

Tested with:
- Corresponding name changes in [`smithy-swift`](https://github.com/awslabs/smithy-swift/pull/464) and [`aws-crt-swift`](https://github.com/awslabs/aws-crt-swift/pull/93).
- A successful release build (see link [here](https://github.com/awslabs/aws-sdk-swift/actions/runs/3396956679/jobs/5648595852).)
- Successfully built/tested as a downstream by `smithy-swift` (se PR above).

Specific changes:
- Declared package name updated to match the inferred name
- `codegen/Package.swift` updated to use inferred naming
- `scripts/generatePackageSwift.swift` updated to render manifest with inferred naming
- Fix a few generated `import` statements which were importing by package name (which worked by coincidence until this name change)

[SE-0226]: https://github.com/apple/swift-evolution/blob/main/proposals/0226-package-manager-target-based-dep-resolution.md


epau (commented):
lgtm deferring to @ganeshnj though since this is a breaking change

ganeshnj:
When we do the release with this change, we must create a discussion with code snippet to show how to migrate from old package name to `aws-sdk-swift`.

jbelkins:
> When we do the release with this change, we must create a discussion with code snippet to show how to migrate from old package name to `aws-sdk-swift`.

You mean in the release notes for the release?

.github/workflows/release.yml
```diff
@@ -4,6 +4,7 @@ on:
   push:
     branches:
         - 'release/**'
+        - 'jbe/package_rename'
```
jbelkins, line 7: This is a temporary change to test release building.  Will remove before merge.

jbelkins: Removed in [ec470f4](https://github.com/awslabs/aws-sdk-swift/pull/678/commits/ec470f4180d897687bd00c9ab20ddea30e343362)

ganeshnj (approved):

PR #684, opened 2022-11-08
fix!: update shape and member extensions to use same casing as rest of the system

ganeshnj:
### Motivation

While doing testing figured due to casing change UseFIPS is not translate to useFIPS by preserving the captalization for `FIPS` which is expected. This change broke the build because the config is named as `useFips`.

### Changes

- Match FIPS config name
- Some related changes from https://github.com/awslabs/smithy-swift/pull/467

### Result

Code compiles

## Conventional Commits


jbelkins:
Should we build a release on this branch as a retest?

jbelkins (approved):

epau (approved):

PR #709, opened 2022-11-11
fix! Converts Expires members in S3 output structures to String shape

### LINKED ISSUE #200: S3 Expires top level output member should be a string instead timestamp
aws/aws-sdk-go@5bdeab7

*internal id*: 176722195

epau:
## Issue \#
fixes #200 

## Description of changes
This PR adds a customization during codegen, that will convert members to String shape. 
This is restricted to members named "Expires" on "Output" structures in the S3 service.

## Conventional Commits

### DIFF
--- a/codegen/smithy-aws-swift-codegen/src/main/kotlin/software/amazon/smithy/aws/swift/codegen/customization/s3/S3Expires.kt
+++ b/codegen/smithy-aws-swift-codegen/src/main/kotlin/software/amazon/smithy/aws/swift/codegen/customization/s3/S3Expires.kt
@@ -9,8 +9,6 @@ import software.amazon.smithy.swift.codegen.model.defaultName
 import software.amazon.smithy.swift.codegen.model.expectShape
 
 class S3Expires : SwiftIntegration {
-    override val order: Byte
-        get() = 127
 
     override fun enabledForService(model: Model, settings: SwiftSettings): Boolean {
         return model.expectShape<ServiceShape>(settings.service).isS3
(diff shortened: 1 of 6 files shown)


codegen/smithy-aws-swift-codegen/src/main/kotlin/software/amazon/smithy/aws/swift/codegen/customization/s3/S3Expires.kt
```diff
@@ -0,0 +1,31 @@
...
+
+    override fun enabledForService(model: Model, settings: SwiftSettings): Boolean {
+        return model.expectShape<ServiceShape>(settings.service).isS3
+    }
+
+    override fun preprocessModel(model: Model, settings: SwiftSettings): Model {
+        // Find all the members named "Expires" in all of the output structures
+        // and change their shape from `Timestamp` to `String`
+        val updates = model.structureShapes
+            .filter { it.defaultName().endsWith("Output") }
```
epau, line 21: So this is indeed the suffix we want to check for, but this currently is failing for tests (although it works as we want when generating the models). All output models are given a `Response` suffix but the timing (and perhaps other factors) are different when executed for tests. I need to dig in to what exactly is going on in our test flow to get tests passing.

codegen/smithy-aws-swift-codegen/src/test/kotlin/software/amazon/smithy/aws/swift/codegen/TestContextGenerator.kt
```diff
@@ -36,7 +36,6 @@ class TestContextGenerator {
             val service = model.getShape(ShapeId.from(serviceShapeIdWithNamespace)).get().asServiceShape().get()
             val settings = buildDefaultSwiftSettingsObjectNode(serviceShapeIdWithNamespace)
             val swiftSettings = SwiftSettings.from(model, settings)
-            model = AddOperationShapes.execute(model, swiftSettings.getService(model), swiftSettings.moduleName)
```
epau, line 39: This is not necessary and is inconsistent with how this is performed in release.
By calling this here, we end up adding operation shapes before any of the integrations are executed. In release, we execute the integrations first, then add our operation shapes. By removing this line, we make our testing environment more consistent with our releases.

This logic is perform within `SwiftCodegenPlugin().execute(pluginContext)` which occurs a few lines below this.

jbelkins (approved):

PR #719, opened 2022-11-18
fix: fix S3 presigning, centralize generation of signature params

### LINKED ISSUE #718: Set `useDoubleURIEncode` to `false` for S3 SigV4Config
### Describe the bug

We are facing an issue where the `presignURL` captured from `GetObjectInput` , `PutObjectInput`  and `UploadPartInput` has `useDoubleURIEncode` set to `true` which fails to upload to S3.

### Expected Behavior

The presignURL captured from the S3 Input objects should result in successful uploads.

### Current Behavior

The presignURL results in a not authorized 403 error. 
```
The request signature we calculated does not match the signature you provided. Check your key and signing method."), _requestID: Optional("DKGVF6G0WJQ2ETGD"), _retryable: false, _type: ClientRuntime.ErrorType.unknow
```

### Reproduction Steps

Create a presignedURL form the input and use that to upload to S3 using URLSession. 
```
let config: S3ClientConfigurationProtocol = // your configuration
let input = GetObjectInput(bucket: bucket, key: key)
let preSignedUrl = try await input.presignURL(config: config, expiration: expiration)

```

### Possible Solution

Set `useDoubleURIEncode` set to `false` for S3 during SigV4 creation.

### Additional Information/Context

_No response_

### AWS SWIFT SDK version used

0.5.0

### Compiler and Version used

Xcode 14.1

### Operating System and version

iOS 16.1

jbelkins:
## Issue \#
Fixes #718

## Description of changes

When preparing the canonical request for AWS v4 code signing, the URI is to be normalized and double URI-encoded, unless the service is S3, in which case the URI is not normalized and is single-encoded.  See https://docs.aws.amazon.com/general/latest/gr/sigv4-create-canonical-request.html step 2.

When using the `presign` and `presignURL` methods on the S3 operations it is enabled for, the signed URL fails because the signature is calculated with double URL-encoding.  This PR ensures that the signature for any request with S3 as the service is calculated using single URI-encoding and that the URI is not normalized, whether pre-signing or performing the request.  In all other respects, signature configuration should be unchanged.

In order to deduplicate and simplify signing logic, the following refactors have been performed:
- `AWSSigningMiddleware` is now configured with a set of parameters that includes the service & operation.  These are used to determine the proper config for the Swift request signer middleware.
- Methods for setting Swift parameter strings are removed from various middleware & Swift integrations.  Instead, they configure the signer middleware as appropriate.
- There were two `SwiftIntegration`s that were overwriting previous signing settings.  The logic in these has been removed and integrated into the signing middleware, because there is the possibility that a `SwiftIntegration` could overwrite previous config that was not meant to be overwritten.
  - The `S3SigningConfig` Swift integration is removed, since its logic is now provided entirely by `AWSSigningMiddleware`.
  - The  `GlacierChecksum` Swift integration has its signature logic removed, since it is now in `AWSSigningMiddleware`. (It still provides Glacier-specific tree hash middleware, which is separate from the AWSv4 signature logic.)
- Logic for double- or single- encoding and for URI normalization are moved into `AWSSigningMiddleware` and are based on the identity of the service (i.e. is it S3 or not.)

Test coverage for presigning:
- Existing tests are all passing.
- Tests have been added for S3-specific logic when generating a request presigner.
- Tests have been created for URL presigning of all three supported operations.
- I have operationally tested all the URLs to verify they work.

Additionally, it was discovered that the `sdkId` that is defined in the `@service` trait is not copied into the `SwiftSettings` when a test context is created.  This was preventing the tests for service- and operation- specific logic from being tested successfully.  So, the following were done to correct it:
- `sdkId` is now set on test context `SwiftSettings` based on the Smithy definition.
- A couple parameters were removed from the test context constructor method, because they were unused and misleading.
- Several tests were fixed so expected values reflect the `sdkId` in the Smithy file for that test.


codegen/smithy-aws-swift-codegen/src/test/kotlin/software/amazon/smithy/aws/swift/codegen/TestContextGenerator.kt
```diff
@@ -68,11 +68,13 @@ class TestContextGenerator {
 
         fun buildDefaultSwiftSettingsObjectNode(
             serviceShapeId: String,
+            sdkId: String = "ExampleSDK",
             moduleName: String = "Example",
             moduleVersion: String = "1.0.0"
         ): ObjectNode {
             return Node.objectNodeBuilder()
                 .withMember("service", Node.from(serviceShapeId))
+                .withMember("sdkId", Node.from(sdkId))
```
jbelkins, line 77: This wasn't being set, so the test context was falling back to the service name

codegen/smithy-aws-swift-codegen/src/test/kotlin/software/amazon/smithy/aws/swift/codegen/TestContextGenerator.kt
```diff
@@ -28,13 +28,13 @@ data class TestContext(
 
 class TestContextGenerator {
     companion object {
-        fun initContextFrom(modelFile: String, serviceShapeIdWithNamespace: String, protocol: ShapeId, rootPackageNamespace: String, sdkId: String): TestContext {
+        fun initContextFrom(modelFile: String, serviceShapeIdWithNamespace: String, protocol: ShapeId): TestContext {
```
jbelkins, line 31: `rootPackageName` was removed as a param because it was unused.
`sdkId` was removed as a param because it was unused, and because it can be retrieved from the service instead.

codegen/smithy-aws-swift-codegen/src/test/kotlin/software/amazon/smithy/aws/swift/codegen/AWSSigningMiddlewareTests.kt
```diff
@@ -73,122 +76,164 @@ class AWSSigningMiddlewareTests {
     }
 
     @Test
-    fun `renderSigningMiddleware unsignedBody true`() {
```
jbelkins, line 76: This and the other tests below were refactored to get a lot of duplicate logic out of them.

codegen/smithy-aws-swift-codegen/src/test/kotlin/software/amazon/smithy/aws/swift/codegen/awsquery/AWSQueryOperationStackTest.kt
```diff
@@ -17,7 +17,7 @@ class AWSQueryOperationStackTest {
     @Test
     fun `operation stack has required middlewares`() {
         val context = setupTests("awsquery/query-empty-input-output.smithy", "aws.protocoltests.query#AwsQuery")
-        val contents = getFileContents(context.manifest, "/Example/AwsQueryClient.swift")
+        val contents = getFileContents(context.manifest, "/Example/QueryProtocolClient.swift")
```
jbelkins, line 20: Filename changed because the `sdkId` is now correctly set on the test context.

codegen/smithy-aws-swift-codegen/src/test/kotlin/software/amazon/smithy/aws/swift/codegen/ec2query/Ec2QueryHttpResponseBindingErrorGeneratorTests.kt
```diff
@@ -101,11 +101,11 @@ class Ec2QueryHttpResponseBindingErrorGeneratorTests {
                 public var _retryable: Swift.Bool = false
                 public var _isThrottling: Swift.Bool = false
                 public var _type: ClientRuntime.ErrorType = .client
-                public var nested: AwsEc2ClientTypes.ComplexNestedErrorData?
+                public var nested: EC2ProtocolClientTypes.ComplexNestedErrorData?
                 public var topLevel: Swift.String?
             
                 public init (
-                    nested: AwsEc2ClientTypes.ComplexNestedErrorData? = nil,
+                    nested: EC2ProtocolClientTypes.ComplexNestedErrorData? = nil,
```
jbelkins, line 108: Name of this type changed because it matches the `sdkId` in this test's Smithy def.

codegen/smithy-aws-swift-codegen/src/test/kotlin/software/amazon/smithy/aws/swift/codegen/restxml/AWSRestXMLHttpResponseBindingErrorGeneratorTests.kt
```diff
@@ -106,12 +106,12 @@ class AWSRestXMLHttpResponseBindingErrorGeneratorTests {
                 public var _isThrottling: Swift.Bool = false
                 public var _type: ClientRuntime.ErrorType = .client
                 public var header: Swift.String?
-                public var nested: RestXmlClientTypes.ComplexXMLNestedErrorData?
+                public var nested: RestXmlerrorsClientTypes.ComplexXMLNestedErrorData?
```
jbelkins, line 109: Name of this type changed because it matches the `sdkId` in this test's Smithy def.

codegen/smithy-aws-swift-codegen/src/test/kotlin/software/amazon/smithy/aws/swift/codegen/restxml/serde/S3UnwrappedXMLOutputTraitTests.kt
```diff
@@ -23,7 +23,7 @@ class S3UnwrappedXMLOutputTraitTests {
             
                 public init (from decoder: Swift.Decoder) throws {
                     var containerValues = try decoder.unkeyedContainer()
-                    let locationConstraintDecoded = try containerValues.decodeIfPresent(RestXmlClientTypes.BucketLocationConstraint.self)
+                    let locationConstraintDecoded = try containerValues.decodeIfPresent(S3ClientTypes.BucketLocationConstraint.self)
```
jbelkins, line 26: Name of this type changed because it matches the `sdkId` in this test's Smithy def.

codegen/smithy-aws-swift-codegen/src/test/kotlin/software/amazon/smithy/aws/swift/codegen/PresignerGeneratorTests.kt
```diff
@@ -21,7 +21,7 @@ class PresignerGeneratorTests {
             
             extension GetFooInput {
                 public func presign(config: ExampleClientConfigurationProtocol, expiration: Swift.Int64) async throws -> ClientRuntime.SdkHttpRequest? {
-                    let serviceName = "Example"
+                    let serviceName = "example"
```
jbelkins, line 24: Name of this service changed because it matches the `sdkId` in this test's Smithy def.

codegen/smithy-aws-swift-codegen/src/main/kotlin/software/amazon/smithy/aws/swift/codegen/middleware/AWSSigningMiddleware.kt
```diff
@@ -57,12 +65,22 @@ open class AWSSigningMiddleware(
     }
 
     private fun middlewareParamsString(op: OperationShape): String {
```
jbelkins, line 67: This is now the single place that parameters to create a Swift `SigV4Config` are generated, based on params that are passed into this middleware.

codegen/smithy-aws-swift-codegen/src/main/kotlin/software/amazon/smithy/aws/swift/codegen/middleware/AWSSigningMiddleware.kt
```diff
@@ -22,13 +23,20 @@ import software.amazon.smithy.swift.codegen.middleware.MiddlewareRenderable
 import software.amazon.smithy.swift.codegen.middleware.MiddlewareStep
 import software.amazon.smithy.swift.codegen.model.expectTrait
 import software.amazon.smithy.swift.codegen.model.hasTrait
+import java.util.Locale
 
-typealias AWSSigningMiddlewareParamsCallback = (OperationShape) -> String
+data class AWSSigningParams(
+    val useSignatureTypeQueryString: Boolean = false,
+    val forceUnsignedBody: Boolean = false,
+    val signedBodyHeaderContentSHA256: Boolean = false,
```
epau, line 31: Nit: Can we name this `signBodyHeaderContentSHA256` to keep the tense consistent with the other properties? 
```suggestion
    val signBodyHeaderContentSHA256: Boolean = false,
```

Additionally, can we add comments for each of these parameters that describe what they control?

codegen/smithy-aws-swift-codegen/src/main/kotlin/software/amazon/smithy/aws/swift/codegen/middleware/AWSSigningMiddleware.kt
```diff
@@ -57,12 +65,22 @@ open class AWSSigningMiddleware(
...
+
+        // Create param strings for each setting, or null for default param
+        val params = listOf(
+            "signatureType: .requestQueryParams".takeIf { params.useSignatureTypeQueryString },
+            "useDoubleURIEncode: false".takeIf { serviceIsS3 },
+            "shouldNormalizeURIPath: false".takeIf { serviceIsS3 },
+            "expiration: expiration".takeIf { params.useExpiration },
+            "signedBodyHeader: .contentSha256".takeIf { useSignedBodyHeader },
+            "unsignedBody: " + ("true".takeIf { useUnsignedPayload } ?: "false")
+        )
```
epau, line 81: Can we add properties to `AWSSigningParams` for all the values here? Such as `useDoubleURIEncode` `shouldNormalizeURIPath`, and `useSignedBodyHeader`. 
Then can we add an initializer to `AWSSigningParams` that takes a service as an input and performs the service specific logic to configure the parameters?

This makes this more testable as this chunk of code simply converts `AWSSigningParams` to a string. 
And the various initializers on `AWSSigningParams` are responsible for the setting the properties correctly according to the service. This makes things a bit easier to change and test in the future as we add/remove service specific handlings.

codegen/smithy-aws-swift-codegen/src/main/kotlin/software/amazon/smithy/aws/swift/codegen/middleware/AWSSigningMiddleware.kt
epau: i see now that these names mimic the values in the output string. In that case, ignore the naming change but comments would help.

codegen/smithy-aws-swift-codegen/src/main/kotlin/software/amazon/smithy/aws/swift/codegen/customization/presignable/PresignableUrlIntegration.kt
```diff
@@ -150,7 +144,14 @@ class PresignableUrlIntegration(private val presignedOperations: Map<String, Set
         operationMiddlewareCopy.removeMiddleware(op, MiddlewareStep.SERIALIZESTEP, "OperationInputHeadersMiddleware")
         operationMiddlewareCopy.removeMiddleware(op, MiddlewareStep.FINALIZESTEP, "ContentLengthMiddleware")
         operationMiddlewareCopy.removeMiddleware(op, MiddlewareStep.FINALIZESTEP, "AWSSigningMiddleware")
-        operationMiddlewareCopy.appendMiddleware(op, AWSSigningMiddleware(::customSigningParameters, context.model, context.symbolProvider))
+        val opID = op.id.toString()
+        val params = AWSSigningParams(
+            useSignatureTypeQueryString = true,
+            forceUnsignedBody = opID == "com.amazonaws.s3#PutObject" || opID == "com.amazonaws.s3#GetObject",
```
epau, line 150: A comment describing why this logic is needed here would be super helpful (since it's not obvious why its here and not also included in the other AWSSigningParams instances), especially for our future selves.

epau (changes requested):
This is looking really good! Thanks for tracking this down and I very much appreciate the detailed PR description. 

Just a few small asks

codegen/smithy-aws-swift-codegen/src/main/kotlin/software/amazon/smithy/aws/swift/codegen/customization/presignable/PresignableUrlIntegration.kt
ganeshnj: trivial, create a set of operation ids than plain || check.

codegen/smithy-aws-swift-codegen/src/main/kotlin/software/amazon/smithy/aws/swift/codegen/middleware/AWSSigningMiddleware.kt
ganeshnj: I would prefer to not have default values, they are useful but very easy to miss because not enforced and likely to result in an undesired outcome.

codegen/smithy-aws-swift-codegen/src/main/kotlin/software/amazon/smithy/aws/swift/codegen/customization/presignable/PresignableUrlIntegration.kt
ganeshnj: Does this change only affect these two ops or affect other operations also?

codegen/smithy-aws-swift-codegen/src/main/kotlin/software/amazon/smithy/aws/swift/codegen/middleware/AWSSigningMiddleware.kt
jbelkins: Comments have been added to `AWSSigningParams` values, and default values have been removed (they were never relied on anyway.)

codegen/smithy-aws-swift-codegen/src/main/kotlin/software/amazon/smithy/aws/swift/codegen/middleware/AWSSigningMiddleware.kt
jbelkins: Done.  `AWSSigningParams` absorbed the logic out of `AWSSigningMiddleware`, and now has its own tests in `AWSSigningMiddlewareTests`.

codegen/smithy-aws-swift-codegen/src/main/kotlin/software/amazon/smithy/aws/swift/codegen/customization/presignable/PresignableUrlIntegration.kt
jbelkins: I refactored this check into `listof().contains()` and added a comment that this logic should be overhauled when new resigned URL types are added because while this logic is manageable for three URL types, it will get ugly fast if more are added.

codegen/smithy-aws-swift-codegen/src/main/kotlin/software/amazon/smithy/aws/swift/codegen/middleware/AWSSigningMiddleware.kt
jbelkins: I have now additionally broke the logic away from `AWSSigningParams` into `SigV4Configurator` which generates the param string for use with the Swift middleware.

`SigV4Configurator` has a convenience initializer that takes a set of params & transforms those into which fields should be set in the Swift param string.  This improves testability by exposing the inputs to the param string directly.

codegen/smithy-aws-swift-codegen/src/main/kotlin/software/amazon/smithy/aws/swift/codegen/AWSSigningParams.kt
```diff
@@ -0,0 +1,58 @@
...
+    val forceUnsignedBody: Boolean,
+    // When set to true, the expiration is included in the params string so it may be set to a custom
+    // value named `expiration`.
+    val useExpiration: Boolean
+)
+
+class SigV4Configurator(
+    val useSignatureTypeQueryString: Boolean,
+    val useExpiration: Boolean,
+    val serviceIsS3: Boolean,
```
epau, line 31: Do we have to expose `serviceIsS3` here? Can we add values for `useDoubleURIEncode` and `shouldNormalizeURIPath` instead (and set those values accordingly in the constructor)?

This future proofs this a bit as it makes it easier to add customizations for other services in the future.

epau (approved):
one nit comment otherwise lgtm

codegen/smithy-aws-swift-codegen/src/main/kotlin/software/amazon/smithy/aws/swift/codegen/AWSSigningParams.kt
jbelkins: Changed.  `SigV4Configurator` now knows nothing about services (except in the convenience initializer).  I added a few more tests to cover as well.

PR #732, opened 2022-12-01
feat: Test code-generated Waiters components

### LINKED ISSUE #733: Smithy Waiter support for Swift 5.5 & 5.6
### Describe the feature

Smithy Waiters are currently being built in https://github.com/awslabs/smithy-swift/pull/463.

It has been discovered that the code generated from a JMESPath expression may not compile on Swift before 5.7 due to improvements to the compiler's type inference that were introduced in Swift 5.7.  Providing the type information may prove to require significant effort.

Determine an approach that will allow Smithy Waiters to compile on Swift before 5.7 so that all SDK customers may have use of this feature.

### Use Case

To provide the Smithy Waiters feature on all supported versions of the Swift language.

### Proposed Solution

Use Smithy shape declarations to determine the required Swift types to be explicitly stated at points of variable declaration and/or method calls so that Swift versions before 5.7 can successfully compile code generated from JMESPath expressions.

### Other Information

_No response_

### Acknowledgements

- [ ] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

jbelkins:
## Issue \#
Part of https://github.com/awslabs/smithy-swift/issues/448
Fixes #733

## Description of changes
Provides a Smithy definition of a service/operation with a waiter, code-generates it in conjunction with protocol tests, and tests the code-generated waiter to verify that its logic behaves as expected.

Also includes a couple of handy scripts for use during codegen development.


scripts/regen.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+osascript -e 'quit app "Xcode"'
```
epau, line 3: Mind leaving a comment here explaining what this script does?

codegen/protocol-test-codegen-local/Tests/swift-codegen/waitersTests/WaitersTests.swift
```diff
@@ -0,0 +1,65 @@
+import XCTest
+@testable import waiters
+@testable import ClientRuntime
+
+class WaitersTests: XCTestCase {
```
epau, line 5: I would expect these tests to be codegenerated and end up in a location like `/codegen/protocol-test-codegen-local/build/smithyprojections/protocol-test-codegen-local/....` following the pattern as the other protocol tests

scripts/regen.sh
jbelkins: I actually didn't mean to commit this, but it is useful when doing codegen development.  Maybe I clean it up / comment it & leave it in `scripts/` permanently...

epau: yeah that'd be dope

Package.swift
```diff
@@ -340,7 +340,7 @@ let package = Package(
         .library(name: "AWSXRay", targets: ["AWSXRay"]),
     ],
     dependencies: [
-        .package(url: "https://github.com/awslabs/smithy-swift", branch: "main"),
+        .package(url: "https://github.com/awslabs/smithy-swift", branch: "jbe/acceptors"),
```
jbelkins, line 343: The smithy PR https://github.com/awslabs/smithy-swift/pull/483 must merge along with this one, and this must be reverted to `main` prior to this PR merging

codegen/Package.swift
```diff
@@ -103,7 +105,7 @@ if let smithySwiftDir = ProcessInfo.processInfo.environment["SMITHY_SWIFT_CI_DIR
     ]
 } else {
     package.dependencies += [
-        .package(path: "~/Projects/Amplify/SwiftSDK/smithy-swift"),
-        .package(path: "~/Projects/Amplify/SwiftSDK/aws-sdk-swift"),
+        .package(path: "../../smithy-swift"),
+        .package(path: "../../aws-sdk-swift"),
```
jbelkins, line 109: This should remove the last requirement that the SDK be installed at `~/Projects/Amplify/SwiftSDK` for the project or any of its subprojects to function correctly in development. 🎉

packageDependencies.plist
```diff
@@ -7,7 +7,7 @@
 	<key>awsCRTSwiftVersion</key>
 	<string>0.3.1</string>
 	<key>clientRuntimeBranch</key>
-	<string>main</string>
+	<string>jbe/acceptors</string>
```
jbelkins, line 10: The smithy PR https://github.com/awslabs/smithy-swift/pull/483 must merge along with this one, and this must be reverted to `main` prior to this PR merging

codegen/Package.swift
jbelkins: (installing `aws-sdk-swift`, `smithy-swift`, and `aws-crt-swift` in the same dir is still required but it can be anywhere on the file system)

AWSClientRuntime/Sources/Protocols/RestJSON/RestJSONError.swift
```diff
@@ -50,7 +50,7 @@ public struct RestJSONError {
     }
 
     /// Filter additional information from error name and sanitize it
-    // Reference: https://awslabs.github.io/smithy/1.0/spec/aws/aws-restjson1-protocol.html#operation-error-serialization
+    /// Reference: https://awslabs.github.io/smithy/1.0/spec/aws/aws-restjson1-protocol.html#operation-error-serialization
```
jbelkins, line 53: Fixes a SwiftLint warning

scripts/regen.sh
jbelkins: Much more detail now

packageDependencies.plist
jbelkins: (Could also point this to `feat/waiters` if not merging the Waiters feature immediately after this)

codegen/protocol-test-codegen-local/Tests/swift-codegen/WaitersTests/ErrorTypeMatcherTests.swift
```diff
@@ -0,0 +1,66 @@
...
+        let response = GetWidgetOutputResponse()
+        let subject = try WaitersClient.errorTypeMatcherWaiterConfig().acceptors[0]
+        let match = subject.evaluate(input: anInput, result: .success(response))
+        XCTAssertNil(match)
+    }
+}
+
+// Error types used in tests above
+
+private struct WaiterTypedErrorImpl: WaiterTypedError, Equatable {
```
epau, line 51: While wordy, its preferable to use `Implementation` here over `Impl` for clarity.
From https://www.swift.org/documentation/api-design-guidelines/#promote-clear-usage
>Avoid abbreviations. Abbreviations, especially non-standard ones, are effectively terms-of-art, because understanding depends on correctly translating them into their non-abbreviated forms.

jbelkins: lol time to give up on `Int` and `Bool` then

jbelkins: Renamed these errors without abbreviations (and with more descriptive names too)

epau (approved):
Looks great!
Just revert the clientRuntimeBranch back to `main` then merge

Package.swift
jbelkins: This has been pointed to `feat/waiters` since https://github.com/awslabs/smithy-swift/pull/483 has merged to that branch.

jbelkins:
@epau Pointed smithy-swift to `feat/waiters` instead of `main` since the corresponding smithy-swift work merges there first.

Once there are green builds on `feat/waiters` on both repos, we can merge those each to `main` (after changing the smithy-swift branch to `main`.)

Issue #756, opened 2022-12-12
feat: Update SDK & smithy-swift to use aws-crt-swift 0.5.0


jbelkins:
### Describe the feature

AWS CRT for Swift v0.5.0 was recently released: https://github.com/awslabs/aws-crt-swift/releases/tag/0.5.0

This release includes breaking changes in its API.  `aws-sdk-swift` and `smithy-swift` will need to be adapted to the new CRT APIs.

### Use Case

To get use of latest CRT features

### Acknowledgements

- [ ] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

PR #777, opened 2022-12-21
feat: Update CRT to 0.5.0+

epau:
## Issue \#
#756 

## Description of changes
Updates code to make it compatible with CRT 0.5.0+

## Conventional Commits


AWSClientRuntime/Sources/Auth/AWSCredentialsProvider.swift
```diff
@@ -6,89 +6,123 @@ import AwsCommonRuntimeKit
...
-    let crtCredentialsProvider: CRTAWSCredentialsProvider
-    private let sharedDefaultIO: SDKDefaultIO = SDKDefaultIO.shared
+    let crtCredentialsProvider: AwsCommonRuntimeKit.CredentialsProvider
     
-    init(awsCredentialsProvider: CRTAWSCredentialsProvider) {
+    init(awsCredentialsProvider: AwsCommonRuntimeKit.CredentialsProvider) {
         self.crtCredentialsProvider = awsCredentialsProvider
     }
     
     public static func fromEnv() throws -> AWSCredentialsProvider {
```
ganeshnj, line 15: all these static methods look scary from testing point of view. Not related to this PR. I would much prefer a singleton pattern.

AWSClientRuntime/Sources/Auth/AWSCredentialsProviderProfileOptions.swift
```diff
@@ -17,44 +17,15 @@ public struct AWSCredentialsProviderProfileOptions {
...
-extension AWSCredentialsProviderProfileOptions {
-    func toCRTType() -> CredentialsProviderProfileOptions {
-        return CredentialsProviderProfileOptions(configFileNameOverride: configFileNameOverride,
-                                                 profileFileNameOverride: profileFileNameOverride,
-                                                 credentialsFileNameOverride: credentialsFileNameOverride,
-                                                 shutdownOptions: shutdownCallback)
-    }
-}
-
-struct CredentialsProviderProfileOptions: CRTCredentialsProviderProfileOptions {
```
ganeshnj, line 40: not required?

AWSClientRuntime/Sources/Auth/AWSCredentialsProviderSTSConfig.swift
```diff
@@ -7,64 +7,26 @@
...
-                                            tlsContext: SDKDefaultIO.shared.tlsContext,
-                                            credentialsProvider: credentialsProvider.crtCredentialsProvider,
-                                            roleArn: roleArn,
-                                            sessionName: sessionName,
-                                            durationSeconds: durationSeconds,
-                                            shutDownOptions: shutDownCallback)
-    }
-}
-
-struct CredentialsProviderSTSConfig: CRTCredentialsProviderSTSConfig {
```
ganeshnj, line 44: not required?

AWSClientRuntime/Sources/Auth/AWSCredentialsProviderStaticConfig.swift
```diff
@@ -16,44 +16,15 @@ public struct AWSCredentialsProviderStaticConfig {
...
-extension AWSCredentialsProviderStaticConfig {
-    func toCRTType() -> CredentialsProviderStaticConfig {
-        return CredentialsProviderStaticConfig(accessKey: accessKey,
-                                               secret: secret,
-                                               sessionToken: sessionToken,
-                                               shutDownOptions: shutDownCallback)
-    }
-}
-
-struct CredentialsProviderStaticConfig: CRTCredentialsProviderStaticConfigOptions {
```
ganeshnj, line 39: not required?

AWSClientRuntime/Sources/Auth/AWSCredentialsProviderWebIdentityConfig.swift
```diff
@@ -10,35 +10,8 @@ import ClientRuntime
...
-
-extension AWSCredentialsProviderWebIdentityConfig {
-    func toCRTType() -> CredentialsProviderWebIdentityConfig {
-        return CredentialsProviderWebIdentityConfig(bootstrap: SDKDefaultIO.shared.clientBootstrap,
-                                                    tlsContext: SDKDefaultIO.shared.tlsContext,
-                                                    shutDownOptions: shutDownCallback)
-    }
-}
-
-struct CredentialsProviderWebIdentityConfig: CRTCredentialsProviderWebIdentityConfig {
```
ganeshnj, line 28: not required?

AWSClientRuntime/Sources/Endpoints/AWSEndpointsResolvedEndpoint.swift
```diff
@@ -6,29 +6,49 @@ import AwsCommonRuntimeKit
...
     }
     
     public func getType() -> AWSEndpointsResolvedEndpointType {
-        AWSEndpointsResolvedEndpointType(crtType: crtResolvedEndpoint.getType())
+        AWSEndpointsResolvedEndpointType(crtType: crtResolvedEndpoint)
     }
     
-    public func getError() throws -> String? {
-        try crtResolvedEndpoint.getError()
+    public func getError() -> String? {
```
ganeshnj, line 19: this would require changes in endpoint resolver. you would have to include those changes in this PR to pass checks.

ganeshnj (changes requested):

AWSClientRuntime/Sources/Middlewares/Sha256TreeHashMiddleware.swift
```diff
@@ -50,30 +50,19 @@ public struct Sha256TreeHashMiddleware<OperationStackOutput: HttpResponseBinding
...
-                break
-            }
-            let hash = oneMbTempBuffer.sha256()
-            hashes.append(hash.toByteArray())
-        }
-        
-        return (bytes.sha256().encodeToHexString(), computeTreeHash(hashes: hashes))
+    private func computeHashes(data: Data) throws -> (String?, String?) {
+        let ONE_MB = 1024 * 1024
+        let hashes: [[UInt8]] = try data.chunked(size: ONE_MB).map{ try $0.sha256().bytes() }
```
jbelkins, line 57: where is `chunked(size:)` defined?  I don't see it on Foundation.  I think I understand what it does but would like to se the implementation / docs for it

also fix the spacing in `map{` here so SwiftLint is happy

AWSClientRuntime/Sources/PrimitiveExtensions/Data+Extension.swift
```diff
@@ -0,0 +1,27 @@
+//
+//  File.swift
+//  
+//
+//  Created by Paulosky, Ed on 12/21/22.
+//
```
jbelkins, line 6: again use the standard file header

AWSClientRuntime/Sources/Regions/DefaultRegionResolver.swift
```diff
@@ -27,14 +27,10 @@ public struct DefaultRegionResolver: RegionResolver {
...
-            do {
-                logger.debug("Attempting to resolve region with: \(String(describing: type(of: provider)))")
-                if let region = try await provider.resolveRegion() {
-                    logger.debug("Resolved region with: \(String(describing: type(of: provider)))")
-                    return region
-                }
-            } catch {
-                return nil
+            logger.debug("Attempting to resolve region with: \(String(describing: type(of: provider)))")
+            if let region = try? await provider.resolveRegion() {
```
jbelkins, line 31: If this call throws, do we need to do something more than log failure?  With `try?` we don't even see the cause of the failure.

jbelkins (changes requested):
A question and a few superficial changes

AWSClientRuntime/Sources/Auth/AWSCredentialsProvider.swift
epau: Are we only ever supposed to have a single credentials provider instance (for each type?), if not then im not sure we want singleton pattern here. (ideally i'd like to avoid singletons and global state whenever possible). 

But I agree, the current interface for creating credential providers has much room for improvement

AWSClientRuntime/Tests/DataExtensionTests.swift
```diff
@@ -0,0 +1,48 @@
+//
+//  File.swift
+//  
+//
+//  Created by Paulosky, Ed on 12/22/22.
+//
```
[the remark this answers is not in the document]
epau: oops, great catch!!

AWSClientRuntime/Sources/PrimitiveExtensions/Data+Extension.swift
epau: oops great catch!

AWSClientRuntime/Sources/Auth/AWSCredentialsProviderProfileOptions.swift
epau: Nope. This struct no longer has any purpose. 
Checkout the related factory method in [AwsCredentialsProvider]](https://github.com/awslabs/aws-sdk-swift/pull/777/files#diff-c858c06a42a458b91bb72c1ce5190f515a40cf4463775e74d8f8f91d4e53c506) to see why its no needed.

AWSClientRuntime/Sources/Auth/AWSCredentialsProviderSTSConfig.swift
epau: Nope. This struct no longer has any purpose. 
Checkout the related factory method in [AwsCredentialsProvider]](https://github.com/awslabs/aws-sdk-swift/pull/777/files#diff-c858c06a42a458b91bb72c1ce5190f515a40cf4463775e74d8f8f91d4e53c506) to see why its no needed.

AWSClientRuntime/Sources/Auth/AWSCredentialsProviderStaticConfig.swift
epau: Nope. This struct no longer has any purpose. 
Checkout the related factory method in [AwsCredentialsProvider]](https://github.com/awslabs/aws-sdk-swift/pull/777/files#diff-c858c06a42a458b91bb72c1ce5190f515a40cf4463775e74d8f8f91d4e53c506) to see why its no needed.

AWSClientRuntime/Sources/Middlewares/Sha256TreeHashMiddleware.swift
epau: It is defined in the new [Data+Extensions](https://github.com/awslabs/aws-sdk-swift/pull/777/files#diff-ad633ba5ccecb6552a23daf2e47879407d56c81a3eae9bcb5201f480a73f7b99R16)

AWSClientRuntime/Sources/Regions/DefaultRegionResolver.swift
epau: We need to escape the error here somehow. The best we can do is log the error, which i can add in if that's desired. Otherwise, if we catch and return nil then we may not resolve a region if an earlier resolver fails. We should not fail, and instead move on to the next resolver in the list. (im looking for a SEP to verify this is indeed the intended behavior)

epau: Verified the intended behavior in slack.

jbelkins (approved):
Approved, pending resolution of @ganeshnj's comments and a green build.

ganeshnj (approved):

Issue #747, opened 2022-12-06
AWSSecurityLake fails to build


jbelkins:
### Describe the bug

After updating models on 12/6/2022 in preparation for release `0.6.0`, the following error was received when attempting to build the generated SDK:
```
/Users/runner/work/aws-sdk-swift/aws-sdk-swift/release/AWSSecurityLake/models/Models.swift:621:25: error: type 'Dictionary<String, [String]>.Element' (aka '(key: String, value: Array<String>)') cannot conform to 'Encodable'
                    try alldimensionsmap0Container.encode(valueset1)
                        ^
```
Note that this error occurred while running the CI action for Release, which runs Xcode 13.2.1.  This code seems to compile without error on Xcode 14.

The `AWSSecurityLake` service was excluded from release `0.6.0` and the release was continued.

### Expected Behavior

Generated SDK compiles without error.

### Current Behavior

Generated SDK compiles with the error logged above.

### Reproduction Steps

See [this CI run](https://github.com/awslabs/aws-sdk-swift/actions/runs/3633276301/jobs/6130098127#step:5:40272) for an occurrence of this error.

### AWS SWIFT SDK version used

Latest

### Compiler and Version used

Xcode 13.2.1 on Github runner

### Operating System and version

macOS 11

jbelkins:
This is the extension that causes the failure:

```swift
extension CreateAwsLogSourceInput: Swift.Encodable {
    enum CodingKeys: Swift.String, Swift.CodingKey {
        case enableAllDimensions
        case enableSingleDimension
        case enableTwoDimensions
        case inputOrder
    }

    public func encode(to encoder: Swift.Encoder) throws {
        var encodeContainer = encoder.container(keyedBy: CodingKeys.self)
        if let enableAllDimensions = enableAllDimensions {
            var enableAllDimensionsContainer = encodeContainer.nestedContainer(keyedBy: ClientRuntime.Key.self, forKey: .enableAllDimensions)
            for (dictKey0, alldimensionsmap0) in enableAllDimensions {
                var alldimensionsmap0Container = enableAllDimensionsContainer.nestedUnkeyedContainer(forKey: ClientRuntime.Key(dictKey0))
                for valueset1 in alldimensionsmap0 {
                    try alldimensionsmap0Container.encode(valueset1)   // <--- compile failure happens here
                }
            }
        }
        if let enableSingleDimension = enableSingleDimension {
[...]

PR #779, opened 2022-12-23
fix: Add protocol tests for nested & sparse collections

jbelkins:
## Issue \#
Related to #747 

## Description of changes
#747 identifies an AWS service (`AWSSecurityLake`) that fails to build because `smithy-swift` does not codegen `encode(to:)` correctly for a Smithy shape with nested maps.

This PR adds four new tests to the `rest-json-extras` protocol test suite:
- `CreateNestedMap` request & response tests, which check serialization for a shape with the same pattern of nested maps that caused compile failure in the `AWSSecurityLake` service.
- `CreateNestedSparseMap` request & response tests, which check serialization for combinations of nested and sparse maps & lists.

The fix for codegen has been made in https://github.com/awslabs/smithy-swift/pull/492, not here, but these tests verify that valid, working serialization code is generated.

Because the `AWSSecurityLake` service now builds successfully, it is no longer excluded from SDK builds.


codegen/Package.swift
```diff
@@ -61,6 +61,10 @@ appendLibTarget(name: "rest_json_extras", path: "\(baseDirLocal)/rest_json_extra
 appendTstTarget(name: "rest_json_extrasTests", path: "\(baseDirLocal)/rest_json_extras", dependency: "rest_json_extras")
 appendLibTarget(name: "Waiters", path: "\(baseDirLocal)/Waiters")
 appendTstTarget(name: "WaitersTests", path: "./protocol-test-codegen-local/Tests", dependency: "Waiters")
+appendLibTarget(name: "NestedAggregates", path: "\(baseDirLocal)/NestedAggregates")
+appendTstTarget(name: "NestedAggregatesTests", path: "\(baseDirLocal)/NestedAggregates", dependency: "NestedAggregates")
+appendLibTarget(name: "NestedSparseAggregates", path: "\(baseDirLocal)/NestedSparseAggregates")
+appendTstTarget(name: "NestedSparseAggregatesTests", path: "\(baseDirLocal)/NestedSparseAggregates", dependency: "NestedSparseAggregates")
```
epau, line 67: nit: imo it'd be great to combine these. There are many many traits so I'm not sure we need to get this granular as it could grow uncontrollably. Since they are all using the `restJson1` protocol, perhaps we add these tests to the `rest_json_extras` and `rest_json_extrasTests` targets?

epau (approved):
Just one nit comment, but also something we can do in the future

codegen/Package.swift
jbelkins: I will look at consolidating these tests with the existing `restJson` tests.

I will also look at separating the waiters tests from the `restJson` suite.

jbelkins: @epau The new tests here are added to `rest-json-extras` instead of their own protocol test suites.
Waiters remains its own protocol test suite, and its Smithy definitions have been split to a separate file.

codegen/protocol-test-codegen-local/model/rest-json-extras.smithy
```diff
@@ -0,0 +1,264 @@
+$version: "1.0"
```
ganeshnj, line 1: is this model file processed in CI?

ganeshnj (approved):

Issue #786, opened 2022-12-28
Replace rogue encoder and decoder instances


epau:
### Describe the bug

There are a few instances where one-off encoders and decoders are being created by models. We should be using the encoder/decoder from the client instead of creating new instances. 

There are a few benefits to this 
- we get any configuration that was performed on the encoder/decoder
- we remove coupling and follow proper dependency injection pattern

One example: https://github.com/awslabs/aws-sdk-swift/blob/39db4f1c33ef04169540849af496d1747f7b26f4/AWSClientRuntime/Sources/Protocols/Ec2Query/Ec2QueryError.swift#L18

### Expected Behavior

N/A

### Current Behavior

N/A

### Reproduction Steps

N/A

### AWS SWIFT SDK version used

N/A

### Compiler and Version used

N/A

### Operating System and version

N/A

jbelkins:
Closed due to new closure-based serde.

Issue #761, opened 2022-12-17
S3 404 errors not parsing


jbelkins:
### Describe the bug

When a 404 error is received for AWS S3 HeadObject or HeadBucket, the error fails to parse.

### Expected Behavior

Expect to receive a NotFound SDK error parsed from the response.

### Current Behavior

A client-side XML deserialization error is returned.
From inspecting the response, the HTTP code is 404 but the body appears to be empty.

### Reproduction Steps

Using the S3Client, perform a `headBucket` operation for a bucket that doesn't exist; observe the error thrown by the client.

### AWS SWIFT SDK version used

main branch

### Compiler and Version used

Xcode 14.2

### Operating System and version

macOS 12.6.1

epau:
Possibly same issue as https://github.com/awslabs/aws-sdk-swift/issues/759

Issue #759, opened 2022-12-16
EC2 errors not parsing


jbelkins:
### Describe the bug

When an error is received from an operation on EC2, the error is not being parsed correctly.

Example code:
```swift
        let client = try EC2Client(region: "us-east-1")
        let in1 = RunInstancesInput()  // there are required parameters for this operation
        let out1 = try await client.runInstances(input: in1)
```

The actual body of the returned error:
```
<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n<Response><Errors><Error><Code>MissingParameter</Code><Message>The request must contain the parameter ImageId</Message></Error></Errors><RequestID>2e86d905-6423-44ac-8fba-52158ee01129</RequestID></Response>
```

The error returned by the Swift SDK:
```
client(ClientRuntime.ClientError.retryError(Swift.DecodingError.keyNotFound(CodingKeys(stringValue: "RequestId", intValue: nil), Swift.DecodingError.Context(codingPath: [CodingKeys(stringValue: "RequestId", intValue: nil)], debugDescription: "No attribute or element found for key CodingKeys(stringValue: \"RequestId\", intValue: nil) (\"RequestId\").", underlyingError: nil))), nil)
```

### Expected Behavior

MissingParameterError should be returned by the call to `runInstances`

### Current Behavior

A Swift DecodingError was returned

### Reproduction Steps

Run the code in the description above

### AWS SWIFT SDK version used

Latest

### Compiler and Version used

Latest

### Operating System and version

Latest

PR #787, opened 2022-12-28
fix: Fixes S3 and Ec2 error parsing

epau:
This PR fixes issues with parsing S3 and Ec2 errors. 

S3 errors were failing to parse correctly when the error was a 404 with an empty body. The cause of this issue is that our local logic to check if the body is empty was incorrect. It previously, defined `isEmpty` as having a httpBody instance that equaled `.data(nil)`. Now, it defines `isEmpty` as having a httpBody instance that is either `.none`, or the data is nil or is empty. 

Ec2 errors were failing to parse correctly because the request id key in the response is `RequestID` while we were expecting `RequestId`. According to the smithy spec for Ec2Query protocol, `RequestId` is correct so this appears to a special case for the Ec2 service. Now the Ec2 error models will attempt to parse using both `RequestID` and `RequestId`. 

After [this issue](https://github.com/awslabs/aws-sdk-swift/issues/786) is done, we can come back here and fix the Ec2 errors in a different way. We can define a custom key decoding strategy, and always replace the key `RequestId` with `RequestID` just for the Ec2 client. 

## Issue \#

#761 
#759 

## Conventional Commits


AWSClientRuntime/Sources/Protocols/Ec2Query/Ec2NarrowedResponse.swift
```diff
@@ -12,5 +12,16 @@ public struct Ec2NarrowedResponse<T>: Decodable where T: Decodable {
     enum CodingKeys: String, CodingKey {
         case errors = "Errors"
         case requestId = "RequestId"
+        case requestID = "RequestID"
+    }
+    
+    public init(from decoder: Decoder) throws {
```
jbelkins, line 18: Was this type using an auto-generated `init(from:)` before?

AWSClientRuntime/Sources/Protocols/Ec2Query/Ec2Response.swift
```diff
@@ -12,5 +12,16 @@ public struct Ec2Response: Decodable {
     enum CodingKeys: String, CodingKey {
         case errors = "Errors"
         case requestId = "RequestId"
+        case requestID = "RequestID"
+    }
+    
+    public init(from decoder: Decoder) throws {
```
jbelkins, line 18: Same question: Was this type using an auto-generated `init(from:)` before?

codegen/smithy-aws-swift-codegen/src/main/kotlin/software/amazon/smithy/aws/swift/codegen/customization/s3/S3ErrorIntegration.kt
```diff
@@ -92,28 +96,6 @@ class S3ErrorIntegration : SwiftIntegration {
             writer.dedent()
             writer.write("}")
         }
-
-        writer.openBlock("extension $operationErrorName {", "}") {
```
jbelkins, line 96: Weird that this was all code generated.  Nothing in it is customized except the name of the type being extended.  Sounds basically like a protocol + extension... which you made it into.

AWSClientRuntime/Tests/Protocols/Ec2Query/Ec2ErrorRequestIdTests.swift
```diff
@@ -0,0 +1,73 @@
...
+// SPDX-License-Identifier: Apache-2.0
+//
+        
+
+import ClientRuntime
+import SmithyTestUtil
+import XCTest
+@testable import AWSClientRuntime
+
+class Ec2ErrorRequestIdTests: XCTestCase {
```
jbelkins, line 14: Can/should these tests be written as EC2-specific protocol tests?  If so what's the advantages or disadvantages of one approach over another?

jbelkins (commented):
Overall good, just a couple Q's.

Good that we are moving logic out of generated code along with fixing these bugs.

AWSClientRuntime/Sources/Protocols/Ec2Query/Ec2NarrowedResponse.swift
epau: yep!

AWSClientRuntime/Sources/Protocols/Ec2Query/Ec2Response.swift
epau: yep!

codegen/smithy-aws-swift-codegen/src/main/kotlin/software/amazon/smithy/aws/swift/codegen/customization/s3/S3ErrorIntegration.kt
epau: yeah i thought so too, there was much duplication of the same exact logic previously.

AWSClientRuntime/Tests/Protocols/Ec2Query/Ec2ErrorRequestIdTests.swift
epau: I don't see any reason to write these tests as protocol tests since the code being tested isn't generated.
I do think integration tests would be a nice addition here though, to test the same end-to-end flow I tested manually, but we currently don't have a mechanism for that. Once we do, this (like the other service specific bugs) are good candidates to add integration tests for. 

curious to hear your thoughts on this as well!

ganeshnj:
LGTM, could you add documentation for public APIs?

AWSClientRuntime/Sources/Http/HttpResponse+AWS.swift
```diff
@@ -0,0 +1,21 @@
...
+        return statusCode == .notFound && body.isEmpty
+    }
+    
+    var requestId: String? {
+        return headers.value(for: "x-amz-request-id")
+    }
+    
+    var requestId2: String? {
+        return headers.value(for: "x-amz-id-2")
+    }
```
ganeshnj, line 20: I wonder if this extension should be `headers`  collection.

epau: Good question, and can go either way here. IMO the fact that these values come from headers are implementation details. That being said, it could be cool to live in a world where all the headers are defined on `Headers`.

jbelkins (approved):

AWSClientRuntime/Sources/Http/HttpResponse+AWS.swift
ganeshnj: Makes sense.

ganeshnj (approved):

Issue #214, opened 2021-07-13
Route53 Custom Error Unmarshalling


wooj2:
ChangeResourceRecordSets uses a custom error format: aws/aws-sdk-go@3265860

*internal id*: 176722143

Issue #789, opened 2022-12-29
Investigate a mechanism to hand write service specific customizations


epau:
### Describe the feature

There are times where we need to make service specific customizations. One example is #214. Currently, we need to implement the customization via codegen, even if all the code is static (does not depend on smithy models). 

It would be much simpler if we could implement these customizations directly as swift code (without needing to go through the code generation process). We still would only want these files to be included in their corresponding service module, which makes this tricky. 

Investigate a way that we can write service specific swift files (and tests). 

### Use Case

N/A

### Acknowledgements

- [ ] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

ganeshnj:
not adding to GA as we are not blocked for any feature that needs this.

sichanyoo:
Moved to 2168

PR #791, opened 2022-12-29
fix: Support custom error format for Route53 ChangeResourceRecordSets operation

epau:
This PR adds support for a custom error format that is sometimes returned by the `ChangeResourceRecordSets` operation in the Route53 service. 

The custom error looks something like the following: 
```xml
<?xml version="1.0" encoding="UTF-8"?>
<InvalidChangeBatch xmlns="https://route53.amazonaws.com/doc/2013-04-01/">
  <Messages>
    <Message>Tried to create resource record set duplicate.example.com. type A, but it already exists</Message>
  </Messages>
  <RequestId>b25f48e8-84fd-11e6-80d9-574e0c4664cb</RequestId>
</InvalidChangeBatch>
```

## Issue \#
#214 

## Conventional Commits


epau:
This PR is targeting the `epau/feat/support-crt-0.5.0` branch since it depends on changes made there. The target branch will be updated to main once `epau/feat/support-crt-0.5.0` merges.

codegen/smithy-aws-swift-codegen/src/main/kotlin/software/amazon/smithy/aws/swift/codegen/customization/route53/Route53InvalidBatchErrorIntegration.kt
```diff
@@ -0,0 +1,90 @@
...
+                    writer.write("statusCode: httpResponse.statusCode")
+                }
+                writer.write("self = .invalidChangeBatch(invalidChangeBatchError)")
+                writer.write("return")
+            }
+        }
+        writer.write(previousCode)
+    }
+
+    override fun writeAdditionalFiles(ctx: CodegenContext, protocolGenerationContext: ProtocolGenerator.GenerationContext, delegator: SwiftDelegator) {
```
epau, line 43: It would be awesome to have a way to just write this in swift, since it isn't dependent on the smithy model. 
I filed [this issue](https://github.com/awslabs/aws-sdk-swift/issues/789) to investigate some mechanism that would allow us to write service specific swift code. I imagine the solution here will be related (if not the same) as solving the service specific integration test problem.

codegen/smithy-aws-swift-codegen/src/main/kotlin/software/amazon/smithy/aws/swift/codegen/customization/route53/Route53InvalidBatchErrorIntegration.kt
```diff
@@ -0,0 +1,90 @@
...
+        delegator.useFileWriter("${ctx.settings.moduleName}/models/ChangeResourceRecordSetsOutputError+Customization.swift") { writer ->
+            writer.addImport(SwiftDependency.CLIENT_RUNTIME.target)
+            renderCustomInvalidBatchError(writer)
+            renderInvalidChangeBatch(writer)
```
epau, line 47: It would be awesome to have a way to just write this in swift, since it isn't dependent on the smithy model. 
I filed [this issue](https://github.com/awslabs/aws-sdk-swift/issues/789) to investigate some mechanism that would allow us to write service specific swift code. I imagine the solution here will be related (if not the same) as solving the service specific integration test problem.

codegen/smithy-aws-swift-codegen/src/main/kotlin/software/amazon/smithy/aws/swift/codegen/customization/route53/Route53InvalidBatchErrorIntegration.kt
```diff
@@ -0,0 +1,90 @@
...
+        }
+    }
+
+    private fun renderCustomInvalidBatchError(writer: SwiftWriter) {
+        writer.openBlock("struct CustomInvalidBatchError: Decodable {", "}") {
```
ganeshnj, line 52: public APIs/structs must have docs.

ganeshnj (changes requested):

ganeshnj:
LGTM, just small comment on docs.

ganeshnj (approved):

codegen/smithy-aws-swift-codegen/src/main/kotlin/software/amazon/smithy/aws/swift/codegen/customization/route53/Route53InvalidBatchErrorIntegration.kt
epau: This isn't public. This is just internal as it's just used for decoding and then it's transformed into a `InvalidChangeBatch` instance.

PR #797, opened 2023-01-04
retroactively merging 0.8.0 release changes

ganeshnj:
## Issue \#

I wasn't able to merge https://github.com/awslabs/aws-sdk-swift/pull/795 because of not able to execute the checks on the generated commit.

My hunch here is, because we add a new commit as part of PR, which doesn't trigger the checks per https://github.com/EndBug/add-and-commit#the-commit-from-the-action-is-not-triggering-ci

Added backlog to address this https://github.com/awslabs/aws-sdk-swift/issues/798

## Conventional Commits


.github/workflows/release.yml
```diff
@@ -30,6 +25,7 @@ env:
 
 jobs:
   macos-release:
+    if:  startsWith(github.head_ref, 'release/')
```
ganeshnj, line 28: this allows to run workflow on release branch only.

.github/workflows/release.yml
```diff
@@ -1,12 +1,7 @@
 name: Release
 
 on:
-  push:
-    branches:
-        - 'release/**'
   pull_request:
-    branches:
-        - 'release/**'
```
ganeshnj, line 9: this was wrong, here the list of branches listed where the PR is targeted to.

.github/workflows/release.yml
```diff
@@ -1,12 +1,7 @@
 name: Release
 
 on:
-  push:
-    branches:
-        - 'release/**'
```
ganeshnj, line 6: this causes recursive triggers.

epau (approved):

PR #799, opened 2023-01-04
chore: Reverts version pinning

epau:
Reverts version pinning to smithy and crt repos. 

## Conventional Commits

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -356,10 +356,6 @@ let package = Package(
         .library(name: "AWSWorkSpacesWeb", targets: ["AWSWorkSpacesWeb"]),
         .library(name: "AWSXRay", targets: ["AWSXRay"]),
     ],
-    dependencies: [
-        .package(url: "https://github.com/awslabs/smithy-swift", branch: "main"),
-        .package(url: "https://github.com/awslabs/aws-crt-swift", branch: "main")
-    ],
     targets: [
         .target(
             name: "AWSClientRuntime",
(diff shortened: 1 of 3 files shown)


Package.swift
```diff
@@ -357,8 +357,8 @@ let package = Package(
         .library(name: "AWSXRay", targets: ["AWSXRay"]),
     ],
     dependencies: [
-        .package(url: "https://github.com/awslabs/smithy-swift", .exact("0.8.0")),
-        .package(url: "https://github.com/awslabs/aws-crt-swift", .exact("0.4.0"))
```
ganeshnj, line 361: how would the release work here?

epau: Good question, this would get overwritten with every release, since the release workflow regenerates the Package.swift. 

So perhaps we don't make this change since CI will generate a Package.swift anyways?

ganeshnj: We must do something like based on a env var, we use local deps, else use the version mentioned here

Something like https://github.com/apple/swift-syntax/blob/main/Package.swift#L198-L207

epau: 👍🏻 Should be addressed now.

ganeshnj: Very nice,

https://github.com/awslabs/smithy-swift/blob/60de8a4792f11d0d676455fd60656801c2f5d643/smithy-swift-codegen/src/main/kotlin/software/amazon/smithy/swift/codegen/PackageManifestGenerator.kt#L12 be updated?

epau: Is there a scenario where we would want the generated protocol test packages to use different versions of smithy-swift and aws-sdk-swift than the local versions that were used to generate the packages? 

I can't think of any, so I'm not confident it makes sense to add this logic here since these packages aren't stored in git and therefore need to be generated from local copies of smithy-swift and aws-sdk-swift.

ganeshnj (approved):

Issue #801, opened 2023-01-04
Create workflow/job that regenerates models and builds all services


epau:
### Describe the feature

It'd be great to have a workflow/job that does the following:
- regenerates all models
- builds all services

This workflow just needs to be able to be triggered manually. 

### Use Case

Having a workflow for this will help improve the thoroughness of testing while also reducing the burden of individual developers by moving the work from local machines to CI.

### Acknowledgements

- [ ] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

PR #802, opened 2023-01-04
chore: Adds new workflow for building the full sdk

epau:
This PR adds a workflow for regenerating all models, building the full sdk (models included) and running tests.

## Issue \#

#801 

## Conventional Commits

### DIFF
--- a/.github/workflows/codegen-build-test.yml
+++ b/.github/workflows/codegen-build-test.yml
@@ -1,0 +1,49 @@
+# This script is for building the full SDK
+# It will regenerate all models, build the full SDK with the new models, and run tests.
+name: Codegen, Build, and Test
+
+on:
+  workflow_dispatch:
+
+env:
+  BUILDER_VERSION: v0.8.19
+  BUILDER_SOURCE: releases
+  # host owned by CRT team to host aws-crt-builder releases. Contact their on-call with any issues
+  BUILDER_HOST: https://d19elf31gohf1l.cloudfront.net
+  PACKAGE_NAME: aws-sdk-swift
+  LINUX_BASE_IMAGE: ubuntu-16-x64
+  RUN: ${{ github.run_id }}-${{ github.run_number }}
+  AWS_SDK_SWIFT_CI_DIR: /Users/runner/work/aws-sdk-swift/aws-sdk-swift
+  AWS_CRT_SWIFT_CI_DIR: /Users/runner/work/aws-sdk-swift/aws-sdk-swift/target/build/deps/aws-crt-swift
+  SMITHY_SWIFT_CI_DIR: /Users/runner/work/aws-sdk-swift/aws-sdk-swift/target/build/deps/smithy-swift
+
+jobs:
+  codegen-build-test:
+    runs-on: macos-11
+    steps:
+      - name: Checkout sources
+        uses: actions/checkout@v2
+      - uses: actions/cache@v2
+        with:
+          path: |
+            ~/.gradle/caches
+            ~/.gradle/wrapper
+          key: ${{ runner.os }}-gradle-${{ hashFiles('**/*.gradle*') }}
+          restore-keys: |
+            ${{ runner.os }}-gradle-
+      - uses: actions/setup-java@v1
+        with:
+          java-version: '11'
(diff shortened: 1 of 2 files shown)


.github/workflows/codegen-build-test.yml
```diff
@@ -0,0 +1,49 @@
...
+      - uses: actions/setup-java@v1
+        with:
+          java-version: '11'
+      - name: Build and Test ${{ env.PACKAGE_NAME }}
+        run: |
+          python3 -c "from urllib.request import urlretrieve; urlretrieve('${{ env.BUILDER_HOST }}/${{ env.BUILDER_SOURCE }}/${{ env.BUILDER_VERSION }}/builder.pyz?run=${{ env.RUN }}', 'builder.pyz')"
+          chmod a+x builder.pyz
+          AWS_CRT_SWIFT_CI_DIR="${{ env.AWS_CRT_SWIFT_CI_DIR }}" AWS_SDK_SWIFT_CI_DIR="${{ env.AWS_SDK_SWIFT_CI_DIR }}" SMITHY_SWIFT_CI_DIR="${{ env.SMITHY_SWIFT_CI_DIR }}" ./builder.pyz build -p ${{ env.PACKAGE_NAME }}
+          ./gradlew -p codegen/sdk-codegen build
+          ./gradlew -p codegen/sdk-codegen stageSdks
```
jbelkins, line 43: I would `rm -rf release/` before running `stageSdks` to ensure that no files from previous generation stick around for your build

jbelkins (changes requested):

.github/workflows/codegen-build-test.yml
epau: ooc why isn't that an issue in the `release.yml` workflow?

jbelkins: Interesting that it's not in there, but it should be.  (It wasn't necessary back when we used to not commit `release/` since there were no pre-existing files; not adding `rm-rf release/` when we did the work to make `main` buildable was an oversight on my part.)

If you add that command in your new workflow and the existing release workflow, I will approve.

ganeshnj:
IMHO, this new workflow is redundant to CI job. What we should do it to make CI do all of this so we get the feedback early.

I totally understand the timing issue as it will take a while to build, but there are ways to speed up the build process for example batching the build to separate hosts
- Build Runtimes
- Start n number of hosts to build services in parallel

epau:
@ganeshnj i agree, do we think it's acceptable to add this now (we can remove it when we add this functionality to the CI job). I'll file an issue to address adding this functionality to the CI job.

jbelkins (approved):

.github/workflows/codegen-build-test.yml
```diff
@@ -0,0 +1,50 @@
...
+      - uses: actions/setup-java@v1
+        with:
+          java-version: '11'
+      - name: Build and Test ${{ env.PACKAGE_NAME }}
+        run: |
+          python3 -c "from urllib.request import urlretrieve; urlretrieve('${{ env.BUILDER_HOST }}/${{ env.BUILDER_SOURCE }}/${{ env.BUILDER_VERSION }}/builder.pyz?run=${{ env.RUN }}', 'builder.pyz')"
+          chmod a+x builder.pyz
+          AWS_CRT_SWIFT_CI_DIR="${{ env.AWS_CRT_SWIFT_CI_DIR }}" AWS_SDK_SWIFT_CI_DIR="${{ env.AWS_SDK_SWIFT_CI_DIR }}" SMITHY_SWIFT_CI_DIR="${{ env.SMITHY_SWIFT_CI_DIR }}" ./builder.pyz build -p ${{ env.PACKAGE_NAME }}
+          ./gradlew -p codegen/sdk-codegen build
+          rm -rf release/
```
ganeshnj, line 43: this is worrying, we must root cause what is being generated/added that causes us to remove the dir.

ganeshnj:
> @ganeshnj i agree, do we think it's acceptable to add this now (we can remove it when we add this functionality to the CI job). I'll file an issue to address adding this functionality to the CI job.

backlog is fine as of now. We must keep this workflow trigger to manual dispatch.

Issue #776, opened 2022-12-21
chore: Run service-specific generated test suites on CI


jbelkins:
### Describe the feature

Starting with Endpoints 2.0, some code-generated services have generated Swift test suites with tests related to endpoint resolution.  There will also likely be services in the future with similar generated tests.

Although these tests are generated, they are not currently run either during development or CI.

These tests should be added to continuous integration on some schedule, and should also be able to be run on dev machines by use of a script or some other means.

### Use Case

To ensure that tested functions of services are not broken accidentally.

### Acknowledgements

- [ ] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

PR #803, opened 2023-01-05
chore: aws-sdk-swift package structuring

ganeshnj:
## Issue \#

Fixes https://github.com/awslabs/aws-sdk-swift/issues/776

## Description of changes

Companion PR https://github.com/awslabs/smithy-swift/pull/503

This PR foundation for testing story.

The repo now looks like
- Sources
  - Core (contains runtime)
  - Services (contains generated services)
- Tests
  - Core (contains runtime tests)
  - Services (contains generated service tests)

In future we will add Integration following the same structure.
- IntegrationTests
  - Core
  - Services

I had trouble with getting the checks passed when any service test is enabled. swift cli fatal fails. I'm going to cut an issue to Swift team about it. However, devs can uncomment any test target and use it in Xcode without any problem.

**Refrain to review Services dirs.**

## Conventional Commits

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -1063,6 +1063,6 @@ case (false, true):
 case (false, false):
     package.dependencies += [
         .package(url: "https://github.com/awslabs/smithy-swift", .exact("0.10.0")),
-        .package(url: "https://github.com/awslabs/aws-crt-swift", .exact("0.5.3"))
+        .package(url: "https://github.com/awslabs/aws-crt-swift", .exact("0.5.4"))
     ]
 }
--- a/packageDependencies.plist
+++ b/packageDependencies.plist
@@ -5,7 +5,7 @@
 	<key>clientRuntimeVersion</key>
 	<string>0.10.0</string>
 	<key>awsCRTSwiftVersion</key>
-	<string>0.5.3</string>
+	<string>0.5.4</string>
 	<key>clientRuntimeBranch</key>
 	<string>main</string>
 	<key>awsCRTSwiftBranch</key>


epau (commented):
Can we place all the code generated services within a folder `Sources/Services/..` ? 
This way its much easier to find `AWSClientRuntime` and its contents within xcode? 

Additionally, we can do the same for the services in tests -> `Tests/Services/..`

ganeshnj:
> Can we place all the code generated services within a folder `Sources/Services/..` ? This way its much easier to find `AWSClientRuntime` and its contents within xcode?
> 
> Additionally, we can do the same for the services in tests -> `Tests/Services/..`

Still working on it, I plan to put generated services in `Sources/Services` and runtimes in `Sources/Runtimes/AWSClientRuntime`, same for tests.

scripts/mergeModels.sh
```diff
@@ -30,19 +30,20 @@ mergeFiles() {
...
-	    exit 1
+	    continue
 	fi
 	pushd ${MODELDIR}    
 	if [ -f ${MODELS} ]; then
 	    echo "${sdk} has ${MODELS} already generated"
-	else
-	    mergeFiles
+        echo "Removing ${MODELS}"
+        rm -f ${MODELS}
```
epau, line 44: Nice!!! 

Nit: can we align the indentation? It appears line 42 is indented more than lines 43 and 44, but they should all have the same indentation

scripts/generatePackageSwift.swift
```diff
@@ -160,48 +160,52 @@ private func dependency(url: String, version: String, branch: String?, path: Str
...
-"""
-    print(targetsBeginning)
+func generateServiceTargets(_ releasedSDKs: [String]) {
+    print("        // MARK: - Service Targets")
+    for sdk in releasedSDKs {
+        print(#"        .target(name: "\#(sdk)", dependencies: ["AWSClientRuntime"], path: "./Sources/Services/\#(sdk)"),"#)
+    }
+    print()
+    print("        // MARK: - Service Test Targets")
+    print("        // TODO: enable test targets for all services")
```
epau, line 170: Do we have an issue filed for this? Can we add a link to the issue in the comment here as well?

.github/workflows/continuous-integration.yml
```diff
@@ -44,9 +44,9 @@ jobs:
           chmod a+x builder.pyz
           AWS_CRT_SWIFT_CI_DIR="${{ env.AWS_CRT_SWIFT_CI_DIR }}" AWS_SDK_SWIFT_CI_DIR="${{ env.AWS_SDK_SWIFT_CI_DIR }}" SMITHY_SWIFT_CI_DIR="${{ env.SMITHY_SWIFT_CI_DIR }}" ./builder.pyz build -p ${{ env.PACKAGE_NAME }} --target=ios-armv8
   macos-compat:
-    runs-on: macos-11
+    runs-on: macos-12
     env:
-      DEVELOPER_DIR: /Applications/Xcode_13.2.1.app
+      DEVELOPER_DIR: /Applications/Xcode_14.0.app
```
epau, line 49: Won't this change the swift version this is run on? 
Do we want to continue to use Xcode 13.2.1 to force swift 5.5.x?

.swiftlint.yml
```diff
@@ -1,5 +1,5 @@
 included:
-  - AWSClientRuntime/Sources
+  - Sources/Core/AWSClientRuntime
```
epau, line 2: Thoughts for when we get around to cleaning up the naming of things. `AWSCore` wouldn't be a terrible rename for `AWSClientRuntime`.

epau (approved):
This looks great!
Nothing blocking, but a few minor questions/comments

.swiftlint.yml
ganeshnj: I have thoughts on this, but don't want to break customers as we will do renaming during the modularization.

.github/workflows/continuous-integration.yml
ganeshnj: I wanted to see if latest swift version solves the problem. Need to revert this back to original value. Don't want to this change as part of this PR as it is unrelated.

scripts/generatePackageSwift.swift
ganeshnj: https://github.com/awslabs/aws-sdk-swift/issues/814

scripts/mergeModels.sh
ganeshnj: fixed

.swiftlint.yml
epau: yep, definitely not something to tackle now

jbelkins (changes requested):

.github/workflows/continuous-integration.yml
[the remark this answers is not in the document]
ganeshnj: outdated

Package.swift
```diff
@@ -358,358 +361,686 @@ let package = Package(
...
+        .target(name: "AWSWorkDocs", dependencies: ["AWSClientRuntime"], path: "./Sources/Services/AWSWorkDocs"),
+        .target(name: "AWSWorkLink", dependencies: ["AWSClientRuntime"], path: "./Sources/Services/AWSWorkLink"),
+        .target(name: "AWSWorkMail", dependencies: ["AWSClientRuntime"], path: "./Sources/Services/AWSWorkMail"),
+        .target(name: "AWSWorkMailMessageFlow", dependencies: ["AWSClientRuntime"], path: "./Sources/Services/AWSWorkMailMessageFlow"),
+        .target(name: "AWSWorkSpaces", dependencies: ["AWSClientRuntime"], path: "./Sources/Services/AWSWorkSpaces"),
+        .target(name: "AWSWorkSpacesWeb", dependencies: ["AWSClientRuntime"], path: "./Sources/Services/AWSWorkSpacesWeb"),
+        .target(name: "AWSXRay", dependencies: ["AWSClientRuntime"], path: "./Sources/Services/AWSXRay"),
+
+        // MARK: - Service Test Targets
+        // TODO: enable test targets for all services
```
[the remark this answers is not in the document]
ganeshnj: No, Initially I though I could just turn on atleast S3 but no. Will keep disabled for some time.

scripts/codegen.sh
```diff
@@ -18,14 +18,11 @@ fi
 # Regenerate code
 ./gradlew -p codegen/sdk-codegen build
 
-# Delete the current autogenerated SDK code
-rm -rf release/
```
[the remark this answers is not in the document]
ganeshnj: if there is "left over", we need to fix root cause than duct taping by removing release/services dir. 

bigger workspace is an issue and by adding tests we are amplifying it. I don't have solution for it. Potentially making multiple Packages.swift for each service but solution to this problem is not related to this PR.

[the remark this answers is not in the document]
jbelkins: Deleting previous generated code isn't duct taping, it should be an integral part of this process because it prevents artifacts from previous generation of the project from compiling if they're not explicitly overwritten.

If you don't want to delete artifacts from previous code generation before staging the new ones, then that means if we ever remove an AWS service or even a source file within a service, we have to come up with a way to get rid of it if we don't delete all of the previous ones.

If you don't want to have to write `rm -rf ...` into several Github workflows and build scripts, then we should do it at the start of the `stageSdks` Gradle task.

Package.swift
```diff
@@ -358,358 +361,686 @@ let package = Package(
...
-        .target(name: "AWSWorkSpacesWeb", dependencies: [.product(name: "ClientRuntime", package: "smithy-swift"), "AWSClientRuntime"], path: "./release/AWSWorkSpacesWeb"),
-        .target(name: "AWSXRay", dependencies: [.product(name: "ClientRuntime", package: "smithy-swift"), "AWSClientRuntime"], path: "./release/AWSXRay"),
+        // MARK: - Core Targets
+        .target(name: "AWSClientRuntime", dependencies: [clientRuntime], path: "./Sources/Core/AWSClientRuntime"),
+
+        // MARK: - Core Test Targets
+        .testTarget(name: "AWSClientRuntimeTests", dependencies: ["AWSClientRuntime", smithyTestUtil], path: "./Tests/Core/AWSClientRuntimeTests"),
+
+        // MARK: - Service Targets
+        .target(name: "AWSACM", dependencies: ["AWSClientRuntime"], path: "./Sources/Services/AWSACM"),
```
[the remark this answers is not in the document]
jbelkins: I will go along with this structure for now, but we should remain open to using the official SPM structure if needed for compatibility down the road.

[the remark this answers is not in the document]
ganeshnj: SPM doesn't dictate how repo must be setup, but yes if this happens to be the case in future, we can definitely make adjustments. There needs to be balance of what makes our life better and what is standard.

scripts/codegen.sh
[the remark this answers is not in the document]
ganeshnj: Sorry, I don't agree with deletion. Specially in a case where we a failure in the codegen (if skipped) for a (more services) would lead to a removal of service.

If we remove a file/service it must be manual as accidental removal of something will lead to catastrophic outcomes. 

For a legit service removal, we must delete the model and delete the generated code manually.

jbelkins (approved):

PR #818, opened 2023-01-11
feat: add SigV4a signing support

### LINKED ISSUE #355: SigV4a Signer
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue, please leave a comment

## Describe the Feature

Investigate the sigv4a work done by the CRT team and bind any necessary new calls that they implemented. Expose the signer upwards and make some determination about when to use sigv4 vs sigv4a because sigv4a takes longer and will have performance effects.

Acceptance criteria:
* Expose CRT implemented sigv4a signer to swift
* Investigate whether or not smithy denotes when to use sigv4a or not.
  * implement code generators appropriately

ganeshnj:
## Issue \#

Fixes #355 

## Description of changes

When resolving endpoint, we get auth schemes which contains which a unstructured bag of data. This is purposefully unstructured so it can be involved in future. No customer is expected to logic around it and recommend to not use it.

Currently we only look for `sigv4`, if this auth scheme exists, pipeline continues else fail. With this change, we now look for first auth scheme (per recommendation, auth scheme are ordered) and use it for region, name and algorithm.

This change has been tested with [MRAP](https://aws.amazon.com/s3/features/multi-region-access-points/) which is a S3 feature. It allows accessing bucket content irrespective of the region application running on.

Example snippet
```swift
let config = try S3Client.S3ClientConfiguration(disableMultiRegionAccessPoints: false, region: "us-west-2")
let client = S3Client(config: config)

let output = try await client.getObject(input: GetObjectInput(bucket: "arn:aws:s3::727033484140:accesspoint/m4tkad9k7m655.mrap", key: "foobar"))
print(output)
```

In this case, foobar object exists in a bucket which is located in `us-east-1`.

None

## Conventional Commits


Sources/Core/AWSClientRuntime/Endpoints/AWSEndpoint.swift
```diff
@@ -58,6 +58,11 @@ public struct AWSEndpoint: Equatable {
 }
 
 extension Endpoint {
+
+    /// Returns the auth scheme for the given name
+    /// This is an internal API and subject to change without notice
```
epau, line 63: Not for this PR: perhaps we should prefix internal API's to make that more clear and consistent. Such as with a `_` like `_authScheme`

Sources/Core/AWSClientRuntime/Endpoints/AWSEndpoint.swift
```diff
@@ -70,4 +75,48 @@ extension Endpoint {
         
         return scheme
     }
-}
+
+    /// Returns the first auth scheme
+    /// This is an internal API and subject to change without notice
+    /// - Returns: first auth scheme if present
+    public func firstAuthScheme() -> [String: Any]? {
```
epau, line 82: Nit: Removing this function and replacing it with a function that returns authSchemes would help reduce key duplication (the `authScheme(name: String)` function also accesses authSchemes), and since the return type is an array, callers already know how to get the first item in an array so no need for an explicit function to do so. 

```swift
public func authSchemes() -> [[String: Any]]? {
    return properties["authSchemes"] as? [[String: Any]]
}

let firstAuthScheme = authSchemes()?.first
```

epau (approved):
Looks great! Love the docs and tests! 

imo it'd be nice to have the dictionary keys defined in some embedded type. but not required and something we can always do in the future.

```swift
enum PropertyKeys {
    static let authSchemes = "authSchemes"
}

enum AuthSchemeKeys {
    static let signingRegion = "signingRegion"
    static let signingRegionSet = "signingRegionSet" 
    // etc...
}
```

Sources/Core/AWSClientRuntime/Endpoints/AWSEndpoint.swift
```diff
@@ -58,6 +58,11 @@ public struct AWSEndpoint: Equatable {
 }
 
 extension Endpoint {
+
+    /// Returns the auth scheme for the given name
+    /// This is an internal API and subject to change without notice
+    /// - Parameter name: Name of the auth scheme
+    /// - Returns: auth scheme for given auth scheme name if present
     public func authScheme(name: String) -> [String: Any]? {
```
jbelkins, line 66: I'm guessing that `authScheme` comes from CRT, but we should try to avoid types that are just a dictionary of stuff if we can.

Sources/Core/AWSClientRuntime/Signing/AWSSigV4Signer.swift
```diff
@@ -11,14 +11,15 @@ import Foundation
 
 public class AWSSigV4Signer {
```
jbelkins, line 12: Since this type now handles `sigv4` and `sigv4a`, should it be renamed for clarity?

Sources/Core/AWSClientRuntime/Signing/AWSSigningAlgorithm.swift
```diff
@@ -6,14 +6,16 @@
 //
 import AwsCommonRuntimeKit
 
-public enum AWSSigningAlgorithm {
+public enum AWSSigningAlgorithm: String {
```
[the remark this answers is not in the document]
jbelkins: In must cases I'd prefer a method that converts an enum value to a string for use in a specific place over just making an enum have String raw value.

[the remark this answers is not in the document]
jbelkins: Also even if they're simple, this public type could use docs

jbelkins (changes requested):

Sources/Core/AWSClientRuntime/Endpoints/AWSEndpoint.swift
ganeshnj: This is interesting suggestion but my worry is, anything with _ prefix is used for private access only. Having something with _ and public would make things harder for reviews.

Rather I would like to have something like @internal attribute which is more explicit. Sadly, Swift doesn't have custom attribute support yet.

https://forums.swift.org/t/pitch-introduce-custom-attributes/21335/21

Sources/Core/AWSClientRuntime/Endpoints/AWSEndpoint.swift
ganeshnj: This has been discussed intensively and we don't to type it. This is purposefully kept a free form bag of data. Hence, we safely try to extract if there is anything.

Sources/Core/AWSClientRuntime/Signing/AWSSigV4Signer.swift
ganeshnj: It is still SigV4, we ever have V5, then we will add another type eventually.

Sources/Core/AWSClientRuntime/Signing/AWSSigningAlgorithm.swift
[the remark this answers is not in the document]
ganeshnj: https://github.com/awslabs/aws-sdk-swift/blob/3f437ce63c2f09900b3fe8864425ca27170cc50f/Sources/Core/AWSClientRuntime/HttpContextBuilder+Extension.swift#L33-L38

It is saved in HttpContext to allow sharing this across middlewares. I would prefer it to be a string than any other type. Easy to investigate and debug.

Sources/Core/AWSClientRuntime/Endpoints/AWSEndpoint.swift
ganeshnj: > Looks great! Love the docs and tests!
> 
> imo it'd be nice to have the dictionary keys defined in some embedded type. but not required and something we can always do in the future.
> 
> ```swift
> enum PropertyKeys {
>     static let authSchemes = "authSchemes"
> }
> 
> enum AuthSchemeKeys {
>     static let signingRegion = "signingRegion"
>     static let signingRegionSet = "signingRegionSet" 
>     // etc...
> }
> ```

adapted to suggestion

codegen/smithy-aws-swift-codegen/src/main/kotlin/software/amazon/smithy/aws/swift/codegen/middleware/EndpointResolverMiddleware.kt
```diff
@@ -53,9 +53,12 @@ class EndpointResolverMiddleware(
...
 
-        writer.write("""let authScheme = endpoint.authScheme(name: "sigv4")""")
-
-        writer.write("""let awsEndpoint = AWSEndpoint(endpoint: endpoint, signingName: authScheme?["signingName"] as? String, signingRegion: authScheme?["signingRegion"] as? String)""")
+        writer.write("""let authScheme = endpoint.firstAuthScheme()""")
+        writer.write("""let signingName = Endpoint.signingName(from: authScheme)""")
+        writer.write("""let signingRegion = Endpoint.signingRegion(from: authScheme)""")
+        writer.write("""let signingAlgorithm = Endpoint.signingAlgorithm(from: authScheme)""")
+            .write("")
+        writer.write("""let awsEndpoint = AWSEndpoint(endpoint: endpoint, signingName: signingName, signingRegion: signingRegion)""")
```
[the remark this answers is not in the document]
ganeshnj: copy-paste, but triple quotes are very useful when you have tones of string literal. escaping isn't required for them.

Sources/Core/AWSClientRuntime/Signing/AWSSigningAlgorithm.swift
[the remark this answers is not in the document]
ganeshnj: >In must cases I'd prefer a method that converts an enum value to a string for use in a specific place over just making an enum have String raw value.

Any particular reason where runtime has direct support for it. If needed you can modify the raw value.

codegen/smithy-aws-swift-codegen/src/main/kotlin/software/amazon/smithy/aws/swift/codegen/middleware/EndpointResolverMiddleware.kt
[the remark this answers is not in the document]
jbelkins: Yeah, the deleted code had string literals but the code added in its place had none.  That's why I asked.

Sources/Core/AWSClientRuntime/Signing/AWSSigningAlgorithm.swift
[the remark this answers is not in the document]
jbelkins: If the raw values are supposed to correspond to certain strings (i.e. some string identifier in Smithy or CRT) that should be noted in the documentation for the type.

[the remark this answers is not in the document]
ganeshnj: added

codegen/smithy-aws-swift-codegen/src/main/kotlin/software/amazon/smithy/aws/swift/codegen/middleware/EndpointResolverMiddleware.kt
```diff
@@ -53,9 +53,12 @@ class EndpointResolverMiddleware(
...
-        writer.write("""let awsEndpoint = AWSEndpoint(endpoint: endpoint, signingName: authScheme?["signingName"] as? String, signingRegion: authScheme?["signingRegion"] as? String)""")
+        writer.write("let authScheme = endpoint.authSchemes()?.first")
+        writer.write("let signingName = Endpoint.signingName(from: authScheme)")
+        writer.write("let signingRegion = Endpoint.signingRegion(from: authScheme)")
+        writer.write("let signingAlgorithm = Endpoint.signingAlgorithm(from: authScheme)")
+            .write("")
+        writer.write("let awsEndpoint = AWSEndpoint(endpoint: endpoint, signingName: signingName, signingRegion: signingRegion)")
             .write("")
 
         writer.write("""var host = """"")
```
jbelkins, line 64: I know what this is doing, but that doesn't mean it doesn't make my head spin

Sources/Core/AWSClientRuntime/Endpoints/AWSEndpoint.swift
epau: Unfortunately that proposal doesn't have much traction these days and there doesn't appear to be a timeline of when, or even if, it will get completed.

We don't necessarily need to use a preceding `_`, but i would like us to come up with a convention that will make it easy for us to identify cases like this when reading code. Otherwise imo we should remove the "for internal use only comment" as there is nothing preventing non-internal use.

codegen/smithy-aws-swift-codegen/src/main/kotlin/software/amazon/smithy/aws/swift/codegen/middleware/EndpointResolverMiddleware.kt
ganeshnj: this is definitely better than escaping every quote. Specially good at you can just copy paste when you want to emit exact Swift code without any substitution.

jbelkins: In this specific case I would consider it fine, maybe even better, to escape two double quote characters.  But no doubt that if I had a big blob of JSON or the like, containing numerous double quote characters, I would contain my string in triple quotes.  And don't get me started on including a literal `$` in a Kotlin multiline string...

Sources/Core/AWSClientRuntime/Endpoints/AWSEndpoint.swift
ganeshnj: no matter what, nothing stops customers to use this method, it is public.

Intent is to communicate. Either via a comment, an attribute or something else. So in future if just remove this, we can say that this was expected.

Most SDKs struggle with this, there is no good way to share code without making it public. Some namespace it with `internal`, some add attribute or some add comment.

jbelkins: More fundamentally, is there a reason we need this disclaimer on this function?  The function is documented and does a pretty straightforward thing.  Why treat it different from any other public function?

ganeshnj: that's because of nature of data it deals with. EndpointRuleSet are expected to evolve and setting an expectation on a dynamic data is not good.

We have had trouble in past, for example in some SDKs we support giving list of supported region. This seems fair at high level but gets complicated to maintain. For example, if the source of data goes away and replaced by something that is not compatible.

jbelkins: I will approve this because it follows the existing pattern in the project, but we should revisit it.  The fact that we feel the need to declare something "public but not public", modularization or not, suggests that something is wrong with the patterns we are employing.

I think doing a better job of encapsulating these fields & data in types meant to hold and work with them will prevent us from having to make these types of declarations and make generated code simpler since we won't have to pick through individual data fields when configuring reusable types like middleware.

jbelkins (approved):

PR #819, opened 2023-01-12
chore: remove trailing whitespaces in all Swift files

ganeshnj:
## Description of changes

Getting rid of some annoyance.

```
find . -name "*.swift" -print0 | xargs -0 sed -i '' -E "s/[[:space:]]*$//"
```

## Conventional Commits


.swiftlint.yml
```diff
@@ -6,7 +6,6 @@ analyzer_rules:
   - unused_declaration
 
 disabled_rules:
-  - trailing_whitespace
```
jbelkins, line 9: Why was this rule disabled in the first place?  I generally hate picky linter settings, but this is one of the few rules I would want to have.

jbelkins (approved):

.swiftlint.yml
ganeshnj: likely because there were tons of places to be fixed

Issue #820, opened 2023-01-12
Redundant aws-crt-swift in aws-sdk-swift package 


ganeshnj:
### Describe the bug

smithy-swift already takes a dep on aws-crt-swift and it is again added to aws-sdk-swift which is redundant. 
```
'smithy-swift' dependency on 'https://github.com/awslabs/aws-crt-swift.git' conflicts with dependency on '/Users/jangirg/Projects/Amplify/SwiftSDK/aws-crt-swift' which has the same identity 'aws-crt-swift'. this will be escalated to an error in future versions of SwiftPM.
```

### Expected Behavior

Warning must not be shown.

### Current Behavior

```
'smithy-swift' dependency on 'https://github.com/awslabs/aws-crt-swift.git' conflicts with dependency on '/Users/jangirg/Projects/Amplify/SwiftSDK/aws-crt-swift' which has the same identity 'aws-crt-swift'. this will be escalated to an error in future versions of SwiftPM.
```

### Reproduction Steps

- Remove `aws-crt-swift` dep from `aws-sdk-swift` and let SPM take care of transitive dependency that comes along `smithy-swift`.
- Need to update Package.swift generation script

### AWS SWIFT SDK version used

main

### Compiler and Version used

5.7.2

### Operating System and version

macOS 12.6.1

PR #822, opened 2023-01-13
fix: use same aws-crt-swift version in aws-sdk-swift as smithy-swift

ganeshnj:
## Issue \#

Fixes https://github.com/awslabs/aws-sdk-swift/issues/820

### Motivation

aws-crt-swift added to aws-sdk-swift and smithy-swift, hence we have to update in both packages. Additionally, SPM started throwing warning on double inclusion of the dependency which will be covered to error in future versions.

Also, aws-crt-swift used as main and specific version smithy-swift which can cause further confusion.

### Modifications

- aws-crt-swift version is same on both aws-sdk-swift and smithy-swift
- every target list their dependencies explicitly rather depending on the transitive deps.

### Result

no warnings in SPM

In CI following 
```swift
let useLocalDeps = ProcessInfo.processInfo.environment["AWS_SWIFT_SDK_USE_LOCAL_DEPS"] != nil
let useMainDeps = ProcessInfo.processInfo.environment["AWS_SWIFT_SDK_USE_MAIN_DEPS"] != nil

switch (useLocalDeps, useMainDeps) {
case (true, true):
    fatalError("Unable to determine which dependencies to use. Please only specify one of AWS_SWIFT_SDK_USE_LOCAL_DEPS or AWS_SWIFT_SDK_USE_MAIN_DEPS.")
case (true, false):
    package.dependencies += [
        .package(path: "../smithy-swift")
    ]
case (false, true):
    package.dependencies += [
        .package(url: "https://github.com/awslabs/smithy-swift", branch: "main")
    ]
case (false, false):
    package.dependencies += [
        .package(url: "https://github.com/awslabs/smithy-swift", .exact("0.10.0"))
    ]
}
```

translated to
```swift
 let useLocalDeps = ProcessInfo.processInfo.environment["AWS_SWIFT_SDK_USE_LOCAL_DEPS"] != nil
let useMainDeps = ProcessInfo.processInfo.environment["AWS_SWIFT_SDK_USE_MAIN_DEPS"] != nil

switch (useLocalDeps, useMainDeps) {
case (true, true):
    fatalError("Unable to determine which dependencies to use. Please only specify one of AWS_SWIFT_SDK_USE_LOCAL_DEPS or AWS_SWIFT_SDK_USE_MAIN_DEPS.")
case (true, false):
    package.dependencies += [
        .package(path: "/Users/runner/work/aws-sdk-swift/aws-sdk-swift/target/build/deps/smithy-swift")
    ]
case (false, true):
    package.dependencies += [
        .package(url: "https://github.com/awslabs/smithy-swift", branch: "main")
    ]
case (false, false):
    package.dependencies += [
        .package(url: "https://github.com/awslabs/smithy-swift", .exact("0.10.0"))
    ]
}
```

## Conventional Commits


epau (approved):

scripts/generatePackageSwift.swift
```diff
@@ -37,13 +37,6 @@ struct PackageDeps: Codable {
     var clientRuntimePath: String?
 }
 
-// When AWS_SDK_RELEASE_IN_PROGRESS is set, the manifest will require
-// exact versions of smithy-swift and aws-crt-swift no matter what
-// branches or paths are set.
-let releaseInProgress: Bool = {
-    env["AWS_SDK_RELEASE_IN_PROGRESS"] != nil
```
jbelkins, line 44: If this env var is no longer being read from, then we should remove the place where it is set in our release workflow

scripts/setup_for_crt_builder.sh
```diff
@@ -6,6 +6,10 @@ set -x
 # with the CRT builder script (see https://github.com/awslabs/aws-crt-builder).
 # For use only as a pre-build step when using the CRT builder script on CI.
 
+# Enables the use of local dependencies, which are required when building with
+# the CRT builder script.
+export AWS_SWIFT_SDK_USE_LOCAL_DEPS=1
```
jbelkins, line 11: Releases use the CRT builder as well.  If this is set for a release, then a release will use whatever reference CRT builder retrieves & installs by default, not the pinned versions in `packageDependencies.plist`

jbelkins (changes requested):
Not sure this will work properly for release.

scripts/generatePackageSwift.swift
ganeshnj: good call, removed.

scripts/setup_for_crt_builder.sh
ganeshnj: not entirely,

what this switch do, is basically manipulate 
```
let useLocalDeps = ProcessInfo.processInfo.environment["AWS_SWIFT_SDK_USE_LOCAL_DEPS"] != nil
let useMainDeps = ProcessInfo.processInfo.environment["AWS_SWIFT_SDK_USE_MAIN_DEPS"] != nil

switch (useLocalDeps, useMainDeps) {
case (true, true):
    fatalError("Unable to determine which dependencies to use. Please only specify one of AWS_SWIFT_SDK_USE_LOCAL_DEPS or AWS_SWIFT_SDK_USE_MAIN_DEPS.")
case (true, false):
    package.dependencies += [
        .package(path: "../smithy-swift") <------- THIS
    ]
case (false, true):
    package.dependencies += [
        .package(url: "https://github.com/awslabs/smithy-swift", branch: "main")
    ]
case (false, false):
    package.dependencies += [
        .package(url: "https://github.com/awslabs/smithy-swift", .exact("0.10.0"))
    ]
}
```

change the path to the dir where repo is located in the runner.

This code goes out in released Package.swift and when both AWS_SWIFT_SDK_USE_LOCAL_DEPS and AWS_SWIFT_SDK_USE_MAIN_DEPS are unset, which we expect to be in customer apps, we will use the pinned version.

jbelkins (approved):

PR #824, opened 2023-01-18
fix: Delete previous generated code before staging

jbelkins:
## Description of changes
Deletes previous generated code before staging new generated code.  Fixes the following problems:
- Ensures that previously generated code doesn't persist to a new generation if it doesn't happen to be explicitly overwritten
- Ensures that source files are deleted when one or more services are excluded from code-generation during development

This was previously done in scattered places around the codebase including CI workflows and test scripts; performing it as part of the `stageSdks` Gradle task DRYs up code by not requiring it to be performed in multiple places.


jbelkins:
Closing to determine a more selective method of doing this.

PR #827, opened 2023-01-18
Release 0.9.0

jbelkins:
## Conventional Commits

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -42,7 +42,7 @@ jobs:
         run: |
           python3 -c "from urllib.request import urlretrieve; urlretrieve('${{ env.BUILDER_HOST }}/${{ env.BUILDER_SOURCE }}/${{ env.BUILDER_VERSION }}/builder.pyz?run=${{ env.RUN }}', 'builder.pyz')"
           chmod a+x builder.pyz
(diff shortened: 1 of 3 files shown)


.github/workflows/release.yml
```diff
@@ -51,7 +51,6 @@ jobs:
           ./gradlew -p codegen/sdk-codegen stageSdks
           ./gradlew --stop
           ./scripts/mergeModels.sh Sources/Services
-      - name: Generate Package.swift
```
ganeshnj, line 54: this should be
```
- name: Generate Package.swift
   run: |

Comment (jbelkins) [reply .github/workflows/release.yml:54]: I simply removed it for now, as it was the minimum change that allows the workflow to parse correctly, and it reverts this workflow to the state it was in for the previous release which was known to work correctly.

File: .github/workflows/release.yml [hunk 2]
```diff
@@ -62,4 +61,4 @@ jobs:
           author_name: AWS CI
           author_email: nobody@amazonaws.com
           message: 'aws-sdk-swift release build artifacts'
-          add: 'release Package.swift'
+          add: 'Sources/Services/* Package.swift'
```
ganeshnj, line 64: Sources/Services/* Tests/Services/* Package.swift

.github/workflows/release.yml
ganeshnj: but forces to share the env var across steps, we don't want to pass env var declared in Build and Test with Generate Package.swift

jbelkins: Understood.  Splitting this into a separate step so `CI_DIR` env vars aren't set when the release `Package.swift` is generated

.github/workflows/release.yml
jbelkins: Changed

.github/workflows/release.yml
ganeshnj: nit: we are only Generating the manifest, SDK generation has been done in earlier step.

ganeshnj (approved):

Issue #808, opened 2023-01-05
Regenerate models and build the full SDK as part of PR checks


epau:
### Describe the feature

We are currently prone to build issues slipping into `main` since we don't currently regenerate the models and build the full SDK on PRs. This leaves this test up to the developer to perform manually. 

Let's add this as part of the CI workflow that is run on every PR but work will need to be done to make this as efficient as possible so we don't negatively impact productivity too much. Perhaps we should define some SLA that CI needs to execute under, such as 30 minutes.

### Use Case

N/A

### Acknowledgements

- [ ] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

ganeshnj:
Resolving in favor of catapult integration.

epau:
Re-opening since this deals with catching issues during development which is unrelated to catapult.

sichanyoo:
There exists `codegen-build-test` CI job now that does this. Closing ticket now.

PR #829, opened 2023-01-19
Delete generated services & service tests before running CI

jbelkins:
## Description of changes
Prevents the entire SDK from being built and tested during CI runs by deleting generated services and their test suites before starting CI.  Then, when the SDK is built & tested, only runtime components are built and tested.

While testing the full, built AWS SDK is generally desirable, it takes too much time as-is to complete during routine CI runs.  #808 tracks the need to establish a CI task to monitor the health of the full SDK.


builder.json
```diff
@@ -6,7 +6,7 @@
     "gradlew": "{source_dir}/gradlew"
   },
   "pre_build_steps": [
-    "cd {source_dir}",
```
ganeshnj, line 9: could you add log (plain echo) here, why are we doing this for posterity.

ganeshnj (approved):

epau (approved):
assuming ganesh's comment is addressed, lgtm

builder.json
ganeshnj: and put a link to https://github.com/awslabs/aws-sdk-swift/issues/808

jbelkins: Done

Issue #838, opened 2023-01-24
Request fails when endpoint rule set does not specify signing algorithm


jbelkins:
### Describe the bug

When making a request on a service which does not specify a signing algorithm in its endpoint rule set, the SDK fails to send the request with this error:
```
Error: client(ClientRuntime.ClientError.authError("AwsSigv4Signer requires a signing algorithm"), nil)
```

Example operation is `com.amazonaws.kendraranking#ListRescoreExecutionPlans` (this operation works on AWS CLI):
https://github.com/awslabs/aws-sdk-swift/blob/main/codegen/sdk-codegen/aws-models/kendra-ranking.2022-10-19.json#L894

### Expected Behavior

Request succeeds

### Current Behavior

Request fails to send with error thrown as listed above

### Reproduction Steps

```
import AWSKendraRanking

        let client = try KendraRankingClient(region: "us-east-1")
        let input = ListRescoreExecutionPlansInput(maxResults: 10, nextToken: "abc123")
        do {
            let output = try await client.listRescoreExecutionPlans(input: input)
            print("Request succeeded")
        } catch {
            print("Request failed")
            print("Error: \(error)")
        }
```

### Possible Solution

Supply default signing algorithm when none is specified in the endpoint rule set.

### AWS SWIFT SDK version used

main branch

### Compiler and Version used

Xcode 14.2

### Operating System and version

macOS 12.6.2

PR #839, opened 2023-01-25
fix: default to operation config when endpoint doesn't have auth schemes

ganeshnj:
## Issue \#

#838 

There will be separate commit for codegen, skipping it for better review experience.

### Motivation
During Endpoint 2.0 implementation we only had sigv4 hence all requests were going through sigv4a irrespective of the operation needs to not.

With SigV4A, we started looking into the auth schemes supplied via the endpoints. Some services don't have endpoints 2.0 models which ends up having no signing and result in the error.

### Changes

If there is no auth scheme in endpoint, use the singing algorithm supported by the service/operation.

Cleaned up some auth scheme parsing logic with their backing structs, these are "public-internal" and must not be used by customers as we expect them to change/break.

### Result

Service like KendraRanking works which don't have authscheme in their endpoint data. 

## Conventional Commits


epau (changes requested):
This looks really good but I think we lean on the compiler and type system a bit more to "compile away" some issues. 

#### Auth Schemes
Currently, `AuthScheme` is a bit of a loaded term. We have an enum `AuthScheme` which represents the supported auth schemes. We also have individual types that represent auth schemes, such as SigV4AuthScheme, NoneAuthScheme etc. We have two ways to represent auth schemes, which brings into question which is the right way to define an auth scheme? 
`let authScheme = AuthScheme.sigv4(SigV4AuthScheme())`
OR
`let authScheme = SigV4AuthScheme()`

Additionally, a developer could define additional auth scheme models and not add them to the enum AuthScheme. In that scenario it becomes unclear what is supported or not. 

#### Auth Scheme Names
We use auth scheme names to be able to define supported auth schemes for our default resolver. Additionally, names are how we determine if an auth scheme is valid when transforming a dictionary into an auth scheme. Because we're using raw strings to define names, we have multiple places where we duplicate the name values and we don't have a consistent list that defines all the supported names. 

#### Proposed changes
[Here's a gist ](https://gist.github.com/epau/bfea1f4f3ee51918e023268b0147dc8b) with some proposed modelling changes that eliminate unnecessary types and consolidate types so that naming and purpose are more clear. This is somewhat psuedo code so please feel free to modify.
- Defines a `Name` enum that defines all supported scheme names
- Removes IdentifiableAuthScheme as it's no longer needed
- Removes NoneAuthScheme as it's unnecessary
- Replaces `SigV4AuthScheme` and `SigV4AAuthScheme` with `SigV4Parameters` and `SigV4AParameters` (please feel free to rename, but let's avoid using the `AuthScheme` suffix.

codegen/smithy-aws-swift-codegen/src/main/kotlin/software/amazon/smithy/aws/swift/codegen/middleware/EndpointResolverMiddleware.kt
```diff
@@ -53,11 +60,32 @@ class EndpointResolverMiddleware(
...
+            writer.write("signingRegion = data.signingRegionSet?.first")
+            writer.write("signingAlgorithm = data.name")
+            writer.dedent()
+            writer.write("case .none:")
+            writer.indent()
+            writer.write("break")
+            writer.dedent()
+            writer.write("}")
+        }
+        writer.write("")
```
jbelkins, line 88: This is a big chunk of logic to be codegen-ed for every service.  Any way we can get this in runtime, where it can be shared among services and placed under functional test?

codegen/smithy-aws-swift-codegen/src/main/kotlin/software/amazon/smithy/aws/swift/codegen/AWSSigningParams.kt
```diff
@@ -22,16 +22,29 @@ data class AWSSigningParams(
...
     // When set to true, the expiration is included in the params string so it may be set to a custom
     // value named `expiration`.
-    val useExpiration: Boolean
+    val useExpiration: Boolean,
+    // The signing algorithm to use which is determined by the service's SigV4Trait.
+    val signingAlgorithm: SigningAlgorithm
 )
 
+enum class SigningAlgorithm {
+    SigV4
```
jbelkins, line 31: Why is sigv4a not an option here?

jbelkins (changes requested):

ganeshnj:
> This looks really good but I think we can lean on the compiler and type system a bit more to "compile away" some issues.
> #### Auth Schemes
> 
> Currently, `AuthScheme` is a bit of a loaded term. We have an enum `AuthScheme` which represents the supported auth schemes. We also have individual types that represent auth schemes, such as SigV4AuthScheme, NoneAuthScheme etc. We have two ways to represent auth schemes, which brings into question which is the right way to define an auth scheme? `let authScheme = AuthScheme.sigv4(SigV4AuthScheme())` OR `let authScheme = SigV4AuthScheme()`
> 
> Additionally, a developer could define additional auth scheme models and not add them to the enum AuthScheme. In that scenario it becomes unclear what is supported or not.
> #### Auth Scheme Names
> 
> We use auth scheme names to be able to define supported auth schemes for our default resolver. Additionally, names are how we determine if an auth scheme is valid when transforming a dictionary into an auth scheme. Because we're using raw strings to define names, we have multiple places where we duplicate the name values and we don't have a consistent list that defines all the supported names.
> #### Proposed changes
> 
> [Here's a gist ](https://gist.github.com/epau/bfea1f4f3ee51918e023268b0147dc8b) with some proposed modelling changes that eliminate unnecessary types and consolidate types so that naming and purpose are more clear. This is somewhat psuedo code so please feel free to modify.
> 
>     * Defines a `Name` enum that defines all supported scheme names
> 
>     * Removes IdentifiableAuthScheme as it's no longer needed
> 
>     * Removes NoneAuthScheme as it's unnecessary
> 
>     * Replaces `SigV4AuthScheme` and `SigV4AAuthScheme` with `SigV4Parameters` and `SigV4AParameters` (please feel free to rename, but let's avoid using the `AuthScheme` suffix.

Mostly taken except the Name enum as this is overkill where a string literal is suffice.

codegen/smithy-aws-swift-codegen/src/main/kotlin/software/amazon/smithy/aws/swift/codegen/middleware/EndpointResolverMiddleware.kt
epau: can we file an issue for this? i agree we shouldn't hold up the PR for it, but it'd be good to have an issue since imo its a pretty straightforward one that anyone can pick up if they have extra time.

epau (approved):
looks great, thanks for the changes!

codegen/smithy-aws-swift-codegen/src/main/kotlin/software/amazon/smithy/aws/swift/codegen/middleware/EndpointResolverMiddleware.kt
ganeshnj: it's not straightforward as we just don't want to extract the logic but want to extract the middleware outside the service client.

Indeed https://github.com/awslabs/aws-sdk-swift/issues/840

jbelkins (approved):

PR #883, opened 2023-02-15
chore: creates pull-request comment triggers for CI jobs

epau:
This PR configures the following github workflows to be triggered by commenting on a PR. 

Release Workflow
- can now also be triggered by commenting `/release` on the PR and the branch name starts with `release/`

CI Workflow
- can now also be triggered by commenting `/ci` on the PR

Lint Workflow
- can now also be triggered by commenting `/ci` on the PR

Unfortunately, this cannot be tested until these changes are present in the default branch. After merging, i will verify behavior.

## Conventional Commits

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,8 +1,9 @@
 name: Release
 
 on:
-  pull_request:
   workflow_dispatch:
+  issue_comment:
+    types: [created]
 
 env:
   BUILDER_VERSION: v0.8.19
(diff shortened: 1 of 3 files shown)


.github/workflows/release.yml
```diff
@@ -22,7 +23,8 @@ env:
 
 jobs:
   macos-release:
-    if:  startsWith(github.head_ref, 'release/')
+    # if the workflow was manually dispatched, or if someone commented on the PR with /release 
+    if: github.event_name == 'workflow_dispatch' || (startsWith(github.ref_name, 'release/') && github.event.issue.pull_request && startsWith(github.event.comment.body, '/release'))
```
jbelkins, line 27: Do we even still need the `startsWith(github.ref_name, 'release/')` condition?  We can create a release branch, then manually trigger with a comment when desired.

This would also prevent us from having to stop the release workflow when a change is pushed to the release branch that is not intended to run the release workflow.

epau: I kept in the release branch restriction so that the `/release` comment will only work on release branch PRs. This adds a bit of safety from accidentally triggering the release workflow on other PRs.

jbelkins (approved):

ganeshnj:
Can you test at least working on fork of the repo?

epau:
Closing this for now as this doesn't update the actions on the PR. 
We'll need to look into exactly what we need to do here, i don't have time for this at the moment but can pick up later.

Issue #886, opened 2023-02-22
Add tests for S3 errors


epau:
### Describe the feature

Add a test to verify the S3 error customizations. 
Specifically we want a test that verifies an error with status code 404 and an empty response body decodes successfully.

It might make sense to wrap this functionality in a "S3Error" type and add a single test for that. Otherwise, we can add a test for all S3 error types or hand pick a couple.

Related Issue: #761 
Related PR: #787 

### Use Case

N/A

### Acknowledgements

- [ ] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

ganeshnj:
It must be done after #307

jbelkins:
We have a test for the S3 "empty body 404 error" case already.  Closing.

Issue #814, opened 2023-01-10
Enable generated service test targets


ganeshnj:
### Describe the bug

Currently we have disabled the test targets because swift CLI fails to run them and errors out.

### Expected Behavior

`swift test` must run service test targets.

### Current Behavior

Fatal fail.

### Reproduction Steps

- in Package.swift, uncomment testTargets for services
- swift test  

### AWS SWIFT SDK version used

man

### Compiler and Version used

5.7.2

### Operating System and version

macOS 12.6.1

ganeshnj:
We can try using xcodebuild for just on macs to atleast run on one OS. But we would like to stick to swift cli.

smivij:
This is done now.

PR #887, opened 2023-02-22
test: enables service tests

epau:
This PR enables service tests. Previously, they were disabled due to an issue with CRT. 
CRT would hang on a call to cleanup, because CRT objects were still being held in memory by the SDKDefaultIO singleton. 

Additionally, this PR removes the following service tests.
- S3 **HeadObjectErrorTest** and **HeadBucketErrorTests**. These tests were not accurate. An issue has been created to create new tests that should take their place https://github.com/awslabs/aws-sdk-swift/issues/886
- S3 **PutObjectTests**. These tests are not actually testing any business logic. They are too coupled to the test configuration and seem to just be testing the details in the test. 
- Route53 tests. Similar to the S3 tests, these aren't actually testing any business logic and appear to just be testing the operation stack configuration in the test itself. 

We should prioritize implementing interceptors as interceptors should be the mechanism used by service tests to verify behavior. For example, to test that a certain operation decodes a specific response successfully, we should add an interceptor after the deserialization step that verifies the deserialized output is as we expect. Currently, we duplicate the entire operation stack in the test which is pointless as it doesn't actually test the specific operation.

## Issue \#
#814 

## Conventional Commits


ganeshnj:
per https://www.conventionalcommits.org/en/v1.0.0/#specification this seems to fit more into `test` category

Tests/Services/AWSRoute53Tests/ListResourceRecordSetsRequestTest.swift
```diff
@@ -1,65 +0,0 @@
-// Code generated by smithy-swift-codegen. DO NOT EDIT!
-
-@testable import AWSRoute53
```
ganeshnj, line 3: do we have a backlog to add it back?

codegen/sdk-codegen/aws-models-test/s3/s3-tests.smithy
```diff
@@ -1,83 +0,0 @@
-$version: "1.0"
```
ganeshnj, line 1: we must add a backlog, ideally a TODO in kt codebase.

ganeshnj (approved):

Tests/Services/AWSRoute53Tests/ListResourceRecordSetsRequestTest.swift
epau: Added one here https://github.com/awslabs/aws-sdk-swift/issues/888

Issue #901, opened 2023-03-07
Create mechanism to increment package version via CI


epau:
### Describe the feature

As part of #768 we need a consistent and reliable way to 
- retrieve the current greatest release version (which isn't always the most recent as we could have hotfixes)
- update the release version 

### Proposal

Create a `Package.version` file which is a text file that only contains the latest version number. Such as
```
0.11.0
```

The flow for updating the version via our release process is then as follows:
- Get the latest version from Package.version
- Set the patch version to 0 and increment the minor version
- Write the new version to Package.version

The logic for determining the next version number may change in the future. For now we simply increment the next patch number. 

We would also add this to the [smithy-swift repository](https://github.com/awslabs/smithy-swift).

### Use Case

N/A

### Acknowledgements

- [ ] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

PR #902, opened 2023-03-07
feat: Adds mechanism for tracking and updating release version

epau:
Creates `Package.version` for tracking the current/latest release version. 
Additionally, creates `UpdatePackageVersion.swift`, a script for updating the minor version of the release to be used during release automation.

## Issue \#
#901 

## Conventional Commits


Package.version
```diff
@@ -0,0 +1 @@
+0.11.0
```
epau, line 1: Opting for a simple plain text file now. 
We can change to `json`, `plist`, or some other structured format when the need arises.

scripts/UpdatePackageVersion.swift
```diff
@@ -0,0 +1,101 @@
...
+import class Foundation.FileManager
+
+// MARK: - Main
+
+/**
+Updates the the version located at the provided file path. 
+The file stored at the path should be a plain text file that only contains the version. 
+The version should be a simple semver formatted version number e.g. MAJOR.MINOR.PATCH
+*/
+func main() {
```
ganeshnj, line 24: this is fine but something to think about making swift packages for these scripts.

Then we can use packages like https://github.com/apple/swift-argument-parser which makes both developer and user life easier. A lot of boilerplate can be avoided.

scripts/UpdatePackageVersion.swift
```diff
@@ -0,0 +1,101 @@
...
+
+func printError(_ message: String) {
+  print(message)
+  exit(1)
+}
+
+struct Version {
+  var major: Int
+  var minor: Int
+  var patch: Int
```
ganeshnj, line 65: trivial: private(set)

scripts/UpdatePackageVersion.swift
ganeshnj: We could even do code sharing with that.

epau: 100%

I'm not tackling that rn given I'm trying to stay focused on getting the release automation wrapped up. 
I'll file an issue for this as I see this part of converting all of our scripts to swift. 

In my mind, we'll have a swift cli package (that pulls in argument parser) and will be the single entry point to all of our scripts. This will allow for code reuse (like you mentioned) as well as adding tests for our scripting logic as well.

scripts/UpdatePackageVersion.swift
epau: I'd like to allow these properties to be modified if needed. Also will make testing really easy. This is just a dumb simple representation of a Version, no reason to lock them down imo especially since theyre still immutable unless the caller creates a `var`.

epau:
Closing this in favor of https://github.com/awslabs/aws-sdk-swift/pull/909

PR #910, opened 2023-03-10
feat: event stream runtime support

ganeshnj:
### Motivation

To support event streaming, SDK needs implementations for `MessageEncoder`, `MessageDecoder`, `MessageDecoderStream` and `MessageEncoderStream`.

### Changes

- Add `AWSMessageEncoder` which is a `MessageEncoder` implementation along with its tests
- Add `AWSMessageDecoder` which is a `MessageDecoder` implementation along with its tests
- Add `AWSMessageDecoderStream` which is a `MessageDecoderStream` implementation along with its tests
- Add `AWSMessageEncoderStream` which is a `MessageEncoderStream` implementation along with its tests
- Integrate Payload signer in the SDK which is used in `MessageEncoderStream` implementation

### Result

All tests pass.

## Conventional Commits


Sources/Core/AWSClientRuntime/EventStream/AWSMessageDecoderStream.swift
```diff
@@ -0,0 +1,65 @@
...
+                            return message
+                        }
+                    }
+
+                    // this is the end of the stream
+                    // notify the decoder that the stream has ended
+                    try decoder.endOfStream()
+                    return nil
+                } catch {
+                    return nil
```
ganeshnj, line 56: logging.

Package.swift
epau, line 1: Idk if you meant to commit this or not. Fine with it for now for testing, but lets not forget to revert it when merging to main

Sources/Core/AWSClientRuntime/EventStream/AWSMessageDecoder.swift
```diff
@@ -0,0 +1,96 @@
...
+//
+
+import AwsCommonRuntimeKit
+import ClientRuntime
+
+extension AWSEventStream {
+
+    /// AWS implementation of `MessageDecoder` for decoding event stream messages
+    /// Note: This is class because struct does not allow using `self` in closure
+    ///      and we need to use `self` to access `messageBuffer` per CRT API.
```
epau, line 15: This is fine, but if we did want this to be a struct, we can create a box to store the mutable properties.
Additionally, this would allow us to convert `decoder` to a `let`, since we won't need to use `self` in the closures.

```swift
struct AWSMessageDecoder: MessageDecoder {
    private class Store {
        var messageBuffer: [EventStream.Mesage] = []
        // etc...
    }
    
    private let store: Store
    
    public init() {
        let store = Store()
        let logger = SwiftLogger(label: "AWSMessageDecoder")
        
        self.decoder = EventStreamMessageDecoder(
            onPayloadSegment: { payload, finalSegment in 
                logger.debug("...")
                store.decodedPayload.append(payload)
            },
            // etc...
         )
         
         self.store = store
         self.logger = logger
    }
}
```

Sources/Core/AWSClientRuntime/EventStream/AWSMessageDecoder.swift
```diff
@@ -0,0 +1,96 @@
...
+                    self.decodededHeaders = []
+                },
+                onError: { code, message in
+                    self.logger.debug("onError: \(code): \(message)")
+                    self.error = AWSEventStreamError.decoding("\(code): \(message)")
+                }
+            )
+        }
+
+        /// Feeds data into the decoder, which may or may not result in a message.
```
epau, line 60: Nit: It'd be nice to provide more details here such as cases where this will result in a message and cases where it will not result in a message.

Sources/Core/AWSClientRuntime/EventStream/AWSMessageDecoder.swift
```diff
@@ -0,0 +1,96 @@
...
+        /// - Parameter data: The data to feed into the decoder.
+        func feed(data: Data) throws {
+            try throwIfErrorOccurred()
```
epau, line 63: Can we add a comment somewhere, perhaps at the top of the class, that explains why we need to call `try throwIfErrorOccurred()` at the beginning of all these functions? This will be helpful for our future selves when we come back to this code down the road.

Sources/Core/AWSClientRuntime/EventStream/AWSMessageDecoder.swift
```diff
@@ -0,0 +1,96 @@
...
+        func endOfStream() throws {
+            try throwIfErrorOccurred()
+
+            guard decodedPayload.isEmpty && decodededHeaders.isEmpty else {
+                throw AWSEventStreamError.decoding("Stream ended before message was complete")
+            }
+        }
+
+        /// Returns the next message in the decoder's buffer
+        /// and removes it from the buffer.
```
epau, line 78:
```suggestion
        /// Returns the next message in the decoder's buffer
        /// and removes it from the buffer.
        /// If the buffer is empty then this returns `nil`.
```

Sources/Core/AWSClientRuntime/EventStream/AWSMessageDecoder.swift
```diff
@@ -0,0 +1,96 @@
...
+        func message() throws -> EventStream.Message? {
+            try throwIfErrorOccurred()
+
+            guard !messageBuffer.isEmpty else {
+                return nil
+            }
+
+            let message = messageBuffer.removeFirst()
+            return message
```
epau, line 87: Nit: 
```suggestion
            return messageBuffer.removeFirst()
```

Sources/Core/AWSClientRuntime/EventStream/AWSMessageDecoder.swift
```diff
@@ -0,0 +1,96 @@
...
+        }
+
+        private func throwIfErrorOccurred() throws {
+            if let error = error {
+                throw error
+            }
```
epau, line 93: Can we add a comment here explaining why we need to use this pattern? This will be helpful for our future selves when we come back to this code down the road.

Sources/Core/AWSClientRuntime/EventStream/AWSMessageDecoderStream.swift
```diff
@@ -0,0 +1,67 @@
...
+                self.stream = stream
+                self.messageDecoder = messageDecoder
+                self.responseDecoder = responseDecoder
+            }
+
+            mutating func next() async throws -> Element? {
+                // if we have a message in the decoder buffer, return it
+                if let message = try messageDecoder.message() {
+                    let event = try Element(message: message, decoder: responseDecoder)
+                    return event
```
epau, line 40: Nit: 
```suggestion
                    return try Element(message: message, decoder: responseDecoder)
```

Sources/Core/AWSClientRuntime/EventStream/AWSMessageDecoderStream.swift
```diff
@@ -0,0 +1,67 @@
...
+                // read until the end of the stream
+                while let data = try stream.read(upToCount: Int.max) {
+                    // feed the data to the decoder
+                    // this may result in a message being returned
+                    try messageDecoder.feed(data: data)
+
+                    // if we have a message in the decoder buffer, return it
+                    if let message = try messageDecoder.message() {
+                        let event = try Element(message: message, decoder: responseDecoder)
+                        return event
```
epau, line 52: Nit: 
```suggestion
                        return  try Element(message: message, decoder: responseDecoder)
```

Sources/Core/AWSClientRuntime/EventStream/AWSMessageDecoderStream.swift
```diff
@@ -0,0 +1,67 @@
...
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import ClientRuntime
+
+extension AWSEventStream {
+
+    /// Stream adapter that decodes input data into `EventStream.Message` objects.
+    struct AWSMessageDecoderStream<Element: MessageUnmarshallable>: MessageDecoderStream {
```
epau, line 13: Would `Event` be a better name here? 

```suggestion
    struct AWSMessageDecoderStream<Event: MessageUnmarshallable>: MessageDecoderStream {
```

Sources/Core/AWSClientRuntime/EventStream/AWSMessageEncoder.swift
```diff
@@ -0,0 +1,23 @@
...
+
+    /// Encodes a `Message` into a `Data` object
+    /// to be sent over the wire.
+    public struct AWSMessageEncoder: MessageEncoder {
+        init() {}
+
+        /// Encodes a `Message` into a `Data` object
+        public func encode(message: EventStream.Message) throws -> Data {
+            let crtMessage = message.toCRTMessage()
+            return try crtMessage.getEncoded()
```
epau, line 20: Ideally this should be injected in to remove the tight coupling to CRT. This would allow us to easily use different logic in the future for encoding (not CRT) and it would allow us to easily mock this if we wanted to. 

I'm also fine with this being out of scope, but still wanted to mention.

Sources/Core/AWSClientRuntime/EventStream/AWSMessageEncoderStream.swift
```diff
@@ -0,0 +1,81 @@
...
+                 requestEncoder: RequestEncoder,
+                 messageSinger: MessageSigner) {
+                self.stream = stream
+                self.messageEncoder = messageEncoder
+                self.messageSinger = messageSinger
+                self.requestEncoder = requestEncoder
+            }
+
+            mutating public func next() async throws -> Data? {
+                var iterator = stream.makeAsyncIterator()
```
epau, line 47: Should we be making this underlying iterator every time `next` is called or does it make more sense to only create this once per instance of this AsyncIterator?

Sources/Core/AWSClientRuntime/EventStream/AWSMessageEncoderStream.swift
```diff
@@ -0,0 +1,81 @@
...
+                guard let event = try await iterator.next() else {
+                    // There are no more messages in the base stream
+                    // if we have not sent the last message, send it now
+                    if !lastMessageSent {
```
epau, line 51: nit:
```suggestion
                    guard lastMessageSent {
```

Sources/Core/AWSClientRuntime/EventStream/AWSMessageEncoderStream.swift
```diff
@@ -0,0 +1,81 @@
...
+
+                // marshall event to message
+                let message = try event.marshall(encoder: requestEncoder)
+
+                // sign the message
+                let signedMessage = try await messageSinger.sign(message: message)
+
+                // encode again the signed message
+                let data = try messageEncoder.encode(message: signedMessage)
+                return data
```
epau, line 70: nit:
```suggestion
                return try messageEncoder.encode(message: signedMessage)
```

Sources/Core/AWSClientRuntime/EventStream/AWSMessageSigner.swift
```diff
@@ -0,0 +1,57 @@
...
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import ClientRuntime
+
+extension AWSEventStream {
+    /// Signs a `Message` using the AWS SigV4 signing algorithm
+    public class AWSMessageSigner: MessageSigner {
+        let context: HttpContext
```
epau, line 13: I don't love the needing this dependency, can we abstract this away perhaps in a closure?

```swift
let signingConfig: () async throws -> AWSSigningConfig
```

epau: Additionally, can we just inject `previousSignature` as well?

Sources/Core/AWSClientRuntime/EventStream/Header+AWS.swift
```diff
@@ -0,0 +1,82 @@
...
+    }
+}
+
+extension Array where Element == EventStreamHeader {
+
+    /// Converts CRT EventStreamHeader array to ClientRuntime EventStream.Header array
+    func toHeaders() -> [EventStream.Header] {
+        self.map {
+            $0.toHeader()
+        }
```
epau, line 80: its a real shame swift doesn't support instance methods as keypaths. Would love to write this as `self.map(\.toHeader)`

Sources/Core/AWSClientRuntime/EventStream/Message+AWS.swift
```diff
@@ -0,0 +1,145 @@
...
+            public let errorCode: String
+            public let message: String?
+        }
+    }
+}
+
+extension EventStream.Message {
+    /// Parses the protocol level headers into a `MessageType`
+    func type() throws -> EventStream.MessageType {
+        let headersByName = Dictionary(grouping: headers, by: { $0.name })
```
epau, line 80: Nit: 
```suggestion
        let headersByName = Dictionary(grouping: headers, by: \.name)
```

Sources/Core/AWSClientRuntime/EventStream/AWSMessageDecoder.swift
ganeshnj: added comment on the method itself.

Sources/Core/AWSClientRuntime/EventStream/AWSMessageDecoderStream.swift
ganeshnj: similar to above.

Sources/Core/AWSClientRuntime/EventStream/AWSMessageDecoderStream.swift
ganeshnj: similar to above

Sources/Core/AWSClientRuntime/EventStream/AWSMessageDecoderStream.swift
ganeshnj: that's fair, when you use Element, you don't have to mapping between the AsyncSequence and other types but that's fine.

Sources/Core/AWSClientRuntime/EventStream/AWSMessageEncoder.swift
ganeshnj: we always swap the implementation without changing the current interfaces. This is implementation details here.

Sources/Core/AWSClientRuntime/EventStream/AWSMessageEncoderStream.swift
ganeshnj: not sure we will gain anything here by making per instance. This is basically like doing a for-loop. 

I generally not prefer using an instance index (ie i). If there is strong case which I'm not aware we can consider for sure.

Sources/Core/AWSClientRuntime/EventStream/AWSMessageEncoderStream.swift
ganeshnj: sure.

Sources/Core/AWSClientRuntime/EventStream/AWSMessageSigner.swift
ganeshnj: I was thinking on this in the shower, we want to delay the signing until the request signature is calculated.

Hence, we need the HttpContext as injected dependency than plan signature. if we pass plain string, it must be computed at the construction time which is not going to work here. Rather it must be calculated when first message is queried.

Sources/Core/AWSClientRuntime/PrimitiveExtensions/String+Extension.swift
```diff
@@ -28,3 +30,15 @@ extension String {
...
+
+extension String {
+    public var hexaData: Data { .init(hexa) }
+    private var hexa: UnfoldSequence<UInt8, Index> {
+        sequence(state: startIndex) { startIndex in
+            guard startIndex < self.endIndex else { return nil }
+            let endIndex = self.index(startIndex, offsetBy: 2, limitedBy: self.endIndex) ?? self.endIndex
+            defer { startIndex = endIndex }
+            return UInt8(self[startIndex..<endIndex], radix: 16)
+        }
```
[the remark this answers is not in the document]
ganeshnj: done

Sources/Core/AWSClientRuntime/EventStream/AWSMessageDecoder.swift
ganeshnj: for now leaving it as it is, because we don't anything wrapping the implementation inside the class.

epau: The one nice thing we get is that `decoder` no longer needs to be a variable and no longer needs to be an optional

Sources/Core/AWSClientRuntime/EventStream/AWSMessageSigner.swift
```diff
@@ -10,22 +10,36 @@ import ClientRuntime
...
-        private var previousSignature: String
+        private var _previousSignature: String?
 
-        public init(context: HttpContext, encoder: MessageEncoder) {
-            self.context = context
-            self.encoder = encoder
+        /// Returns the previous signature used to sign a message
+        /// If no previous signature is available, then the request signature returned
+        /// which acts as previous signature for the first message
+        var previousSignature: String {
```
epau, line 22: Is this just a lazy var? 
`lazy var previousSignature: String = requestSignature()`
This delays computation until the property is first accessed, and it also allows for the value to be explicitly set. 

Its a bit simpler than having the duplicate properties.

epau (approved):
Looks great, just one last minor/nit comment.

Sources/Core/AWSClientRuntime/EventStream/AWSMessageSigner.swift
ganeshnj: could have been lazy if we need it only get, but we set also. hence need a backing field.

Issue #338, opened 2021-08-20
Retry Behavior 2.0


kneekey23:
## Description

Reference from Kotlin: https://github.com/awslabs/aws-sdk-kotlin/issues/199

ganeshnj:
SEP: --/seps/accepted/shared/new-retries.md

PR #921, opened 2023-03-17
feat!: Add error code to AWS errors

jbelkins:
## Issue \#
#338

(Note that this PR merges to the retry feature branch)

## Description of changes
- Adds en error code property to the `AWSHttpServiceError` protocol
- Implements an error code property on generated AWS errors and on the unknown AWS error.

This PR is a breaking change because it adds a field to a public protocol.


Sources/Core/AWSClientRuntime/Errors/AWSHttpServiceError.swift
```diff
@@ -9,4 +9,20 @@ public let X_AMZN_REQUEST_ID_HEADER = "X-Amzn-RequestId"
 
 public protocol AWSHttpServiceError: HttpServiceError {
     var _requestID: String? { get set }
+
+    /// For an error that is identifiable to a Smithy error type, the error code will be the
+    /// non-namespaced name of the error shape.  For unidentified errors, the error code
+    /// may be a Smithy type not handled by this SDK, or may contain other identifying
+    /// information.
```
epau, line 16: imo we should avoid using terms like `smithy` in our code base to future proof against smithy being renamed or replaced.

```suggestion
    /// The non-namespaced name of the error. This value is provided by the operation response and is used to map the response into a specific error. If no error model exists with this name, then an `Unknown` error is created.
```

epau (approved):

PR #935, opened 2023-03-29
feat! Full support for file based configurations

epau:
This PR adds full support for file based configurations via [the config file and the shared credentials file](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-profiles.html). 

The configuration files will be loaded when a service client configuration is created and only at that time.
Previously, the configuration files were loaded every time  a value was read from them. This could be many times throughout a service clients lifetime, hell it could've even been many times during a single operation call. 

In order to fully support this behavior, some additional changes needed to be made:

## Streamlines service client configuration initialization

Previously, there many initializers defined for creating client configuration structs. This PR replaces them with a single initializer that provides default values for every parameter. 

## Makes each credential provider factory method consistent.

This PR adds consistent names for all the structures passed as inputs to the static functions that create credentials providers.

## Adds tests for all credentials providers

Previously there were no tests. Now every credentials provider static function is tested.

## Fixes resolving a region via a profile

Previously, the `ProfileRegionProvider` would only get a region if it was defined in a configuration file located at `~/.aws/config`. This PR fixes this behavior, where now it will correctly get the region from the configuration or shared credentials file at the location specified by the customer (via the proper environment variables).
 

## Issue \#
https://github.com/awslabs/aws-sdk-swift/issues/345
https://github.com/awslabs/aws-sdk-swift/issues/936

## Conventional Commits


Sources/Core/AWSClientRuntime/FileBasedConfiguration/Store/FileBasedConfigurationStore.swift
```diff
@@ -0,0 +1,46 @@
...
+            configPath: configFilePath ?? defaultConfigFilePath,
+            credentialPath: credentialsFilePath ?? defaultCredentialsFilePath
+        )
+        return try await cache.output(for: sources)
+    }
+}
+
+@_spi(Internal)
+extension FileBasedConfigurationStore: FileBasedConfigurationProviding {
+    public func fileBasedConfiguration(configFilePath: String?, credentialsFilePath: String?) async throws -> FileBasedConfigurationSectionProviding? {
```
epau, line 43: Update return type to `FileBasedConfiguration`

Sources/Core/AWSClientRuntime/Signing/AWSSigningConfig.swift
```diff
@@ -58,8 +59,8 @@ extension AWSSigningConfig {
             service: service,
             region: region,
             date: date,
-            credentials: try credentials?.toCRTType(),
-            credentialsProvider: credentialsProvider?.crtCredentialsProvider,
+            credentials: try credentials.map { try CRTCredentials(credentials: $0) },
+            credentialsProvider: nil, //credentialsProvider?.crtCredentialsProvider,
```
epau, line 63: update this

Sources/Core/AWSClientRuntime/Auth/CredentialsProvider/CredentialsProvider.swift
```diff
@@ -0,0 +1,87 @@
...
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import ClientRuntime
+
+public typealias AWSCredentialsProvider = CredentialsProvider
+
+/// An object responsible for providing credentials for an AWS service
+/// Direct initailization is not supported.
```
epau, line 13: typo

jbelkins (changes requested):

Sources/Core/AWSClientRuntime/Auth/CredentialsProvider/CredentialsProvider.swift
```diff
@@ -0,0 +1,87 @@
...
+/// Create instances using its static functions.
+public struct CredentialsProvider {
+    @_spi(Internal)
```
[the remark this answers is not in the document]
epau: That's a good question and something I considered when doing this. I think currently it's unclear how we want to organize this. I'd like to defer that decision to the next time we want use `@_spi()` for a feature. At that time we can play around a bit with this and make a decision that makes most sense.

jbelkins (approved):

PR #957, opened 2023-04-13
fix!: Exposes a sync init for client configs

epau:
This PR enables creating a client and client configuration synchronously. 
- exposes sync inits
- cleans up initializers to reduce duplicative logic
- moves credentials providers to types to better separate concerns and exposes a simpler API
- removes caching of file based config as it is no longer needed. a file based config will be created once during client configuration initialization. 

## Issue \#

https://github.com/awslabs/aws-sdk-swift/issues/958

## Conventional Commits


codegen/smithy-aws-swift-codegen/src/main/kotlin/software/amazon/smithy/aws/swift/codegen/AWSServiceConfig.kt
```diff
@@ -138,20 +139,142 @@ class AWSServiceConfig(writer: SwiftWriter, val ctx: ProtocolGenerator.Generatio
...
+
+        // Create our file based config
+        writer.write("let \$L = try await CRTFileBasedConfiguration.makeAsync()", FILE_BASED_CONFIG_LOCAL_NAME)
         writer.write("")
 
-        // partitionID computed var
-        writer.openBlock("public var partitionID: String? {", "}") {
-            writer.write("return \"\$L - \\(region ?? \"\")\"", serviceName)
+        // Resolve the region resolver
+        writer.write("let resolvedRegionResolver = try \$L ?? DefaultRegionResolver { _, _ in \$L }", REGION_RESOLVER, FILE_BASED_CONFIG_LOCAL_NAME)
```
jbelkins, line 161: So instead of calling the async region resolver here during initialization, we pass in the block that resolves region asynchronously & call it at some later time when async is not inconvenient?

epau: No, we still resolve the region (which is an async call) in this init. This happens just below this on line 170. 

Here we are creating the region resolver, which is either the region resolver provided by the customer or if that is `nil` then its the `DefaultRegionResolver`. The `DefaultRegionResolver` takes a block as an input. This block returns the file based config that should be used by the region resolver.

jbelkins (approved):

PR #972, opened 2023-04-27
fix: Fix streaming for AWS Kinesis

jbelkins:
## Issue \#
https://github.com/awslabs/aws-sdk-swift/issues/973

## Description of changes
- `awsJson1_0` and `awsJson1_1` now generate marshallers/unmarshallers that don't "fast-fail" with `fatalError()`
- Added `AWSKinesisIntegrationTests` to ensure that Kinesis (and `awsJson`-based streaming more generally) work as intended

Also: Fixes to `Package.swift` generation & usage in the "Codegen, Build & Test" workflow

### DIFF
--- a/IntegrationTests/Services/AWSKinesisIntegrationTests/KinesisTests.swift
+++ b/IntegrationTests/Services/AWSKinesisIntegrationTests/KinesisTests.swift
@@ -1,0 +1,104 @@
+//
+// Copyright Amazon.com Inc. or its affiliates.
+// All Rights Reserved.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import XCTest
+import AWSKinesis
+import ClientRuntime
+
+class KinesisTests: XCTestCase {
+
+    func test_kinesisIntegrationTest() async throws {
+        // Performs a test against "live" AWS Kinesis to ensure that the `subscribeToShard` event streaming operation
+        // operates as expected.  Test should take ~20-30 seconds to complete.
+        //
+        // Streaming is over HTTP/1.1.  Streaming is unidirectional (in the response.)
+        // AWS Kinesis uses awsJson1_1 protocol.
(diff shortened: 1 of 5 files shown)


IntegrationTests/Services/AWSKinesisIntegrationTests/KinesisTests.swift
```diff
@@ -0,0 +1,104 @@
...
+            // Monitor the shard event stream
+            for try await event in output.eventStream! {
+                switch event {
+                case .subscribetoshardevent(let event):
+                    event.records?.forEach { record in
+                        let recordString = String(data: record.data ?? Data(), encoding: .utf8)
+                        recordStrings.removeAll { recordString == $0 }
+                    }
+                case .sdkUnknown(let message):
+                    print("Unknown event: \(message)")
```
jbelkins, line 78: The first event to come down the stream is of type `initial-response`:

```
2023-04-28T16:29:53-0500 info AWSMessageDecoder : [Logging] onHeaderReceived: :event-type: string(value: "initial-response")
```

This event is caught & printed out here as unknown.

```
Unknown event: error processing event stream, unrecognized event: initial-response
```

This is followed by an event of type `SubscribeToShardEvent` which is defined in the operation.

jbelkins: This is actually a feature defined in Smithy:
https://smithy.io/2.0/spec/streaming.html?highlight=event%20stream#initial-messages

We need to determine if initial messages should be handled by the SDK versus passed down the event stream to the consumer.

jbelkins: GH issue for tracking handling of initial events: https://github.com/awslabs/aws-sdk-swift/issues/976

PR #1000, opened 2023-05-15
chore: Improve the iOS CI job

jbelkins:
## Description of changes
Replaces the iOS Github CI job with a new job having the following improvements:
- Does not use the CRT Builder script.  Provides its own lightweight scripting for project setup.
- No dependency on access to any AWS account or any special Github credentials.
- No need to maintain any external build container or other resource.
- Builds protocol tests & unit tests together, reducing build time by de-duplicating several build steps.
- Uses `xcodebuild` to build & test the SDK using the iOS Simulator.
- `xcbeautify` is used to clean up & reduce the volume of Swift logs, and to reduce build time.

Once this job has had an opportunity to run for a while and proves reliable & efficient, I plan to adapt it for use with other CI jobs as well.


.github/workflows/continuous-integration.yml
```diff
@@ -24,33 +24,65 @@ jobs:
...
+          restore-keys: |
+            1-${{ runner.os }}-swift5.8.1-spm-${{ hashFiles('Package.swift', 'AWSSDKSwiftCLI/Package.swift') }}
+            1-${{ runner.os }}-swift5.8.1-spm-
+      - name: Setup Java
+        uses: actions/setup-java@v3
+        with:
+          distribution: corretto
+          java-version: 17
+      - name: Setup xcbeautify
+        run: HOMEBREW_NO_AUTO_UPDATE=1 brew install xcbeautify
```
waahm7, line 70: Why set `HOMEBREW_NO_AUTO_UPDATE`? Each CI run will be re-installing xcbeautify.

waahm7 (approved):

.github/workflows/continuous-integration.yml
jbelkins: Homebrew will automatically update whenever it's invoked, unless this var is set.

The Github CI machines are often pretty far behind on the Homebrew repo, so updating homebrew takes a fair amount of time.

`xcbeautify` itself doesn't get updated very often, and I'm OK with being behind the latest version.

PR #1075, opened 2023-08-10
fix: Switch CI log processor from xcbeautify to xcpretty

jbelkins:
## Description of changes
Recently CI builds started failing on macOS 11 CI jobs because `xcbeautify` (installed via Homebrew) was updated to not support macOS 11 & Xcode 13.  `xcbeautify` is not published to Homebrew with versioning, so it is difficult to specify an exact `xcbeautify` version to be installed.

The solution is to use `xcpretty` instead of `xcbeautify` to format logs.  `xcpretty` is slower and less recently updated, but comes pre-installed on every Github mac runner.

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -83,14 +83,14 @@ jobs:
           java-version: 17
       - name: Setup xcbeautify
         run: |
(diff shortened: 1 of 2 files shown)


dayaffe (approved):
LGTM 🚀

Issue #1073, opened 2023-08-09
SDK Process Credential Provider


jbelkins:
### Describe the feature

"Process credentials" are described in a SEP: `--/seps/accepted/shared/extensible-creds.md`

CRT is adding a Process Credential Provider in this ticket: 
https://github.com/awslabs/aws-crt-swift/issues/177

CRT is also adding its process credential provider to its default credential provider chain.

To use the process credential provider outside of the default chain, SDK needs to add access to the CRT process credential provider via a wrapper that conforms to `AWSClientRuntime.CredentialsProviding`.

### Use Case

To allow for use of process credentials outside the default chain.

### Proposed Solution

Create a wrapper for CRT's process credential provider that is located in the SDK.

### Acknowledgements

- [ ] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

jbelkins:
Released in AWS SDK for Swift 0.24.0

PR #1076, opened 2023-08-11
feat: add process credential provider

dayaffe:
"Process credentials" are described in a SEP: --/seps/accepted/shared/extensible-creds.md

Since CRT is adding a Process Credential Provider (see ticket: https://github.com/awslabs/aws-crt-swift/issues/177) we need to create a wrapper for CRT's process credential provider that is located in the SDK.

## Issue \#
https://github.com/awslabs/aws-sdk-swift/issues/1073

## Description of changes
Create a wrapper class ProcessCredentialsProvider with companion unit tests

## Conventional Commits


dayaffe:
Status Update: This PR is on-hold until the next release of CRT and companion version bump in Swift SDK

Tests/Core/AWSClientRuntimeTests/Auth/CredentialsProvidersTests/ProcessCredentialsProviderTests.swift
```diff
@@ -0,0 +1,44 @@
...
+        let credentialsPath = Bundle.module.path(forResource: "credentials", ofType: nil)!
+
+    func testGetCredentialsWithDefaultProfile() async {
+        let subject = try! ProcessCredentialsProvider(
+            configFilePath: configPath,
+            credentialsFilePath: credentialsPath
+        )
+        let credentials = try! await subject.getCredentials()
+
+        XCTAssertNotNil(credentials)
```
jbelkins, line 25: swift nit: This `XCTAssertNotNil` check is not needed because: 
- `let credentials = try! await subject.getCredentials()` will crash if the assignment result is `nil`
- the checks below will necessarily fail if `credentials` is nil

jbelkins (changes requested):
A nit & a little Swift background

Sources/Core/AWSClientRuntime/Auth/CredentialsProviders/ProcessCredentialsProvider.swift
```diff
@@ -0,0 +1,50 @@
...
+/// }
+/// Version here identifies the command output format version.
+public struct ProcessCredentialsProvider: CredentialsSourcedByCRT {
+    let crtCredentialsProvider: CRTCredentialsProvider
+
+    /// Creates a credentials provider that gets credentials from running a command or process.
+    ///
+    /// - Parameters:
+    ///   - fileBasedConfiguration: The file based configuration to read the configuration from.
+    ///   - profileFileNameOverride: (Optional) Override of what profile to use to source credentials from ('default' by default)
```
sichanyoo, line 35: Comment here doesn't match parameters to constructor.

sichanyoo (changes requested):
Nit comment for initializer.

Sources/Core/AWSClientRuntime/Auth/CredentialsProviders/ProcessCredentialsProvider.swift
dayaffe: Oh nice catch!

Tests/Core/AWSClientRuntimeTests/Auth/CredentialsProvidersTests/ProcessCredentialsProviderTests.swift
```diff
@@ -0,0 +1,46 @@
...
+        XCTAssertEqual("SessionToken123", credentials.sessionToken)
+    }
+
+    func testGetCredentialsWithNamedProfileFromConfigFile() async {
+        let subject = try! ProcessCredentialsProvider(
+            profileName: "credentials-process-config-tests-profile",
+            configFilePath: configPath,
+            credentialsFilePath: credentialsPath
+        )
+        let credentials = try! await subject.getCredentials()
```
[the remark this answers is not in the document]
dayaffe: I will review the other credential provider tests and make the same changes so we're consistent

sichanyoo (approved):

IntegrationTests/CredentialsProviders/ProcessCredentialsProviderIntegrationTests/ProcessCredentialsProviderTests.swift
```diff
@@ -0,0 +1,37 @@
...
+        let testConfig = try await S3Client.S3ClientConfiguration()
+        testConfig.credentialsProvider = processCredentialsProvider
+
+        // Initialize our S3 client with the specified configuration
+        client = S3Client(config: testConfig)
+    }
+
+    // This test calls listBuckets() and forces S3Client to use ProcessCredentialsProvider
+    func test_listBuckets() async throws {
+        _ = try await client.listBuckets(input: ListBucketsInput())
```
jbelkins, line 35: Not sure how this test is going to pass, the credentials will be loaded from the `echo` command in the `credentials` config file below, which aren't real AWS credentials

Package.swift
```diff
@@ -116,6 +116,18 @@ func addIntegrationTestTarget(_ name: String) {
     ]
 }
 
+func addCustomIntegrationTestTarget(_ folder: String, _ name: String, dependsOn dependency: String) {
```
jbelkins, line 119: Not sure I want to add this complexity to the package manifest.  Maybe better to locate these tests with S3 since that is the service being authenticated in the test?

In any case, this file is overwritten and can't be hand-modified.  See the `AWSSDKSwiftCLI` subproject for where this file is generated:
https://github.com/awslabs/aws-sdk-swift/tree/main/AWSSDKSwiftCLI

jbelkins (changes requested):

Package.swift
dayaffe: Yup I was just going through that. The issue with not modifying the IntegrationTestTarget function is that right now its setup so that the name of the folder corresponds to the test dependency. So AWSS3IntegrationTests depends on AWSS3. However ProcessCredentialsIntegrationTests depends on AWSS3 so that setup will not work even if its put under Services.

IntegrationTests/CredentialsProviders/ProcessCredentialsProviderIntegrationTests/ProcessCredentialsProviderTests.swift
dayaffe: So we have a few options here. Tests aren't run automatically so we dont have a dedicated AWS Account that these are running against. So all the integration tests are running against the user's personal ~/.aws/config and ~/.aws/credentials from my understanding. User who runs this integ test likely does not have their credentials file setup to be using ProcessCredentialsProviders so either they would have to modify their own or use this one which I provided for convenience. Does that make sense?

Package.swift
dayaffe: I was going to revert my changes to Package.swift and discuss options with you. One workaround is we could put this test file inside of AWSS3

dayaffe: Since different credential providers may be tested against different services depending on use-case perhaps it does make sense to do that. I have changed it in my latest commit.

jbelkins (approved):

Issue #236, opened 2021-07-13
SSO Credential Provider


wooj2:
/seps/accepted/shared/sso-provider/
/seps/accepted/shared/sso-login-flow/

*internal id*: 174915617

benkehoe:
What's the meaning of the paths in the comment above? Can I find them somewhere?

richardhboyd:
You can't find them. They're a mechanism we use to align feedback we get from multiple places.

benkehoe:
So how am I supposed to have any idea what this or any other ticket on this repo means?

ganeshnj:
CRT bindings won't be available mid feb, we have to re-prioritize it.

smivij:
Changing priority to 2 as per discussion with Rust and Kotlin

smivij:
CRT changes in [SSO Credential Provider#95](https://github.com/awslabs/aws-crt-swift/issues/95)

jbelkins:
Released in AWS SDK for Swift 0.24.0

PR #1084, opened 2023-08-15
feat: Add SSO credentials provider

sichanyoo:
Adds wrapper for [CRT's SSO credentials provider implementation](https://github.com/awslabs/aws-crt-swift/pull/182/files).

## Issue \#

https://github.com/awslabs/aws-sdk-swift/issues/236


Tests/Core/AWSClientRuntimeTests/Auth/CredentialsProvidersTests/SSOCredentialsProviderTests.swift
```diff
@@ -0,0 +1,43 @@
...
+    func testCreateCredentialsProviderSSOLegacyProfile() async throws {
+        do {
+            let provider = try SSOCredentialsProvider(
+                profileName: "user",
+                configFilePath: configPath,
+                credentialsFilePath: credentialsPath)
+            XCTAssertNotNil(provider)
+            // get credentials will fail in CI due to expired token, so do not assert on credentials.
+            _ = try? await provider.getCredentials()
+        }
```
dayaffe, line 27: I believe you can remove the do {} block, remove the XCTAssertNotNil and remove the ? from try?. 
- do {} block isnt being caught so doesn't seem necessary
- XCTAssertNotNil isn't necessary since if it results in an exception test will fail automatically
- try? is not necessary since we're not using the output and if its null we'll get an exception causing the test to fail

Tests/Core/AWSClientRuntimeTests/Auth/CredentialsProvidersTests/SSOCredentialsProviderTests.swift
```diff
@@ -0,0 +1,43 @@
...
+    func testCreateCredentialsProviderSSOTokenProviderProfile() async throws {
+        do {
+            let provider = try SSOCredentialsProvider(
+                profileName: "dev",
+                configFilePath: configPath,
+                credentialsFilePath: credentialsPath)
+            XCTAssertNotNil(provider)
+            // get credentials will fail in CI due to expired token, so do not assert on credentials.
+            _ = try? await provider.getCredentials()
+        }
```
dayaffe, line 39: apply same comments as above

dayaffe (changes requested):
Everything else looks good! Just mentioning some test comments that Josh gave me on my ProcessCredentialsProvider CR

Tests/Core/AWSClientRuntimeTests/Auth/CredentialsProvidersTests/SSOCredentialsProviderTests.swift
```diff
@@ -0,0 +1,37 @@
...
+class SSOCredentialsProviderTests: XCTestCase {
+    let configPath = Bundle.module.path(forResource: "sso_tests", ofType: nil)!
+    let credentialsPath = Bundle.module.path(forResource: "credentials", ofType: nil)!
+    
+    func testCreateCredentialsProviderSSOLegacyProfile() async throws {
+        let provider = try SSOCredentialsProvider(
+            profileName: "user",
+            configFilePath: configPath,
+            credentialsFilePath: credentialsPath)
+        XCTAssertNotNil(provider)
```
dayaffe, line 23: nit: line 22 and same in the test below closing ) should be on new line

Just curious, why don't we call provider.getCredentials() anymore? I was thinking the test is

```
let provider = try SSOCredentialsProvider(
            profileName: "user",
            configFilePath: configPath,
            credentialsFilePath: credentialsPath
)
let credentials = try provider.getCredentials()

XCTAssert(credentials.field, expectedValue)
...
```

dayaffe: I think the unit tests here should also 'mock' (swift equivalent) responses from getCredentials() and then make sure that our credentials are loaded based on that response giving us something valid or throwing an error. Right now the scope seems too small to me

sichanyoo: After discussion on slack I have better idea on what to do now - will be making changes and pushing them.

dayaffe: For this provider, since getCredentials() is a method defined by the CRT which makes a service call, we are limited in our ability to sufficiently unit test that logic in the SDK. For test coverage here I am fine with unit tests covering instantiation alone of SSOCredentialsProvider given that we plan to create integration tests in a fast follow up.

dayaffe: Instead of creating provider and checking that its not nil, can either omit that or ideally confirm that all members on the created object are set properly

dayaffe (approved):
LGTM

Sources/Core/AWSClientRuntime/Auth/CredentialsProviders/SSOCredentialsProvider.swift
```diff
@@ -0,0 +1,37 @@
+//
+// Copyright Amazon.com Inc. or its affiliates.
+// All Rights Reserved.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import AwsCommonRuntimeKit
+import ClientRuntime
+import Foundation
```
jbelkins, line 10: nit: I don't think importing Foundation is needed here.  If not, remove the import

Tests/Core/AWSClientRuntimeTests/Auth/CredentialsProvidersTests/SSOCredentialsProviderTests.swift
```diff
@@ -0,0 +1,38 @@
...
+    let configPath = Bundle.module.path(forResource: "sso_tests", ofType: nil)!
+    let credentialsPath = Bundle.module.path(forResource: "credentials", ofType: nil)!
+    
+    func testCreateCredentialsProviderSSOLegacyProfile() async throws {
+        _ = try SSOCredentialsProvider(
+            profileName: "user",
+            configFilePath: configPath,
+            credentialsFilePath: credentialsPath
+        )
+        // SUCCESS: creation didn't throw error
```
jbelkins, line 24: Nothing we can do to verify that the expected profile & config files are accessed?

jbelkins (changes requested):

Tests/Core/AWSClientRuntimeTests/Auth/CredentialsProvidersTests/SSOCredentialsProviderTests.swift
sichanyoo: Found a workaround way to verify the expected profile and config files are accessed by adding a test case that throws exception due to having a nonexistent profile name as one of the inputs to the SSO credentials provider constructor.

jbelkins (approved):

Issue #1081, opened 2023-08-15
SDK ECS Credentials Provider


dayaffe:
### Describe the feature

CRT has provided an ECS Credentials Provider method that we should expose (see [here](https://github.com/awslabs/aws-crt-swift/blob/main/Source/AwsCommonRuntimeKit/auth/credentials/CredentialsProvider.swift#L511)). 

### Use Case

- This provider sources credentials from ECS container metadata.
- ECS creds provider can be used to access creds via either relative uri to a fixed endpoint http://169.254.170.2, or via a full uri specified by environment variables:

### Proposed Solution

Create a new class `ECSCredentialsProvider` and implement a wrapper for the method provided. Some logic may need to be added to the SDK side in order to properly handle relative vs full URI (see [#L511](https://github.com/awslabs/aws-crt-swift/blob/main/Source/AwsCommonRuntimeKit/auth/credentials/CredentialsProvider.swift#L511) and [#L53](https://github.com/awslabs/aws-c-auth/blob/main/source/credentials_provider_default_chain.c#L53))

### Acknowledgements

- [ ] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

PR #1091, opened 2023-08-23
feat: add ecs credentials provider

dayaffe:
Add a wrapper to expose ECS Credentials Provider independent of the default provider chain

Note: If running the integration test prior to merge, please modify the Package.swift file of ECSIntegTestApp to point to this branch `day/ecs-credentials-provider` instead of `main` for `aws-sdk-swift` dependency and run the provided bash script.

## Issue \#
https://github.com/awslabs/aws-sdk-swift/issues/1081

## Description of changes
- Create a wrapper for the CRT bindings that exposes ECS Credentials Provider and properly uses relativeURI with default host or absoluteURI
- Validate URI
- Unit testing for ECS Credentials Provider
- Integration test which   
     - provides a test package (dockerized) to be run inside of an ECS cluster that calls STS with a bash script for deploying to ECR
     - creates an ECS Fargate Arm64 instance and companion resources
     - launches task to run dockerized test package and poll status until it is complete
     - scan latest log stream in related group for keyword emitted by test
     - tear down ECS resources

## Conventional Commits


Sources/Core/AWSClientRuntime/Auth/CredentialsProviders/ECSCredentialsProvider.swift
```diff
@@ -0,0 +1,96 @@
...
+    /// Token is used in auth header but only for absolute uri.
+    /// - Returns: `CredentialsProvider`
+    /// - Throws: CommonRuntimeError.crtError or InitializationError.missingURIs
+    public init(
+        environment: Environment = ProcessEnvironment()
+    ) throws {
+        let relativeURI = environment.environmentVariable(key: "AWS_CONTAINER_CREDENTIALS_RELATIVE_URI")
+        let absoluteURI = environment.environmentVariable(key: "AWS_CONTAINER_CREDENTIALS_FULL_URI")
+
+        guard relativeURI != nil || isValidAbsoluteURI(absoluteURI) else {
```
sichanyoo, line 32: I think this is a bit confusing because to determine when this guard block will be executed, we really gotta think through the conditional because guard executes for false condition, and we're using `||` operator with `!=` on the relativeURI.

Maybe more direct way is to do: 
`if (relativeURI == nil && !isValidAbsoluteURI(absoluteURI)) { \\ throw error }` ?
In this case it's much simpler (imo) to think it through - we only throw error if `relativeURI` is nil & absoluteURI is invalid.

But in general I'm still in process of trying to be friends with guard syntax - so I'm not entirely sure how much of me finding this confusing is due to the difference in my familiarity with if vs. guard. Also, using guard to check for exception case seems more idiomatic approach in Swift since we're literally "guarding" it haha. Let me know what you think.

Sources/Core/AWSClientRuntime/Auth/CredentialsProviders/ECSCredentialsProvider.swift
```diff
@@ -0,0 +1,96 @@
...
+                "Please configure either the relative or absolute URI environment variable!"
+            )
+        }
+
+        let defaultHost = "169.254.170.2"
+        var host = defaultHost
+        var pathAndQuery = relativeURI ?? ""
+
+        if let absoluteString = absoluteURI, relativeURI == nil {
+            if let uriURL = URL(string: absoluteString) {
```
sichanyoo, line 43: I think combining these two using multiple let might make code simpler to read:

`if let absoluteString = absoluteURI, let uriURL = URL(string: absoluteString), relativeURI == nil { \\code }`

sichanyoo (commented):
The code for the wrapper looks good, just left couple comments as suggestions.

Sources/Core/AWSClientRuntime/Auth/CredentialsProviders/ECSCredentialsProvider.swift
dayaffe: I like the idea and in that case I will put `let uriURL = URL(string: absoluteString)` condition last since order here matters and I want to avoid any unnecessary evaluation of absoluteString if relativeURI is not nil.

Sources/Core/AWSClientRuntime/Auth/CredentialsProviders/ECSCredentialsProvider.swift
dayaffe: This is a great comment and in general I agree that yours is more readable at the moment (to both you and me), but I have a strong preference for `guard` here. `if` statements are a general-purpose conditional, so anything could be inside of such a code block. If I were only setting state here I would switch it, but `guard` indicates that this is a required condition for the program to continue. So by using guard I am emphasizing that we cannot continue with execution unless we have a valid relative or absolute URI.

dayaffe: This can be read as "To continue executing (guard) we require either relativeURI is not nil (!=) or (||) absoluteURI is valid, otherwise we'll throw an error"

sichanyoo: > "To continue executing (guard) we require either relativeURI is not nil (!=) or (||) absoluteURI is valid, otherwise we'll throw an error"

Reading it that way does make it much better, and yep, I agree that using `guard` here does put better emphasis on execution condition.

Sources/Core/AWSClientRuntime/Auth/CredentialsProviders/ECSCredentialsProvider.swift
```diff
@@ -0,0 +1,96 @@
...
+enum InitializationError: Error {
+    case missingURIs(String)
+
+    var localizedDescription: String {
+        switch self {
+        case .missingURIs(let message):
+            return message
+        }
+    }
+}
```
waahm7, line 66: SDK has a global error handling and we should probably align the error handling to that.

Sources/Core/AWSClientRuntime/Auth/CredentialsProviders/ECSCredentialsProvider.swift
```diff
@@ -0,0 +1,96 @@
...
+
+private func buildPathAndQuery(from url: URL) -> String {
+    let path = url.path
+    let query = url.query ?? ""
+    let fragment = url.fragment ?? ""
+
+    return path
+        + (query.isEmpty ? "" : "?\(query)")
+        + (fragment.isEmpty ? "" : "#\(fragment)")
+}
```
waahm7, line 76: SDK should already have helper functions to do this. Maybe @jbelkins can guide better on this.

waahm7 (changes requested):
This will run but doesn't fit into overall architecture of the SDK. We should make it consistent with the overall design of the SDK instead of defining private helper functions just for this class.

Sources/Core/AWSClientRuntime/Auth/CredentialsProviders/ECSCredentialsProvider.swift
jbelkins: Usual purpose of guard is to verify initial conditions for the following block of code.  This looks appropriate.

Sources/Core/AWSClientRuntime/Auth/CredentialsProviders/ECSCredentialsProvider.swift
jbelkins: Yes, combining `if let` assignments with commas instead of nesting `if let`s is better.

Sources/Core/AWSClientRuntime/Auth/CredentialsProviders/ECSCredentialsProvider.swift
```diff
@@ -0,0 +1,96 @@
...
+    /// - AWS_CONTAINER_CREDENTIALS_FULL_URI
+    /// - AWS_CONTAINER_AUTHORIZATION_TOKEN
+    ///
+    /// If both relative uri and absolute uri are set, relative uri has higher priority.
+    /// Token is used in auth header but only for absolute uri.
+    /// - Returns: `CredentialsProvider`
+    /// - Throws: CommonRuntimeError.crtError or InitializationError.missingURIs
+    public init(
+        environment: Environment = ProcessEnvironment()
+    ) throws {
```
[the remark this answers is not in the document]
dayaffe: @waahm7 are you saying this is appropriate or suggesting an alternative? The goal here is to be able to pass in a different environment to use during unit testing. What do you suggest?

Sources/Core/AWSClientRuntime/Auth/CredentialsProviders/ECSCredentialsProvider.swift
```diff
@@ -0,0 +1,96 @@
...
+        return false
+    }
+
+    // sanity check host
+    guard url.host != nil else {
+        return false
+    }
+
+    return true
+}
```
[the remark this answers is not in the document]
dayaffe: Will simplify these checks as suggested

Sources/Core/AWSClientRuntime/Auth/CredentialsProviders/ECSCredentialsProvider.swift
dayaffe: Awesome can you point me to where I can read more about how SDKs do error handling?

Sources/Core/AWSClientRuntime/Auth/CredentialsProviders/ECSCredentialsProvider.swift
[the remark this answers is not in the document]
waahm7: We don't necessarily need to pass an environment to use during unit testing. See https://github.com/awslabs/aws-crt-swift/blob/main/Test/AwsCommonRuntimeKitTests/auth/CredentialsProviderTests.swift#L95. What I'm saying is that any property that can be configured via an environment variable should also be configurable via parameters to keep the design consistent. See https://github.com/awslabs/aws-crt-swift/blob/main/Test/AwsCommonRuntimeKitTests/auth/CredentialsProviderTests.swift#L95 for reference. Users should be able to configure ECS using parameter values only.

Sources/Core/AWSClientRuntime/Auth/CredentialsProviders/ECSCredentialsProvider.swift
waahm7: Shared offline

Sources/Core/AWSClientRuntime/Auth/CredentialsProviders/ECSCredentialsProvider.swift
dayaffe: Discussed with @jbelkins offline. He mentioned that the SDK has helpers for URL handling but he doesn't recommend using them outside of constructing URLs for operations. Instead he suggested using URLComponents which is what is being used here.

dayaffe: Looking into if there is a better way to use URLComponents

dayaffe: Discussed further offline and switching to:

```
var components = URLComponents(url: url, resolvingAgainstBaseURL: false)!

components.scheme = nil
components.host = nil

return components.url
```

sichanyoo (approved):
Discussed offline regarding handling of integration test target dependency.
Lgtm!

jbelkins (approved):
There is a lot of heavyweight setup around this test, which admittedly must be there if the test is to be run against live ECS.

But in the long term, I wonder if we should look at "stubbing" ECS (and specific AWS services involved in other, future identity / credential provider tests) so that we can test what we need to in a lighter, quicker, and more portable fashion.

Issue #976, opened 2023-04-29
Determine how to handle initial messages in event streams


jbelkins:
### Describe the feature

Initial messages are defined in the Smithy spec here: https://smithy.io/2.0/spec/streaming.html#initial-messages

Right now, initial messages are streamed to the SDK as `.sdkUnknown` events.  This is kind of contradictory, since initial messages are a documented Smithy feature.

Determine how initial messages should be handled, if different from the _status quo_, and implement.

```[tasklist]
### Tasks
- [ ] https://github.com/awslabs/aws-sdk-swift/issues/1163
- [ ] https://github.com/awslabs/aws-sdk-swift/issues/1164
- [ ] https://github.com/awslabs/aws-sdk-swift/issues/1172
- [ ] https://github.com/awslabs/aws-sdk-swift/issues/1175
```

### Use Case

To correctly handle initial messages and make them easily available to stream consumers.

### Acknowledgements

- [ ] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

PR #1125, opened 2023-09-13
feat: add initial-response support

dayaffe:
Add support for parsing initial-response and making it available to users prior to parsing the Event Stream. 
See related: https://smithy.io/2.0/spec/streaming.html#initial-messages

## Issue \#
#976 

## Description of changes
- Services should send initial-response as the first message in response. `AWSMessageDecoder` will detect if the header is of `:event-type` =  `initial-response` and if so set private variable `self.initialResponse = message.payload` to be parsed based on service-specific expected initial-response fields (codegen).
- `awaitInitialResponse` converts `retrieveInitialResponse` function into an async function. By using this callback method we're able to wait for self.initialMessage to be populated and then return it asynchronously. This method only waits until the first message is processed as if an initial-response is present it must be the very first message.
- `retrieveInitialResponse` immediately calls the completion handler with the message if initialMessage is already processed, otherwise it saves the completion handler to be called later when the message is ready
- Update unit tests to include initial-response message decoding, encoding, and `awaitInitialResponse`

Ex. `initial-response` Usage
```
// Create the subscription stream
let input = SubscribeToShardInput(consumerARN: consumerARN, shardId: shard?.shardId, startingPosition: KinesisClientTypes.StartingPosition(sequenceNumber: shard?.sequenceNumberRange?.startingSequenceNumber, type: .atSequenceNumber))
let output = try await client.subscribeToShard(input: input)

// initial-response: If initialResponseMember is a non-streaming trait defined in the Kinesis model
guard let initialResponse = output.initialResponseMember?.value else {
    throw MyError.initialResponseNil
}

// Monitor the shard event stream
for try await event in output.eventStream! {
    switch event {
    case .subscribetoshardevent(let event):
        event.records?.forEach { record in
            let recordString = String(data: record.data ?? Data(), encoding: .utf8)
            recordStrings.removeAll { recordString == $0 }
        }
    case .sdkUnknown(let message):
        print("Unknown event: \(message)")
    }

    // Once all the events have been received, stop streaming
    if recordStrings.isEmpty { break }
}
```

Note: Will revert `.package(url: smithySwiftURL, .branch("day/initial-messages"))` prior to merge. I only added this to ensure CI passes.

## Conventional Commits


dayaffe:
closing in favor of https://github.com/awslabs/aws-sdk-swift/pull/1165

Issue #1037, opened 2023-06-13
Endpoints 2.0 Refactoring


smivij:

kstich:
The package `smithy-rules-engine` has undergone a major refactoring in an effort to improve validation, separate AWS and non-AWS concerns, add a specification, and better align with Smithy's coding standards. [A code-stable pull request is available here](https://github.com/smithy-lang/smithy/pull/1855). General notes are provided in individual commit messages. The format of the rules documents have _**not**_ changed, meaning a successful migration to the refactored codebase will involve _**no changes**_ to code generated for an SDK client.

The Smithy team plans to release this update on August 15th, 2023. Please reach out to us for any additional information or concerns.

cmoher:
The coordinated release is being pushed to 09/06. This does not change or impact the required changes for the SDKs.

kstich:
The coordinated release of this refactoring has been delayed until 09/20.

jbelkins:
Updating Smithy used by Swift SDK to latest version, in prep for this refactor.

jbelkins:
Moving to Blocked until the refactor ships in the Smithy project.

jbelkins:
Endpoints refactoring shipped in Smithy 1.39.0.  Moving this issue from Blocked to Pending Release.

jbelkins:
Released in AWS SDK for Swift 0.26.0

PR #1126, opened 2023-09-14
chore: Update to Smithy 1.39.0

jbelkins:
## Issue \#
#1037

## Description of changes

Updates the project for breaking changes to the endpoints rule engine in Smithy [1.39.0](https://github.com/smithy-lang/smithy/releases/tag/1.39.0).

- Update the name and/or location of several Smithy classes.
- The Swift SDK codegen phase currently copies `partitions.json` out of the Smithy project for use when code-generating endpoint resolvers.  This is not desirable because the copy in Smithy is infrequently updated.  Eventually, `partitions.json` will be supplied at build time by the SDK release coordination system, but for now, an up-to-date copy is provided in this project, and will be updated when needed.
- `partitions.json` is no longer rendered into every endpoint resolver since it is the same for every service client.  Instead, one common copy is provided in `AWSClientRuntime`.
- An unused `endpoints.json` file is deleted.

Protocol tests:
- It was discovered that Smithy protocol tests for RestJSON were being run twice: once with Smithy tests & again with local tests.  The second execution is omitted.
- There are two new protocol tests that fail.  They are disabled & tracked by separate tickets.


sichanyoo (approved):
lgtm

PR #1129, opened 2023-09-18
fix: Add sso credentials integration test

sichanyoo:
## Issue \#

N/A

## Description of changes

Adds integration test for `SSOCredentialsProvider` that uses `S3::listBuckets`.


IntegrationTests/Services/AWSS3IntegrationTests/SSOCredentialsProviderTests.swift
```diff
@@ -0,0 +1,105 @@
...
+    // Change this region string to the region where AWS SSO instance is in.
+    let region = "us-west-2"
+    
+    private var iamIdentityCenterInstanceArn: String = ""
+    private let permissionSetName = "SSOCredProvIntegTestPermSet"
+    private var permissionSetArn: String = ""
+    private let awsReadOnlyPolicy = "arn:aws:iam::aws:policy/ReadOnlyAccess"
+    
+        
+    override func setUp() async throws {
```
jbelkins, line 45: Not strictly necessary now, and the default implementation in `XCTestCase` does nothing, but it's good to future-proof by adding a call to `super` when overriding setup & teardown methods.

IntegrationTests/Services/AWSS3IntegrationTests/SSOCredentialsProviderTests.swift
```diff
@@ -0,0 +1,105 @@
...
+            permissionSetArn = createPermissionSetOutput.permissionSet!.permissionSetArn!
+            
+            
+            // Attach ReadOnly AWS-managed policy to the created permission set
+            let _ = try await ssoClient.attachManagedPolicyToPermissionSet(input: AttachManagedPolicyToPermissionSetInput(
+                instanceArn: iamIdentityCenterInstanceArn,
+                managedPolicyArn: awsReadOnlyPolicy,
+                permissionSetArn: permissionSetArn))
+        } catch let error as AWSSSOAdmin.ConflictException {
+            if error.message == "PermissionSet with name \(permissionSetName) already exists." {
```
jbelkins, line 85: Assume error messages can change at any time.  Is there a more reliable way to verify this error condition?

IntegrationTests/Services/AWSS3IntegrationTests/SSOCredentialsProviderTests.swift
```diff
@@ -0,0 +1,105 @@
...
+        // Create a S3 client that uses SSO credentials provider
+        try await setUpClient()
+    }
+
+    // The test calls listBuckets() and forces S3Client to use SSOCredentialsProvider
+    func test_listBuckets() async throws {
+        _ = try await client.listBuckets(input: ListBucketsInput())
+    }
+    
+    override func tearDown() async throws {
```
jbelkins, line 59: You can simply omit this override if it does nothing

jbelkins (changes requested):
Nothing huge, just a few small issues to address

IntegrationTests/Services/AWSS3IntegrationTests/SSOCredentialsProviderTests.swift
sichanyoo: Fixed.

IntegrationTests/Services/AWSS3IntegrationTests/SSOCredentialsProviderTests.swift
sichanyoo: Added `super.setUp()` as suggested.

IntegrationTests/Services/AWSS3IntegrationTests/SSOCredentialsProviderTests.swift
sichanyoo: Currently there's no other way to check the specific cause for the error other than checking the message contained int he error. So as a workaround I now look for smaller substring using `contains()` to reduce dependency on the entire error message string staying the same. Now it only checks that `\(permissionSetName) already exists` is part of the error message.

IntegrationTests/Services/AWSS3IntegrationTests/SSOCredentialsProviderTests.swift
```diff
@@ -0,0 +1,105 @@
...
+            let createPermissionSetOutput = try await ssoClient.createPermissionSet(input: CreatePermissionSetInput(
+                description: "Permission set for testing SSO credentials provider.",
+                instanceArn: iamIdentityCenterInstanceArn,
+                name: permissionSetName
+            ))
+            permissionSetArn = createPermissionSetOutput.permissionSet!.permissionSetArn!
+            
+            
+            // Attach ReadOnly AWS-managed policy to the created permission set
+            let _ = try await ssoClient.attachManagedPolicyToPermissionSet(input: AttachManagedPolicyToPermissionSetInput(
```
dayaffe, line 77: same as above

dayaffe (changes requested):
Logic looks good but would like to clean up large helper function

IntegrationTests/Services/AWSS3IntegrationTests/SSOCredentialsProviderTests.swift
```diff
@@ -0,0 +1,102 @@
...
+                instanceArn: iamIdentityCenterInstanceArn,
+                managedPolicyArn: awsReadOnlyPolicy,
+                permissionSetArn: permissionSetArn))
```
dayaffe, line 80: Might be more readable to move contents of the do { ... } to its own helper function

IntegrationTests/Services/AWSS3IntegrationTests/SSOCredentialsProviderTests.swift
```diff
@@ -0,0 +1,105 @@
...
```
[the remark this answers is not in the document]
sichanyoo: Yeah without explicit `let _` that indicates I don't intend to use the result, Xcode yells `Result of call to 'XYZ' is unused`.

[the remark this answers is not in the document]
dayaffe: And what does this method return? Want to make sure it doesnt make sense to sanity check the result == success or something equivalent. If not then I am good with `let _ =`

IntegrationTests/Services/AWSS3IntegrationTests/SSOCredentialsProviderTests.swift
```diff
@@ -0,0 +1,105 @@
...
+    
+    override func tearDown() async throws {
+        // No action required.
+    }
+    
+    /* HELPER METHODS */
+    private func createPermissionSetIfNeeded() async throws {
+        // Get IAM identity center instanceArn
+        let listInstancesOutput = try await ssoClient.listInstances(input: ListInstancesInput())
+        iamIdentityCenterInstanceArn = listInstancesOutput.instances![0].instanceArn!
```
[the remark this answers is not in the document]
sichanyoo: Changed as suggested.

IntegrationTests/Services/AWSS3IntegrationTests/SSOCredentialsProviderTests.swift
```diff
@@ -0,0 +1,105 @@
...
+        
+        do {
+            // Create permission set and save its ARN
+            let createPermissionSetOutput = try await ssoClient.createPermissionSet(input: CreatePermissionSetInput(
+                description: "Permission set for testing SSO credentials provider.",
+                instanceArn: iamIdentityCenterInstanceArn,
+                name: permissionSetName
+            ))
+            permissionSetArn = createPermissionSetOutput.permissionSet!.permissionSetArn!
```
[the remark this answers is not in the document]
sichanyoo: Changed as suggested.

IntegrationTests/Services/AWSS3IntegrationTests/SSOCredentialsProviderTests.swift
sichanyoo: Changed as suggested.

IntegrationTests/Services/AWSS3IntegrationTests/SSOCredentialsProviderTests.swift
```diff
@@ -0,0 +1,116 @@
...
+        _ = try await client.listBuckets(input: ListBucketsInput())
+    }
+    
+    /* HELPER METHODS */
+    private func createPermissionSetIfNeeded() async throws {
+        // Get IAM identity center instanceArn
+        let listInstancesOutput = try await ssoClient.listInstances(input: ListInstancesInput())
+        guard let iamCenterInstances = listInstancesOutput.instances, let iamCenterArn = iamCenterInstances[0].instanceArn else {
+            throw ClientError.dataNotFound("No IAM Identity Center instance found for the account."
+                             + "Ensure that AWS organization is enabled for the account you're using to run the integration test.")
```
dayaffe, line 67: Ideally we should either use multi-line strings """ """ or if the intention is to have this message on a single line we can shorten it to be more concise like "No IAM Identity Center found. Check AWS organization for the test account."

IntegrationTests/Services/AWSS3IntegrationTests/SSOCredentialsProviderTests.swift
```diff
@@ -0,0 +1,116 @@
...
+        guard let permSet = createPermissionSetOutput.permissionSet, let permSetArn = permSet.permissionSetArn else {
+            throw ClientError.dataNotFound("Permission set arn could not be retrieved after creation.")
+        }
+        permissionSetArn = permSetArn
+        
+        // Attach ReadOnly AWS-managed policy to the created permission set
+        _ = try await ssoClient.attachManagedPolicyToPermissionSet(input: AttachManagedPolicyToPermissionSetInput(
+            instanceArn: iamIdentityCenterInstanceArn,
+            managedPolicyArn: awsReadOnlyPolicy,
+            permissionSetArn: permissionSetArn))
```
dayaffe, line 101: I think we can make it even more readable. Can we make 2 helper functions here. Maybe something like

`getPermissionSet()` which does
```
let input = CreatePermissionSetInput(
    description: "Permission set for testing SSO credentials provider.",
     instanceArn: iamIdentityCenterInstanceArn,
     name: permissionSetName
)
  
let output = try await ssoClient.createPermissionSet(input: input)
guard let permSetArn = output.permissionSet?.permissionSetArn else {
    throw ClientError.dataNotFound("Permission set ARN not found after creation.")
}
return permSetArn
```

and  then

`attachManagedPolicy(arn: String)` which does
```
let input = AttachManagedPolicyToPermissionSetInput(
      instanceArn: iamIdentityCenterInstanceArn,
      managedPolicyArn: awsReadOnlyPolicy,
      permissionSetArn: arn
)
_ = try await ssoClient.attachManagedPolicyToPermissionSet(input: input)
```

then `createAndAttachPermissionSet()` instead of `createPermissionSet()` can do 
```
// create permission set and return arn
let permSetArn = try await getPermissionSet()
// attach created policy
try await attachManagedPolicy(arn: permSetArn)
```

IntegrationTests/Services/AWSS3IntegrationTests/SSOCredentialsProviderTests.swift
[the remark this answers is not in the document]
sichanyoo: This method returns a HTTP 200 response with no elements (no headers, nothing in body) if successful. If not successful, it throws an error returned from service.

IntegrationTests/Services/AWSS3IntegrationTests/SSOCredentialsProviderTests.swift
sichanyoo: Changed as suggested.

IntegrationTests/Services/AWSS3IntegrationTests/SSOCredentialsProviderTests.swift
sichanyoo: Refactored to smaller scope for each function as suggested - added three functions.

dayaffe (approved):
LGTM

jbelkins (approved):

Issue #1159, opened 2023-09-28
AWSBedrockRuntime fails to compile


jbelkins:
### Describe the bug

The `AWSBedrockRuntime` service fails to compile.

Compile errors received:
```
/Users/jbelkins/Code/Xcode/aws-sdk-swift/Sources/Services/AWSBedrockRuntime/models/Models.swift:623:39: error: value of type 'ModelStreamErrorException' has no member 'originalMessage'
        if let originalMessage = self.originalMessage {
                                 ~~~~ ^~~~~~~~~~~~~~~
/Users/jbelkins/Code/Xcode/aws-sdk-swift/Sources/Services/AWSBedrockRuntime/models/Models.swift:626:42: error: value of type 'ModelStreamErrorException' has no member 'originalStatusCode'
        if let originalStatusCode = self.originalStatusCode {
                                    ~~~~ ^~~~~~~~~~~~~~~~~~
/Users/jbelkins/Code/Xcode/aws-sdk-swift/Sources/Services/AWSBedrockRuntime/models/Models.swift:882:81: error: cannot find type 'internalServerException' in scope
                    return try decoder.decode(responseBody: message.payload) as internalServerException
                                                                                ^~~~~~~~~~~~~~~~~~~~~~~
/Users/jbelkins/Code/Xcode/aws-sdk-swift/Sources/Services/AWSBedrockRuntime/models/Models.swift:882:40: error: type 'any Error' cannot conform to 'Decodable'
                    return try decoder.decode(responseBody: message.payload) as internalServerException
                                       ^
/Users/jbelkins/Code/Xcode/aws-sdk-swift/Sources/Services/AWSBedrockRuntime/models/Models.swift:882:40: note: only concrete types such as structs, enums and classes can conform to protocols
                    return try decoder.decode(responseBody: message.payload) as internalServerException
                                       ^
/Users/jbelkins/Code/Xcode/smithy-swift/Sources/ClientRuntime/Serialization/Decoder/ResponseDecoder.swift:9:10: note: required by instance method 'decode(responseBody:)' where 'T' = 'any Error'
    func decode<T: Decodable>(responseBody: Data) throws -> T
         ^
error: fatalError
```

### Expected Behavior

AWS SDK for Swift compiles & runs with all services.

### Current Behavior

AWS SDK for Swift fails to compile with compile errors listed above.

### Reproduction Steps

Update service models to include latest
Code-generate SDK
Run `swift test`

[...]

PR #1160, opened 2023-09-29
fix: Correct serde for restJson1 event streams that contain errors

jbelkins:
## Issue \#
#1159

## Description of changes
Fixes the following two compile errors that affect event streams on the `restJson1` protocol, which were discovered with the launch of the AWS Bedrock service.
- Corrects the serializer for an error received via an event stream to use the `properties` namespace.
- Corrects the message unmarshaller to use the correct type when unmarshalling an error from a data stream.

Note: this PR includes the model definitions for the Bedrock service, so the model can be codegen'd & built/tested on CI.


dayaffe (approved):
Looks good! Are there any tests we should or can add around this change?

sichanyoo (approved):
Lgtm

Issue #1163, opened 2023-10-03
Support initial-response in RPC-based protocols


dayaffe:

dayaffe:
Released in 0.28

PR #1165, opened 2023-10-03
feat: support initial-response in RPC based event streams

dayaffe:
## Issue \#
#1163 

## Description of changes
`awaitInitialResponse()` and `retrieveInitialResponse()` are used on in combination to allow code-generated service clients to get and populate initial-response fields. Together they wait for the **first message** to be processed and then return the value of initialResponse or `nil` if it is not set.

Companion PR: [#597](https://github.com/smithy-lang/smithy-swift/pull/597)

Known Risk: We cannot currently this feature in production (integration test) as there is no known service which supports initial-response over RPC protocols.

## Conventional Commits


sichanyoo (commented):
Left a question on a message encoder test case

sichanyoo (approved):
Lgtm!

Sources/Core/AWSClientRuntime/EventStream/AWSMessageDecoder.swift
```diff
@@ -88,6 +103,38 @@ extension AWSEventStream {
...
+        // It uses Swift's concurrency model to asynchronously return the data.
+        public func awaitInitialResponse() async -> Data? {
+            // The 'withCheckedContinuation' function is used to bridge asynchronous code
+            // that doesn't use Swift's concurrency model with code that does.
+            return await withCheckedContinuation { continuation in
+                // Here, we attempt to retrieve the initial response.
+                // Once the data is retrieved (or determined to be nil),
+                // the continuation is resumed with the result.
+                retrieveInitialResponse { data in
+                    continuation.resume(returning: data)
```
jbelkins, line 116: It is required, in all cases, that the continuation gets called (it's a rather large memory leak and a frozen path of execution if it doesn't.)

Considering that `retrieveInitialResponse()` will in some cases save the continuation to an instance var, how do we guarantee that it gets called in every case?

jbelkins: Agreed in separate conversation that immediately after running the continuation stored in `didProcessInitialMessage`, set it to nil so it is released and it cannot be called again.

jbelkins (approved):

Issue #1143, opened 2023-09-25
SRA Identity & Auth: [M2] Signer Changes


sichanyoo:
### Relevant issue(s):

Parent ticket: https://github.com/awslabs/aws-sdk-swift/issues/1119
Grandparent ticket: https://github.com/awslabs/aws-sdk-swift/issues/1095
Pre-existing ticket (used in Q4 planning): https://github.com/awslabs/aws-sdk-swift/issues/700

### Summary

Milestone 2 of SRA Identity & Auth Project

    * Add new interfaces / structs
        * Signer protocol
    * Refactor signer for event stream messages & URLs

sichanyoo:
PRs:
smithy-swift: https://github.com/smithy-lang/smithy-swift/pull/598
aws-sdk-swift: https://github.com/awslabs/aws-sdk-swift/pull/1167

PR #1167, opened 2023-10-04
feat: Move over signing flow from SigV4Middleware into AWSSigV4Signer

sichanyoo:
## Issue \#

Milestone ticket: https://github.com/awslabs/aws-sdk-swift/issues/1143

## Description of changes

Corresponding PR in `smithy-swift`: https://github.com/smithy-lang/smithy-swift/pull/598

Make pre-existing `AWSSigV4Signer` conform to new `Signer` protocol added to `smithy-swift` in corresponding PR.
Change is required to allow dynamic dispatching of `sign` method based on which concrete implementation of `Signer` is fetched from resolved auth scheme in SRA identity & auth flow.

Refactor `HttpContext` extension following same pattern as in the corresponding PR in `smithy-swift`.

## Conventional Commits

### DIFF
--- a/Sources/Core/AWSClientRuntime/HttpContextBuilder+Extension.swift
+++ b/Sources/Core/AWSClientRuntime/HttpContextBuilder+Extension.swift
@@ -20,7 +20,7 @@ extension HttpContext {
     func getRequestSignature() -> String {
         return attributes.get(key: AttributeKeys.requestSignature)!
     }
-    
+
     func getSigningAlgorithm() -> AWSSigningAlgorithm? {
         return attributes.get(key: AttributeKeys.signingAlgorithm)
     }
--- a/Sources/Core/AWSClientRuntime/HttpContextBuilder+Extension.swift
+++ b/Sources/Core/AWSClientRuntime/HttpContextBuilder+Extension.swift
@@ -6,7 +6,8 @@
 //
 
 import ClientRuntime
-import Foundation
+import struct Foundation.Date
+import struct Foundation.TimeInterval
 
 extension HttpContext {
     func getCredentialsProvider() -> (any CredentialsProviding)? {
--- a/Sources/Core/AWSClientRuntime/HttpContextBuilder+Extension.swift
+++ b/Sources/Core/AWSClientRuntime/HttpContextBuilder+Extension.swift
@@ -6,43 +6,31 @@
 //
 
 import ClientRuntime
-import struct Foundation.Date
+import Foundation
 
 extension HttpContext {
-    static let credentialsProvider = AttributeKey<(any CredentialsProviding)>(name: "CredentialsProvider")
-    static let region = AttributeKey<String>(name: "Region")
-    public static let signingName = AttributeKey<String>(name: "SigningName")
-    public static let signingRegion = AttributeKey<String>(name: "SigningRegion")
-    public static let signingAlgorithm = AttributeKey<String>(name: "SigningAlgorithm")
-    public static let requestSignature = AttributeKey<String>(name: "AWS_HTTP_SIGNATURE")
-
     func getCredentialsProvider() -> (any CredentialsProviding)? {
-        return attributes.get(key: HttpContext.credentialsProvider)
+        return attributes.get(key: AttributeKeys.credentialsProvider)
     }
 
     func getRegion() -> String? {
-        return attributes.get(key: Self.region)
+        return attributes.get(key: AttributeKeys.region)
+    }
+
+    func getRequestSignature() -> String {
+        return attributes.get(key: AttributeKeys.requestSignature)!
+    }
+    
+    func getSigningAlgorithm() -> AWSSigningAlgorithm? {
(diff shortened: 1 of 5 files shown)


Sources/Core/AWSClientRuntime/Signing/AWSSigV4Signer.swift
```diff
@@ -9,7 +9,89 @@ import AwsCommonRuntimeKit
...
+
+        let signingConfig = try constructSigningConfig(identity: identity, signingProperties: signingProperties)
+
+        let unsignedRequest = requestBuilder.build()
+        let crtUnsignedRequest: HTTPRequestBase
+        if isBidirectionalStreamingEnabled {
+            crtUnsignedRequest = try unsignedRequest.toHttp2Request()
+        } else {
+            crtUnsignedRequest = try unsignedRequest.toHttpRequest()
+        }
```
dayaffe, line 36:
```
let unsignedRequest = requestBuilder.build()
let crtUnsignedRequest = isBidirectionalStreamingEnabled ? 
        try unsignedRequest.toHttp2Request() : 
        try unsignedRequest.toHttpRequest()
```

Sources/Core/AWSClientRuntime/Signing/AWSSigV4Signer.swift
```diff
@@ -9,7 +9,89 @@ import AwsCommonRuntimeKit
...
+
+        let crtSignedRequest = try await Signer.signRequest(
+            request: crtUnsignedRequest,
+            config: signingConfig.toCRTType()
+        )
+        let signedRequest = requestBuilder.update(from: crtSignedRequest, originalRequest: unsignedRequest)
+        return signedRequest
```
dayaffe, line 43: return directly rather than store in a variable that is unused and use a comment string to indicate that you are getting the signedRequest for readability

Sources/Core/AWSClientRuntime/Signing/AWSSigV4Signer.swift
```diff
@@ -9,7 +9,89 @@ import AwsCommonRuntimeKit
...
+    public typealias IdObj = Credentials
+
+    public func sign(
+        requestBuilder: SdkHttpRequestBuilder,
+        identity: Credentials,
+        signingProperties: ClientRuntime.Attributes
+    ) async throws -> SdkHttpRequestBuilder {
+        guard let isBidirectionalStreamingEnabled = signingProperties.get(
+            key: AttributeKeys.bidirectionalStreaming
+        ) else {
```
dayaffe, line 22: if its short enough preferred swift spacing I believe is to have it on one line other than the error handling

```
guard let isBidirectionalStreamingEnabled = signingProperties.get(key: AttributeKeys.bidirectionalStreaming) else {
    throw ClientError.authError("Signing properties must contain a flag for bidirectional streaming.")
}
```

dayaffe (changes requested):
Just a few small things :)

Sources/Core/AWSClientRuntime/Signing/AWSSigV4Signer.swift
sichanyoo: Originally had it that way, but swiftlint complains about length - so had to shorten it this way

Sources/Core/AWSClientRuntime/Signing/AWSSigV4Signer.swift
sichanyoo: Changed as suggested

Sources/Core/AWSClientRuntime/Signing/AWSSigV4Signer.swift
sichanyoo: Changed as suggested.

dayaffe (approved):
Looks good

Sources/Core/AWSClientRuntime/HttpContextBuilder+Extension.swift
```diff
@@ -83,51 +72,68 @@ extension HttpContext {
...
     @discardableResult
-    public func withRequestSignature(value: String) -> HttpContextBuilder {
-        self.attributes.set(key: HttpContext.requestSignature, value: value)
+    public func withSigningRegion(value: String?) -> HttpContextBuilder {
+        self.attributes.set(key: AttributeKeys.signingRegion, value: value)
         return self
     }
 }
+
+extension AttributeKeys {
```
jbelkins, line 123: If you put these keys in the `AttributeKey` namespace (i.e. their own type), then you can refer to them without the namespace:
```
        self.attributes.set(key:.signingRegion, value: value)
```
instead of 
```
        self.attributes.set(key: AttributeKeys.signingRegion, value: value)
```

sichanyoo: https://github.com/smithy-lang/smithy-swift/pull/598#discussion_r1348019020

Reasoning in above link, from previous comment on same issue.

Issue #1144, opened 2023-09-25
SRA Identity & Auth: [M3] AuthScheme Changes


sichanyoo:
### Relevant issue(s):

Parent ticket: https://github.com/awslabs/aws-sdk-swift/issues/1119
Grandparent ticket: https://github.com/awslabs/aws-sdk-swift/issues/1095
Pre-existing ticket (used in Q4 planning): https://github.com/awslabs/aws-sdk-swift/issues/700

### Summary

Milestone 3 of SRA Identity & Auth Project

    * Add new interfaces / structs
        * AuthScheme struct
        * AuthOption struct
        * SelectedAuthScheme struct
        * AuthSchemeResolver protocol
        * AuthSchemeResolverParameters protocol
        * IdentityResolverConfiguration protocol
        * DefaultIdentityResolverConfiguration struct

sichanyoo:
PRs:
- smithy-swift: https://github.com/smithy-lang/smithy-swift/pull/601
- aws-sdk-swift: https://github.com/awslabs/aws-sdk-swift/pull/1174

PR #1174, opened 2023-10-09
feat: auth scheme changes

sichanyoo:
## Issue \#

Issue: https://github.com/awslabs/aws-sdk-swift/issues/1144

Corresponding PR in `smithy-swift`: https://github.com/smithy-lang/smithy-swift/pull/601

## Description of changes

Adds auth scheme struct for SigV4 authentication.

## Conventional Commits


Sources/Core/AWSClientRuntime/Auth/AuthSchemes/SigV4AuthScheme.swift
```diff
@@ -0,0 +1,15 @@
...
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import Foundation
+import ClientRuntime
+
+public struct SigV4AuthScheme: ClientRuntime.AuthScheme {
+    public let schemeId: String = "aws.auth#sigv4"
+    public let signer: any ClientRuntime.Signer = AWSSigV4Signer()
```
dayaffe, line 13: I am cautious about using `any` as there are type-safety and performance implications. Is there any issue with using generics here instead?

```
public struct SigV4AuthScheme<S: ClientRuntime.Signer>: ClientRuntime.AuthScheme {
    public let schemeId // same
    public let signer: S
    public let idType // same

    public init(signer: S = AWSSigV4Signer()) {
        self.signer = signer
    }
}
```

Example Usages: 
- `let scheme = SigV4AuthScheme()` defaults to AWSSigV4Signer()
- `let scheme = SigV4AuthScheme<AWSSigV4Signer>()` same as above but explicit
- `let scheme = SigV4AuthScheme(signer: SomeCustomSigner())`

dayaffe (commented):
See comment on using `any`

Sources/Core/AWSClientRuntime/Auth/AuthSchemes/SigV4AuthScheme.swift
dayaffe: Its interesting because if the code is cleaner we may want to keep using `any` over generics since the code is cleaner. But, if we expect any function using `any` to be called a lot we should not use it and prefer generics. It looks like historically, the evolution proposal’s motivation paragraph for `any` mentions that “Existential types in Swift have significant limitations and performance implications.” ([SE-0335](https://github.com/apple/swift-evolution/blob/main/proposals/0335-existential-any.md)) and “Existential types are also significantly more expensive than using concrete types.” ([SE-0335](https://github.com/apple/swift-evolution/blob/main/proposals/0335-existential-any.md)).

Tldr; `any` was a keyword introduced in Swift 5.x and required in Swift 6 so that the user acknowledges that they are using an existential type which comes with certain drawbacks compared to generics. The same concept was already supported prior to Swift 6 implicitly.

Let me know your thoughts!

jbelkins: I do agree that generics should be used when possible; problem is that generics don't work if the type can vary at runtime.  However, I don't expect that using `any` for one signer, identity resolver, etc. per API request will cause a measurable performance hit.

sichanyoo: I actually changed the `Signer` protocol to NOT have associated type and used generic function instead to address exactly this issue (in previous pair of PRs). I forgot to delete the unnecessary `any` keywords in here as well as back in `AuthScheme` protocol in `smithy-swift`. They have been updated now.

dayaffe (approved):

Issue #1146, opened 2023-09-25
SRA Identity & Auth: [M5] Codegen Changes


sichanyoo:
### Relevant issue(s):

Parent ticket: https://github.com/awslabs/aws-sdk-swift/issues/1119
Grandparent ticket: https://github.com/awslabs/aws-sdk-swift/issues/1095
Pre-existing ticket (used in Q4 planning): https://github.com/awslabs/aws-sdk-swift/issues/700

### Summary

Milestone 5 of SRA Identity & Auth Project

    * Add codegen for service specific AuthSchemeResolverParameters default implementation
    * Add codegen for service specific, model based AuthSchemeResolver protocol and default implementation
    * Add codegen for default values of service client auth schemes and auth scheme resolver
    * Update AWSServiceSpecificConfiguration - add AuthSchemes field and AuthSchemeResolver field for end-user customizations
    * Refactor client operation codegen to use the new middlewares, and stop using SigV4Middleware

sichanyoo:
PRs: 
- https://github.com/smithy-lang/smithy-swift/pull/609
- https://github.com/awslabs/aws-sdk-swift/pull/1190

PR #1190, opened 2023-10-25
feat: codegen changes

sichanyoo:
## Issue \#

https://github.com/awslabs/aws-sdk-swift/issues/1146

Companion PR in `smithy-swift`: https://github.com/smithy-lang/smithy-swift/pull/609

## Description of changes

- Add `authSchemeResolver: AWSAuthSchemeResolver` field to `AWSServiceSpecificConfiguration` and update constructors in both `AWSServiceSpecificConfiguration` and `AWSClientConfiguration` accordingly
- Add `authSchemes: [ClientRuntime.AuthScheme]?` field to `AWSClientConfiguration` and update constructors accordingly
- Add signing properties customization to `SigV4AuthScheme`, filling in missing properties that `SigV4Signer` needs
- Move Smithy generic methods in middleware context (`HttpContext`) to Smithy repo, and change some middleware context functions that are needed in `SigV4AuthScheme::customizeSigningProperties` from internal to public
- Add codegen for default auth scheme resolver and auth scheme config values, and middleware context construction code that adds auth scheme resolver and auth schemes from client config
- Connect auth scheme resolver generation in `smithy-swift` into existing codegen flow via `super.renderInternals`
- Update codegen test cases to reflect new changes.


Sources/Core/AWSClientRuntime/Auth/AuthSchemes/SigV4AuthScheme.swift
```diff
@@ -12,10 +12,38 @@ public struct SigV4AuthScheme: ClientRuntime.AuthScheme {
     public let signer: ClientRuntime.Signer = AWSSigV4Signer()
     public let idKind: ClientRuntime.IdentityKind = .aws
 
+    public init() {}
+
     public func customizeSigningProperties(signingProperties: Attributes, context: HttpContext) -> Attributes {
         var copy = signingProperties
```
dayaffe, line 18: nit: copy is a bit unclear maybe `updatedSigningProperties` or `signingPropertiesWithAdditionalAttributes`

Tests/Core/AWSClientRuntimeTests/AWSClientConfigurationTests.swift
```diff
@@ -74,17 +75,40 @@ class AWSClientConfigurationTests: XCTestCase {
...
 
     var endpointResolver: EndpointResolver
 
-    init(endpointResolver: EndpointResolver?) throws {
+    init(endpointResolver: EndpointResolver?,
+         authSchemeResolver: ClientRuntime.AuthSchemeResolver?,
+         authSchemes: [ClientRuntime.AuthScheme]?
+    ) throws {
         self.endpointResolver = endpointResolver ?? EndpointResolver()
+        self.authSchemes = authSchemes ?? []
```
dayaffe, line 93: Question: Will we need to do anything new/special to support optionalAuth trait?
https://smithy.io/2.0/spec/authentication-traits.html#optionalauth-trait

Nit: If `[]` is a reference to supporting DisabledAuth maybe we can add a comment with a link to what this means

dayaffe (commented):
A few small comments and a question

Sources/Core/AWSClientRuntime/Auth/AuthSchemes/SigV4AuthScheme.swift
sichanyoo: Changed to `updatedSigningProperties` as suggested.

Tests/Core/AWSClientRuntimeTests/AWSClientConfigurationTests.swift
sichanyoo: SRA Identity & Auth work will handle optionalAuth trait as part of it (in fact the codegen for service specific `AuthSchemeResolver`s that's added in the [smithy-swift PR](https://github.com/smithy-lang/smithy-swift/pull/609) currently handles optoinal auth, there's nothing new/special that's needed. 

Also - `[]` isn't a reference to anything, it was just the simplest filler that compiler stopped complaining about - no additional tests have been added in this file that tests auth schemes.

dayaffe (approved):

Sources/Core/AWSClientRuntime/AWSServiceSpecificConfiguration.swift
```diff
@@ -31,5 +34,19 @@ public protocol AWSServiceSpecificConfiguration {
...
+    init(
+        endpointResolver: AWSServiceEndpointResolver?,
+        authSchemeResolver: ClientRuntime.AuthSchemeResolver?,
+        authSchemes: [ClientRuntime.AuthScheme]?
+    ) throws
+
+    /// List of configured auth schemes for the service.
+    ///
+    /// If none is provided at compile time, value will default to auth schemes supported by the service as defined in its model.
+    var authSchemes: [ClientRuntime.AuthScheme] { get }
```
jbelkins, line 46: Why does this property need to be in `ServiceSpecificConfiguration` and not in `AWSClientConfiguration`?

`endpointResolver` is here because every service has its own type of endpoint resolver.  You can see it is of an `associatedtype` of this protocol, `AWSServiceEndpointResolver`, which is determined by the specific AWS service being configured.  That's what makes `endpointResolver` service-specific.

Since `authSchemes` is always of a single type that is defined in Smithy, it would be better placed in `AWSClientConfiguration`.

Same comment goes for `authSchemeResolver` as well.

jbelkins (changes requested):

Sources/Core/AWSClientRuntime/AWSServiceSpecificConfiguration.swift
sichanyoo: I see, that's a good point. However every service has its own type of auth scheme resolver as well, in the format of `ServiceNameAuthSchemeResolver`. I'll update this so that `authSchemes` is moved to non-service specific config, and have `authSchemeResolver` be an associated type as well.

sichanyoo: Actually, on second thought, I think both naturally belong in `ServiceSpecificConfiguration` because every service has its own type of auth scheme resolver, and every service has its own unique set of auth schemes configured for it based on the smithy model. And in both cases they need to be code-generated based on the model & the service name.

jbelkins: It's not the value that goes in the property that determines where it should go, it's the property itself.  If the property & its type are not specific to the service, it should go on the main config and not on `ServiceSpecificConfiguration`.

Reason is that `ServiceSpecificConfiguration` is code-generated.  So, adding two properties to it adds bloat to all ~350 service-specific configurations when we can instead add it to the one shared config.

sichanyoo: Each service has its own unique auth scheme resolver type. And each service has a unique set of auth schemes it supports, which has to be code generated from its model. I think given these two conditions they both belong in `ServiceSpecificConfiguration`. - Will discuss further offline

sichanyoo: After further offline discussion, concluded to move `authSchemes` property to general AWS config, and `authSchemeResolver` to service specific config. Reasoning being there exists service specific resolver _type_ for `authSchemeResolver`, whereas there does not exist service specific type for `authSchemes`.

jbelkins (approved):

Issue #837, opened 2023-01-23
S3 : preSignedURL is not documented / not on the correct class


sebsto:
### Describe the feature

- The S3 `presignURL` should be on the `S3Client` class instead on the `GetObjectInput` class
- This method should be part of the top level documentation for the usage of S3 

### Use Case

I came accros a use case where developer want to use `AsyncImage` to download and display a picture from S3. I could not find a S3 “presignedUrl()” API in the S3 client like there is on other platforms (Java, .Net, JS, and  — I think — Python)) - I even thought it was part of the S3 API but looks like it is not, it is a SDK-level addition.

Finally, Ganesh pointed to me to

```
extension GetObjectInput {
    public func presignURL(config: S3ClientConfigurationProtocol, expiration: Foundation.TimeInterval) async throws -> ClientRuntime.URL? {
```

### Proposed Solution

Move the method to `S3Client` and write a section in the documentation about it.

### Acknowledgements

- [X] I may be able to implement this feature request
- [X] This feature might incur a breaking change

ganeshnj:
Few things
**- what other SDKs are doing**  we should be consistent
- no difference between in underlying impl except Http call
- it will be a breaking change

sebsto:
**Java, JavaScript**
The S3Signer class is a wrapper around GetObject and PutObject requests 
https://docs.aws.amazon.com/sdk-for-java/latest/developer-guide/examples-s3-presign.html#generate-presignedurl
https://aws.amazon.com/blogs/developer/generate-presigned-url-modular-aws-sdk-javascript/

**Python**  It is an fun on the S3 client object.

You can see code sample for multiple AWS SDK here 
https://docs.aws.amazon.com/AmazonS3/latest/userguide/example_s3_Scenario_PresignedUrl_section.html

sichanyoo:
PR: https://github.com/awslabs/aws-sdk-swift/pull/1198

sebsto:
Fixed in https://github.com/awslabs/aws-sdk-swift/releases/tag/0.31.0

Thank you !

PR #1198, opened 2023-10-26
feat: expose presign / presignURL methods in Models.swift to service client object

sichanyoo:
## Issue \#

https://github.com/awslabs/aws-sdk-swift/issues/837
Companion PR: https://github.com/smithy-lang/smithy-swift/pull/613

## Background
Currently, `presign` / `presignURL` methods are in `Model.swift` of each service SDK directory rather than in the client file, e.g., `S3Client.swift`. 

Other SDKs expose this API as methods that can be invoked with service client object.

Codegen for exposing these APIs in client object have been added where applicable.

## Description of changes

- Code generation for wrapper methods have been added to client generation.

*Note that the changes in Polly, S3, and STS client files have been included in the PR to show codegen output for the reviewers. 

Customers will be able to presign requests or get presigned URL for requests from the service client object they create.

## Conventional Commits


Sources/Services/AWSPolly/PollyClient.swift
```diff
@@ -481,3 +481,23 @@ extension PollyClient: PollyClientProtocol {
     }
 
 }
+
+extension PollyClient {
+    public func presignURLSynthesizeSpeech(input: SynthesizeSpeechInput, expiration: Foundation.TimeInterval) async throws -> ClientRuntime.URL {
+        let presignedURL = try await input.presignURL(config: config, expiration: expiration)
+        guard let presignedURL else {
```
jbelkins, line 488: nit: you can combine this & the above statement

codegen/smithy-aws-swift-codegen/src/main/kotlin/software/amazon/smithy/aws/swift/codegen/customization/presignable/PresignableUrlIntegration.kt
```diff
@@ -144,6 +150,27 @@ class PresignableUrlIntegration(private val presignedOperations: Map<String, Set
...
+    private fun renderPresignURLAPIInServiceClient(
+        writer: SwiftWriter,
+        clientName: String,
+        op: OperationShape,
+        inputType: String
+    ) {
+        writer.apply {
+            openBlock("extension $clientName {", "}") {
+                val params = listOf("input: $inputType", "expiration: Foundation.TimeInterval")
+                val returnType = "ClientRuntime.URL"
```
jbelkins, line 162: What is `ClientRuntime.URL`?  Why can't this just be Foundation's URL type?

codegen/smithy-aws-swift-codegen/src/main/kotlin/software/amazon/smithy/aws/swift/codegen/customization/presignable/PresignableUrlIntegration.kt
```diff
@@ -144,6 +150,27 @@ class PresignableUrlIntegration(private val presignedOperations: Map<String, Set
...
+                openBlock("public func presignURL${op.toUpperCamelCase()}(${params.joinToString()}) async throws -> $returnType {", "}") {
+                    write("let presignedURL = try await input.presignURL(config: config, expiration: expiration)")
+                    openBlock("guard let presignedURL else {", "}") {
+                        write("throw ClientError.unknownError(\"Returned URL from input.presignURL() was nil.\")")
```
jbelkins, line 166: To help make it clear where this error originated, what do you think of including the operation name in the error message?

jbelkins (changes requested):
For naming, what do you think about:
```
func presignedURLForGetObject(input: GetObjectInput, expiration: Foundation.TimeInterval) -> Foundation.URL
```
and
```
func presignedRequestForGetObject(input: GetObjectInput, expiration: Foundation.TimeInterval) -> ClientRuntime.SdkHttpRequest
```
It's a little more verbose, but makes it a little more clear what the method does & what is being returned.

Maybe we should also consider returning a `Foundation.HTTPURLRequest` for request instead of a SDK type, since the point of the method is to get the request out of the SDK & use it elsewhere.

codegen/smithy-aws-swift-codegen/src/main/kotlin/software/amazon/smithy/aws/swift/codegen/customization/presignable/PresignableUrlIntegration.kt
sichanyoo: It's a primitive type extension of `Foundation.URL` as seen here: [link](https://github.com/smithy-lang/smithy-swift/blob/53fbde27f536f4287edfd4ff08bb5ebe8dc75169/Sources/ClientRuntime/PrimitiveTypeExtensions/URL%2BExtension.swift#L8), and this is the type returned by the wrapped method in `Models.swift`, as seen here: [link](https://github.com/awslabs/aws-sdk-swift/blob/8d8bacaf1cd81e9d7534a1ea28fd525bf6e179c3/Sources/Services/AWSS3/models/Models.swift#L10664)

codegen/smithy-aws-swift-codegen/src/main/kotlin/software/amazon/smithy/aws/swift/codegen/customization/presignable/PresignableUrlIntegration.kt
sichanyoo: That would be great, will add that in

Sources/Services/AWSPolly/PollyClient.swift
sichanyoo: changed as suggested.

codegen/smithy-aws-swift-codegen/src/main/kotlin/software/amazon/smithy/aws/swift/codegen/customization/presignable/PresignableUrlIntegration.kt
sichanyoo: Given that the extension should be applied anyways, will change the return type of wrapper method to `Foundation.URL`.

Sources/Services/AWSSTS/STSClient.swift
```diff
@@ -462,3 +462,12 @@ extension STSClient: STSClientProtocol {
     }
 
 }
+
+extension STSClient {
+    public func presignRequestForGetCallerIdentity(input: GetCallerIdentityInput, expiration: Foundation.TimeInterval) async throws -> URLRequest {
+        guard let presignedRequest = try await input.presign(config: config, expiration: expiration) else {
+            throw ClientError.unknownError("Returned request from input.presign() was nil for the operation GetCallerIdentity.")
```
jbelkins, line 469: The error message should not make reference to SDK internals; I would say something to the effect of "A presigned URL for \<operation\> could not be created."

Thought of another way, the current message would not make sense anymore if we refactored internally.

codegen/smithy-aws-swift-codegen/src/main/kotlin/software/amazon/smithy/aws/swift/codegen/PresignerGenerator.kt
```diff
@@ -50,6 +51,16 @@ class PresignerGenerator : SwiftIntegration {
                 var serviceConfig = AWSServiceConfig(writer, protoCtx)
                 renderPresigner(writer, ctx, delegator, op, inputType, serviceConfig)
             }
+            // Expose presign-request as a method for service client object
+            val symbol = protoCtx.symbolProvider.toSymbol(protoCtx.service)
+            protoCtx.delegator.useFileWriter("./${ctx.settings.moduleName}/${symbol.name}.swift") { writer ->
+                renderPresignAPIInServiceClient(writer, symbol.name, op, inputType)
+                // Add special-case import statement for Linux
+                writer.write("// In Linux, Foundation.URLRequest is moved to FoundationNetworking.")
```
jbelkins, line 59: I would not render a comment like this into generated code.  Instead, it can be a comment here in the codegen source file.

Sources/Services/AWSSTS/STSClient.swift
```diff
@@ -462,3 +462,12 @@ extension STSClient: STSClientProtocol {
...
+        }
+        return try await presignedRequest.toURLRequest()
```
jbelkins, line 471: `toURLRequest()` is not a Swift-convention way to make such a conversion. `to...` is more of a Java-land convention.

I would prefer extending `URLRequest` with a custom initializer, i.e.
```
return try await URLRequest(sdkHttpRequest: presignedRequest)
```

Sources/Services/AWSPolly/PollyClient.swift
```diff
@@ -481,3 +481,21 @@ extension PollyClient: PollyClientProtocol {
...
+    public func getPresignedURLForSynthesizeSpeech(input: SynthesizeSpeechInput, expiration: Foundation.TimeInterval) async throws -> Foundation.URL {
+        guard let presignedURL = try await input.presignURL(config: config, expiration: expiration) else {
+            throw ClientError.unknownError("Returned URL from input.presignURL() was nil for the operation SynthesizeSpeech.")
+        }
+        return presignedURL
+    }
+}
+
+extension PollyClient {
+    public func presignRequestForSynthesizeSpeech(input: SynthesizeSpeechInput, expiration: Foundation.TimeInterval) async throws -> URLRequest {
```
jbelkins, line 495: Why do URL & request use different language in the method name? They should be consistent with one another:
`presignRequestForSynthesizeSpeech` and `presignURLForSynthesizeSpeech`

jbelkins (changes requested):

codegen/smithy-aws-swift-codegen/src/main/kotlin/software/amazon/smithy/aws/swift/codegen/PresignerGenerator.kt
sichanyoo: Changed as suggested.

Sources/Services/AWSPolly/PollyClient.swift
sichanyoo: Changed as suggested.

Sources/Services/AWSSTS/STSClient.swift
sichanyoo: Changed as suggested.

Sources/Services/AWSSTS/STSClient.swift
sichanyoo: I see, but I think it makes sense to keep it this way given this is the convention used in `SdkHttpRequest`, as shown here: https://github.com/smithy-lang/smithy-swift/blob/6037a8157082d00da6ff1be1462022dcaccb3662/Sources/ClientRuntime/Networking/Http/SdkHttpRequest.swift#L60-L86

Sources/Services/AWSPolly/PollyClient.swift
```diff
@@ -481,3 +481,25 @@ extension PollyClient: PollyClientProtocol {
     }
 
 }
+
+extension PollyClient {
+    public func presignURLForSynthesizeSpeech(input: SynthesizeSpeechInput, expiration: Foundation.TimeInterval) async throws -> Foundation.URL {
+        guard let presignedURL = try await input.presignURL(config: config, expiration: expiration) else {
```
dayaffe, line 487: its better to separate this into 2 lines for debugging/readability

```
let presignedURL = try await input.presignURL(config: config, expiration: expiration)
guard let url = presignedURL else {
     // throw error
} 
// use url here
```

This is better imo for 2 reasons
- if an error occurs during the function call it won't be confused with an error being thrown violating the guard that is distinct from nullity
- it makes it more readable as we separate concerns into first we populate a variable with the return of the call the input.presignURL and then we guard on nullity

Sources/Services/AWSPolly/PollyClient.swift
```diff
@@ -481,3 +481,25 @@ extension PollyClient: PollyClientProtocol {
...
+            throw ClientError.unknownError("Could not generate presigned URL for the operation SynthesizeSpeech.")
+        }
+        return presignedURL
+    }
+}
+
+extension PollyClient {
+    public func presignRequestForSynthesizeSpeech(input: SynthesizeSpeechInput, expiration: Foundation.TimeInterval) async throws -> URLRequest {
+        guard let presignedRequest = try await input.presign(config: config, expiration: expiration) else {
```
dayaffe, line 496: same as above / apply elsewhere in PR

dayaffe (changes requested):

Sources/Services/AWSPolly/PollyClient.swift
sichanyoo: Refer to this comment: https://github.com/awslabs/aws-sdk-swift/pull/1198#discussion_r1373860445
That's actually what I had at first, but I think the error that's thrown should have a message that allows that distinction anyways.

dayaffe: Oh I mean I would defer to @jbelkins, but I would prefer to read/debug this code with it on 2 separate lines. Either way the same error is thrown. But if `try await input.presignURL()` throws an error or if the guard statement fails because that same line returns a null value (but does not throw an error) the debugger would show the same exact line as having a failure despite 2 separate reasons (failed guard vs error thrown in method call itself) which is confusing

dayaffe: to explain further my reasoning:

```
let x = try await function() // I make function call
guard let _x = x else {
   throw error(something unknown happen or x is null!)
}
```

`try await function()` is marked with try because it has its own reasons for throwing an error. Separating this out to 2 lines imo makes it clear to the reader that by using a `guard` statement I am adding additional conditions that arent inherent to calling the function

Sources/Services/AWSSTS/STSClient.swift
jbelkins: We should reduce the amount of code that doesn't follow Swift convention rather than increase it.  The code you cite should not have been named that way either.

Sources/Services/AWSPolly/PollyClient.swift
sichanyoo: Changed as suggested.

Sources/Services/AWSPolly/PollyClient.swift
sichanyoo: Reverted back to original, after discussion with David: https://github.com/awslabs/aws-sdk-swift/pull/1198#discussion_r1378110524

Sources/Services/AWSSTS/STSClient.swift
sichanyoo: Fair, I'll create extension and put the changes up.

dayaffe: I agree with Josh let's be as Swifty as possible rather than rely on previous code examples :) Especially true as we deprecate older swift versions and gain access to newer Swift syntax features!

jbelkins: I don't know if it was written by developers not familiar with Swift convention or just borrowed from elsewhere without adapting it to Swift, but you will see some code (like this) in our projects that doesn't follow Swift community convention.  We don't need to (can't!) change it all overnight, but we do need to chip away at it when the opportunity presents itself.

Sources/Services/AWSPolly/PollyClient.swift
jbelkins: Fine as-is if you prefer two steps.

dayaffe (approved):

Sources/Services/AWSPolly/PollyClient.swift
```diff
@@ -481,3 +481,27 @@ extension PollyClient: PollyClientProtocol {
...
+    public func presignRequestForSynthesizeSpeech(input: SynthesizeSpeechInput, expiration: Foundation.TimeInterval) async throws -> URLRequest {
+        let presignedRequest = try await input.presign(config: config, expiration: expiration)
+        guard let presignedRequest else {
+            throw ClientError.unknownError("Could not presign the request for the operation SynthesizeSpeech.")
+        }
+        return try await URLRequest(sdkRequest: presignedRequest)
+    }
+}
+
+#if canImport(FoundationNetworking)
```
jbelkins, line 505: Swift will accept an `import` in mid-source file, but it's definitely not a normal practice.

We should add the ability in code-generation (along with the import functionality we already have) to code-generate this pattern (i.e. `#if canImport ...`) at the top, with the other imports.

Sources/Services/AWSS3/S3Client.swift
```diff
@@ -5023,3 +5023,57 @@ extension S3Client: S3ClientProtocol {
     }
 
 }
+
+extension S3Client {
+    public func presignURLForGetObject(input: GetObjectInput, expiration: Foundation.TimeInterval) async throws -> Foundation.URL {
```
jbelkins, line 5028: All of these public `presign...()` methods should have documentation comments.

Sources/Services/AWSPolly/PollyClient.swift
sichanyoo: Added preprocessor conditional import functionality to `SwiftWriter` in companion PR in smithy-swift.

Sources/Services/AWSS3/S3Client.swift
sichanyoo: Added documentation generation to codegen.

Sources/Services/AWSPolly/PollyClient.swift
```diff
@@ -481,3 +484,37 @@ extension PollyClient: PollyClientProtocol {
     }
 
 }
+
+extension PollyClient {
+    /// Presigns the URL for SynthesizeSpeech operation with the given input object SynthesizeSpeechInput.
```
jbelkins, line 489: The doc comment should include the same docs that are rendered onto the regular operation.

codegen/smithy-aws-swift-codegen/src/main/kotlin/software/amazon/smithy/aws/swift/codegen/customization/presignable/PresignableUrlIntegration.kt
```diff
@@ -144,6 +150,40 @@ class PresignableUrlIntegration(private val presignedOperations: Map<String, Set
...
+        clientName: String,
+        op: OperationShape,
+        inputType: String
+    ) {
+        writer.apply {
+            openBlock("extension $clientName {", "}") {
+                val params = listOf("input: $inputType", "expiration: Foundation.TimeInterval")
+                val returnType = "Foundation.URL"
+                renderDocForPresignURLAPI(this, op, inputType)
+                openBlock("public func presignURLFor${op.toUpperCamelCase()}(${params.joinToString()}) async throws -> $returnType {", "}") {
```
jbelkins, line 164: Function name should start with `presignedURLFor...`, not `presignURLFor`.  Same for the request functions.

codegen/smithy-aws-swift-codegen/src/main/kotlin/software/amazon/smithy/aws/swift/codegen/PresignerGenerator.kt
```diff
@@ -105,6 +119,40 @@ class PresignerGenerator : SwiftIntegration {
...
+        clientName: String,
+        op: OperationShape,
+        inputType: String
+    ) {
+        writer.apply {
+            openBlock("extension $clientName {", "}") {
+                val params = listOf("input: $inputType", "expiration: Foundation.TimeInterval")
+                val returnType = "URLRequest"
+                renderDocForPresignAPI(this, op, inputType)
+                openBlock("public func presignRequestFor${op.toUpperCamelCase()}(${params.joinToString()}) async throws -> $returnType {", "}") {
```
jbelkins, line 133: `public func presignedRequestFor...` (see comment on URL presign renderer.)

jbelkins (changes requested):
A couple last changes, then I'll be ready to approve.

codegen/smithy-aws-swift-codegen/src/main/kotlin/software/amazon/smithy/aws/swift/codegen/customization/presignable/PresignableUrlIntegration.kt
sichanyoo: Changed as suggested.

codegen/smithy-aws-swift-codegen/src/main/kotlin/software/amazon/smithy/aws/swift/codegen/PresignerGenerator.kt
sichanyoo: Changed as suggested.

Sources/Services/AWSPolly/PollyClient.swift
sichanyoo: I don't understand what you mean by this - do you mean the links to AWS docs (the ones in Smithy model under @documentation trait)? If so, they aren't available for these presign operations because these are added at SDK level.

sichanyoo: Discussed offline - will add docs as mentioned.

codegen/smithy-aws-swift-codegen/src/main/kotlin/software/amazon/smithy/aws/swift/codegen/PresignerGenerator.kt
jbelkins: I don't see this change in the latest diff.  Maybe check to see if you've pushed all to github?

codegen/smithy-aws-swift-codegen/src/main/kotlin/software/amazon/smithy/aws/swift/codegen/customization/presignable/PresignableUrlIntegration.kt
jbelkins: Same here, rendered function is still named `presignURLFor...`

jbelkins (approved):

Issue #1147, opened 2023-09-25
SRA Identity & Auth: [M6] Customizations


sichanyoo:
### Relevant issue(s):

Parent ticket: https://github.com/awslabs/aws-sdk-swift/issues/1119
Grandparent ticket: https://github.com/awslabs/aws-sdk-swift/issues/1095
Pre-existing ticket (used in Q4 planning): https://github.com/awslabs/aws-sdk-swift/issues/700

### Summary

Milestone 6 of SRA Identity & Auth Project

    * Add rules-based AuthSchemeResolvers for S3 and EventBridge as customizations
    * Refactor message signing middleware flow and presign URL middleware flow to use resolved signer

sichanyoo:
Blocked because awaiting reviews on https://github.com/awslabs/aws-sdk-swift/issues/1146.

sichanyoo:
PRs:
- https://github.com/awslabs/aws-sdk-swift/pull/1221
- https://github.com/smithy-lang/smithy-swift/pull/618

PR #1221, opened 2023-11-10
feat: customizations

sichanyoo:
## Issue \#

https://github.com/awslabs/aws-sdk-swift/issues/1147

Companion PR in smithy-swift: https://github.com/smithy-lang/smithy-swift/pull/618

## Description of changes

- Add `SigV4AAuthScheme`, all same as `SigV4AuthScheme` except the scheme ID. 
- Refactor `SigV4AuthScheme::customizeSigningProperties` for better scalability and readability.
- Add `SigV4Util` to aggregate signing properties customization info into one place & make `SigV4AuthScheme::customizeSigningProperties` and `SigV4AAuthScheme::customizeSignignProperties` more readable.
- Change event stream message signing flow from passing in a `AWSSigV4Signer` to passing a function that retrieves signer from the resolved auth scheme (`SelectedAuthScheme`) in the middleware context.
- Move setter, getter, and key for `requestSignature` from here to `smithy-swift`.
- Modify `AWSSigV4Signer` to conform to modified `Signer` protocol that has `signEvent` API with type-erased argument. 
- Connect rules based auth scheme resolver generator to existing codegen flow via `AWSHttpProtocolCustomizations::renderInternals`.
- Add conversion init codegen for endpoint params, where it takes an instance of auth scheme params and returns endpoint params. Used by rules based auth scheme resolvers, to convert auth scheme params argument given to it to endpoint params, which is then used to call endpoint resolver (rules based auth scheme resolver "delegates" auth scheme resolution to endpoint resolver).
- Remove `SigV4Middleware` usage and codegen-time signing param values customization from presigner codegen (both presign request & presign url flows). Re-locate signing param values customization from codegen to signing properties customization hook in auth schemes. 
- Add rules based auth scheme resolver generator. 
- Add codegen for saving endpoint params to middleware context, to be used by `constructParameters` function of rules based auth scheme resolvers.
- Remove 3 tests from `PresignableUrlIntegrationTests.kt` that tests signing param values customization, which is now moved from previous signing flow involving `SigV4Config` and `SigV4Middleware` to SRA flow. These tests will be covered in new tests that test signing properties customization.
- Add a codegen test for rules based auth scheme resolver generator. Uses a Smithy model named "S3", to trigger rules based auth scheme resolver generation.
- Misc. - update test cases to reflect new changes.
- Misc - some kotlin lint warning changes (e.g., XYZ can be a val instead of var, etc.)

## Conventional Commits


Sources/Core/AWSClientRuntime/Auth/AuthSchemes/SigV4AAuthScheme.swift
```diff
@@ -0,0 +1,68 @@
...
+        // Set signedBodyHeader flag
+        let useSignedBodyHeader = ["S3", "Glacier"].contains(context.getServiceName()) && !unsignedBody
+        updatedSigningProperties.set(
+            key: AttributeKeys.signedBodyHeader,
+            value: useSignedBodyHeader ? .contentSha256 : AWSSignedBodyHeader.none
+        )
+
+        // Flags in SigningFlags object
+        // Set useDoubleURIEncode to false IFF service is S3
+        updatedSigningProperties.set(key: AttributeKeys.useDoubleURIEncode, value: serviceName != "S3")
```
dayaffe, line 61: We should pull out this boolean logic outside of the value statement to make it more readable and reduce redundancy as it is also done in shouldNormalizeURIPath

Sources/Core/AWSClientRuntime/Auth/AuthSchemes/SigV4AuthScheme.swift
```diff
@@ -16,34 +16,53 @@ public struct SigV4AuthScheme: ClientRuntime.AuthScheme {
...
-        updatedSigningProperties.set(key: AttributeKeys.signatureType, value: .requestHeaders)
+        // Expiration is only used for presigning (presign request flow or presign URL flow).
+        updatedSigningProperties.set(key: AttributeKeys.expiration, value: context.getExpiration())
 
-        // SigningFlags
+        // AWSSignatureType.requestQueryParams is only used for presign URL flow.
+        // Out of the AWSSignatureType enum cases, only two are used. .requestHeaders and .requestQueryParams.
+        // .requestHeaders is the deafult signing used for AWS operations.
         let serviceName = context.getServiceName()
+        let isS3PresignURLOp = serviceName == "S3" && ["getObject", "putObject"].contains(context.getOperation())
```
dayaffe, line 36: Since this same line is reused we should have AuthScheme Helper function file or AuthScheme constants so that we minimize duplicated code

Sources/Core/AWSClientRuntime/EventStream/AWSMessageSigner.swift
```diff
@@ -62,10 +65,22 @@ extension AWSEventStream {
...
                                                                    previousSignature: previousSignature,
                                                                    signingConfig: signingConfig)
             return signingResult.output
         }
+
+        private func getSigV4Signer() async throws -> AWSSigV4Signer {
+            // Fetch resolved signer from selected auth scheme in middleware context
+            let signer = try await self.signer()
+            // For now, only AWSSigV4Signer is used for eventstream message signing
+            let sigv4signer = signer as? AWSClientRuntime.AWSSigV4Signer
```
dayaffe, line 79: let's refactor this code now assuming that another signer might be added in the future. Could do a resolveMessageSigner() function that returns SigV4 for example

Sources/Core/AWSClientRuntime/HttpContextBuilder+Extension.swift
```diff
@@ -56,6 +52,16 @@ extension HttpContext {
...
         let messageSigner = AWSClientRuntime.AWSEventStream.AWSMessageSigner(encoder: messageEncoder) {
+            guard let authScheme = self.getSelectedAuthScheme() else {
+                throw ClientError.authError(
+                    "Signer for event stream could not be loaded because auth scheme was not configured."
+                )
+            }
+            guard let signer = authScheme.signer else {
+                throw ClientError.authError("Signer was not configured for the selected auth scheme.")
+            }
+            return signer
```
dayaffe, line 63: this should all be in a helper function as the return statement makes this function confusing without carefully inspecting where inner scope begins and ends + consider ways to make error messages more direct as it is currently difficult to distinguish the errors / would be difficult to read them out of context

Sources/Core/AWSClientRuntime/HttpContextBuilder+Extension.swift
```diff
@@ -56,6 +52,16 @@ extension HttpContext {
...
+        } signingConfig: {
             try await self.makeEventStreamSigningConfig()
         } requestSignature: {
             self.getRequestSignature()
```
dayaffe, line 67: this should be as follows

```
signingConfig: { something },
requestSignature: { something },
```

since these are short they should be scoped and implemented all on one line with a comma after the closing brackets and then onto the next

Sources/Core/AWSClientRuntime/HttpContextBuilder+Extension.swift
```diff
@@ -56,6 +52,16 @@ extension HttpContext {
         // setup client to server
         let messageEncoder = AWSClientRuntime.AWSEventStream.AWSMessageEncoder()
         let messageSigner = AWSClientRuntime.AWSEventStream.AWSMessageSigner(encoder: messageEncoder) {
```
dayaffe, line 54: have this equal the result of `return signer` and then move this to the very end of the function. The thing being returned should be last to make it clear what is being returned and of what type.

So the end of this whole function should have something like:

```
return try AWSClientRuntime.AWSEventStream.AWSMessageSigner(
        encoder: encoder,
        getSigner: { signer },
        signingConfig: { try await self.makeEventStreamSigningConfig() },
        requestSignature: { self.getRequestSignature() }
    )
}
```

Apply to any other relevant areas of this PR

dayaffe (changes requested):

Sources/Core/AWSClientRuntime/HttpContextBuilder+Extension.swift
sichanyoo: This function doesn't return anything, just initializes things (`messageSigner` and `messageEncoder`) and saves them to middleware context for them to be used by operation stack later.

Sources/Core/AWSClientRuntime/HttpContextBuilder+Extension.swift
sichanyoo: Changed formatting to be easier to read, following given suggestion!

Sources/Core/AWSClientRuntime/HttpContextBuilder+Extension.swift
sichanyoo: Changed structure as recommended.

Sources/Core/AWSClientRuntime/Auth/AuthSchemes/SigV4AAuthScheme.swift
sichanyoo: Changed as suggested.

Sources/Core/AWSClientRuntime/Auth/AuthSchemes/SigV4AAuthScheme.swift
```diff
@@ -0,0 +1,68 @@
...
+
+        // Expiration is only used for presigning (presign request flow or presign URL flow).
+        updatedSigningProperties.set(key: AttributeKeys.expiration, value: context.getExpiration())
+
+        // AWSSignatureType.requestQueryParams is only used for presign URL flow.
+        // Out of the AWSSignatureType enum cases, only two are used. .requestHeaders and .requestQueryParams.
+        // .requestHeaders is the deafult signing used for AWS operations.
+        let serviceName = context.getServiceName()
+        let isS3PresignURLOp = serviceName == "S3" && ["getObject", "putObject"].contains(context.getOperation())
+        let isPollyPresignURLOp = serviceName == "Polly" && ["synthesizeSpeech"].contains(context.getOperation())
```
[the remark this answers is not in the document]
sichanyoo: Factored out logic into a helper function that will serve as the "source of truth" that can be edited as new operations support presigning URL.

Sources/Core/AWSClientRuntime/Auth/AuthSchemes/SigV4AuthScheme.swift
sichanyoo: Factored out literals into a helper struct to reflect suggestion.

Sources/Core/AWSClientRuntime/Auth/AuthSchemes/SigV4AAuthScheme.swift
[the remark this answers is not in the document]
jbelkins: I would go further than to separate this logic out to a separate type.  We already have logic on the Kotlin side that does this when the presign methods are generated; let's set a flag or enum or something of that sort during codegen that removes the need for this logic in Swift entirely.

jbelkins (changes requested):
Agreed with David's comments, plus see my suggestion on presign customization logic.

Sources/Core/AWSClientRuntime/Auth/AuthSchemes/SigV4AAuthScheme.swift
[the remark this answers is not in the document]
sichanyoo: After thinking through things more, I think I know of a perfect place to move this flag-setting logic to. There're custom mdidlewares just for presign URL operations, there I can set some sort of flag in middleware context, which can be used in auth scheme file to determine whether it's presign op or not. Will refactor now.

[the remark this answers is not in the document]
sichanyoo: Oh wait... the `.PRESIGN_URL` of enum type `FlowType` that I added as part of refactor IS the flag that's set in Kotlin side and can be used to determine whether it's presign flow or not. LOL There's just no need for checking service and op name pairs anyways, as it's duplicate logic here. Will update accordingly.

Sources/Core/AWSClientRuntime/EventStream/AWSMessageSigner.swift
sichanyoo: Refactored signer structure to accept any custom signer for eventstream message signing.

codegen/smithy-aws-swift-codegen/src/test/kotlin/software/amazon/smithy/aws/swift/codegen/customizations/RulesBasedAuthSchemeResolverGeneratorTests.kt
```diff
@@ -0,0 +1,151 @@
...
+                }
+
+                public func constructParameters(context: HttpContext) throws -> ClientRuntime.AuthSchemeResolverParameters {
+                    return try DefaultS3AuthSchemeResolver().constructParameters(context: context)
+                }
+            }
+
+            public struct DefaultS3AuthSchemeResolver: S3AuthSchemeResolver {
+                public func resolveAuthScheme(params: ClientRuntime.AuthSchemeResolverParameters) throws -> [AuthOption] {
+                    var validAuthOptions = Array<AuthOption>()
```
dayaffe, line 98: nit should use `[AuthOption]()`

codegen/smithy-aws-swift-codegen/src/test/kotlin/software/amazon/smithy/aws/swift/codegen/customizations/RulesBasedAuthSchemeResolverGeneratorTests.kt
```diff
@@ -0,0 +1,151 @@
...
+                        switch scheme {
+                            case .sigV4(let param):
+                                var sigV4Option = AuthOption(schemeID: "aws.auth#sigv4")
+                                sigV4Option.signingProperties.set(key: AttributeKeys.signingName, value: param.signingName)
+                                sigV4Option.signingProperties.set(key: AttributeKeys.signingRegion, value: param.signingRegion)
+                                validAuthOptions.append(sigV4Option)
+                            case .sigV4A(let param):
+                                var sigV4Option = AuthOption(schemeID: "aws.auth#sigv4a")
+                                sigV4Option.signingProperties.set(key: AttributeKeys.signingName, value: param.signingName)
+                                sigV4Option.signingProperties.set(key: AttributeKeys.signingRegion, value: param.signingRegionSet?[0])
```
dayaffe, line 118: is signingRegion required? what will happen if param.signingRegionSet is null or first item is null

codegen/smithy-aws-swift-codegen/src/test/kotlin/software/amazon/smithy/aws/swift/codegen/customizations/RulesBasedAuthSchemeResolverGeneratorTests.kt
```diff
@@ -0,0 +1,151 @@
...
+                                validAuthOptions.append(sigV4Option)
+                            default:
+                                throw ClientError.authError("Unknown auth scheme name: \(scheme.name)")
+                        }
+                    }
+                    return validAuthOptions
+                }
```
dayaffe, line 125: Nit: I would really prefer if this and similarly large/complex functions were broken up into helper functions but I know its difficult since we're code generating it. At the moment it is quite hard to read.

codegen/smithy-aws-swift-codegen/src/test/kotlin/software/amazon/smithy/aws/swift/codegen/customizations/RulesBasedAuthSchemeResolverGeneratorTests.kt
```diff
@@ -0,0 +1,151 @@
...
+                    let endpointParams = EndpointParams(authSchemeParams: serviceParams)
+                    let endpoint = try DefaultEndpointResolver().resolve(params: endpointParams)
+                    guard let authSchemes = endpoint.authSchemes() else {
+                        return try InternalModeledS3AuthSchemeResolver().resolveAuthScheme(params: params)
+                    }
+                    let schemes = try authSchemes.map { (input) -> AWSClientRuntime.AuthScheme in try AWSClientRuntime.AuthScheme(from: input) }
+                    for scheme in schemes {
+                        switch scheme {
+                            case .sigV4(let param):
+                                var sigV4Option = AuthOption(schemeID: "aws.auth#sigv4")
```
dayaffe, line 111: nit can we pull out hardcoded strings and move them to the top of the class

Sources/Core/AWSClientRuntime/Auth/AuthSchemes/SigV4Util.swift
```diff
@@ -0,0 +1,26 @@
...
+
+public class SigV4Util {
+    static let unsignedBodyHeader = ["S3", "Glacier"]
+    static let forceUnsignedBodyForPresigningURL = [
+        "S3": ["getObject", "putObject"]
+    ]
+
+    static func shouldForceUnsignedBody(flow: FlowType, serviceName: String, opName: String) -> Bool {
+        let serviceQualifies = forceUnsignedBodyForPresigningURL.keys.contains(serviceName)
+        let flowQualies = flow == .PRESIGN_URL
```
jbelkins, line 19: spelling

codegen/smithy-aws-swift-codegen/src/test/kotlin/software/amazon/smithy/aws/swift/codegen/customizations/RulesBasedAuthSchemeResolverGeneratorTests.kt
```diff
@@ -0,0 +1,151 @@
...
+                        default:
+                            var sigV4Option = AuthOption(schemeID: "aws.auth#sigv4")
+                            sigV4Option.signingProperties.set(key: AttributeKeys.signingName, value: "weather")
+                            guard let region = serviceParams.region else {
+                                throw ClientError.authError("Missing region in auth scheme parameters for SigV4 auth scheme.")
+                            }
+                            sigV4Option.signingProperties.set(key: AttributeKeys.signingRegion, value: region)
+                            validAuthOptions.append(sigV4Option)
+                    }
+                    return validAuthOptions
```
dayaffe, line 88: You can combine regular and Optional case statements into one in order to reduce code repetition. For example let's take "onlyHttpApiKeyAuth" and "onlyHttpApiKeyAuthOptional" which both add httpApiKeyAuth you can do the following:
```
case "onlyHttpApiKeyAuth", "onlyHttpApiKeyAuthOptional":
    validAuthOptions.append(AuthOption(schemeID: Constants.httpApiKeyAuth))
    if serviceParams.operation.hasSuffix("Optional") { addNoAuth }
```

Here you can add the code snippet for addNoAuth or optionally you could consider making it a small function like so: `let addNoAuth = { validAuthOptions.append(AuthOption(schemeID: Constants.NoAuth)) }` and call `addNoAuth()`

jbelkins (approved):
Just a spelling issue

codegen/smithy-aws-swift-codegen/src/test/kotlin/software/amazon/smithy/aws/swift/codegen/customizations/RulesBasedAuthSchemeResolverGeneratorTests.kt
sichanyoo: If `param.signingRegionSet` is null or first item is null, then that would mean the service team failed to model their endpoint ruleset correctly. Signing region is required for SigV4.

codegen/smithy-aws-swift-codegen/src/test/kotlin/software/amazon/smithy/aws/swift/codegen/customizations/RulesBasedAuthSchemeResolverGeneratorTests.kt
sichanyoo: Changed as suggested, not just for rules based, but for model based auth resolver generation as well.

Sources/Core/AWSClientRuntime/Auth/AuthSchemes/SigV4AuthScheme.swift
```diff
@@ -16,34 +16,63 @@ public struct SigV4AuthScheme: ClientRuntime.AuthScheme {
...
+        let serviceIsS3 = serviceName == "S3"
+        // Set flags in SigningFlags object
         // Set useDoubleURIEncode to false IFF service is S3
-        updatedSigningProperties.set(key: AttributeKeys.useDoubleURIEncode, value: serviceName != "S3")
+        updatedSigningProperties.set(key: AttributeKeys.useDoubleURIEncode, value: !serviceIsS3)
         // Set shouldNormalizeURIPath to false IFF service is S3
-        updatedSigningProperties.set(key: AttributeKeys.shouldNormalizeURIPath, value: serviceName != "S3")
-        // FIXME: Flag doesn't seem to be used by anything - investigate
+        updatedSigningProperties.set(key: AttributeKeys.shouldNormalizeURIPath, value: !serviceIsS3)
         updatedSigningProperties.set(key: AttributeKeys.omitSessionToken, value: false)
```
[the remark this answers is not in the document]
sichanyoo: Moved the logic to `SigV4Util` like some others. Changed as suggested.

Sources/Core/AWSClientRuntime/Auth/AuthSchemes/SigV4AAuthScheme.swift
```diff
@@ -0,0 +1,78 @@
...
+            value: useSignedBodyHeader ? .contentSha256 : AWSSignedBodyHeader.none
+        )
+
+        let serviceIsS3 = serviceName == "S3"
+        // Set flags in SigningFlags object
+        // Set useDoubleURIEncode to false IFF service is S3
+        updatedSigningProperties.set(key: AttributeKeys.useDoubleURIEncode, value: !serviceIsS3)
+        // Set shouldNormalizeURIPath to false IFF service is S3
+        updatedSigningProperties.set(key: AttributeKeys.shouldNormalizeURIPath, value: !serviceIsS3)
+        updatedSigningProperties.set(key: AttributeKeys.omitSessionToken, value: false)
```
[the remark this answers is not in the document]
sichanyoo: Changed as suggested.

Sources/Core/AWSClientRuntime/Auth/AuthSchemes/SigV4AuthScheme.swift
```diff
@@ -16,34 +16,63 @@ public struct SigV4AuthScheme: ClientRuntime.AuthScheme {
...
         let serviceName = context.getServiceName()
+        let isPresignURLFlow = context.getFlowType() == .PRESIGN_URL
+        updatedSigningProperties.set(
+            key: AttributeKeys.signatureType,
+            value: isPresignURLFlow ? .requestQueryParams : .requestHeaders
+        )
+
+        // Set unsignedBody flag
+        // Operation name is guaranteed to be in middleware context from generic codegen.
+        let operationName = context.getOperation()!
```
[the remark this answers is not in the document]
sichanyoo: Changed as suggested.

codegen/smithy-aws-swift-codegen/src/test/kotlin/software/amazon/smithy/aws/swift/codegen/customizations/RulesBasedAuthSchemeResolverGeneratorTests.kt
sichanyoo: In this case I think it's better to have the codegen code be easier to understand than the generated code be easier to understand, since what we're maintaining is the codegen code. To achieve good readability in generated output, the codegen code will be harder to understand as it creates multiple methods that work together in the end result.

Sources/Core/AWSClientRuntime/Auth/AuthSchemes/SigV4Util.swift
sichanyoo: Fixed as pointed out.

codegen/smithy-aws-swift-codegen/src/test/kotlin/software/amazon/smithy/aws/swift/codegen/customizations/RulesBasedAuthSchemeResolverGeneratorTests.kt
sichanyoo: In codegen, the strings come from shape ID of SigV4Trait object. I think it's fine to have hardcoded string in generated code.

codegen/smithy-aws-swift-codegen/src/test/kotlin/software/amazon/smithy/aws/swift/codegen/customizations/RulesBasedAuthSchemeResolverGeneratorTests.kt
dayaffe: in that case should we catch if its null and throw an error accordingly? right now it would be passed in even if null

codegen/smithy-aws-swift-codegen/src/test/kotlin/software/amazon/smithy/aws/swift/codegen/customizations/RulesBasedAuthSchemeResolverGeneratorTests.kt
dayaffe: It doesnt have to be a tradeoff between good codegen code and good generated code we can achieve both! But I don't want to block your PR on this and since Josh doesn't have any issues here Ill leave it be

Issue #763, opened 2022-12-19
SQS change from AWSQuery to AWSJSON


smivij:
Details of the issue is in SIM...
*[SQS change from AWS Query to AWS JSON](https://issues.amazon.com/issues/SDK-82857)* 

SQS created a https://quip-amazon.com/qBbbArTLikRM/Future-proofing-AwsQueryCompatible-supportto support AwsQueryCompatible support in non-GA SDKs, with a call out that while backwards compatibility is not required for their service, any other services that will eventually migrate to AWS JSON would need this trait. SDK team reviewed and agreed. We will add this to the GA checklist for non-GA SDKs by 12/23

smivij:
The expectation is that we only need to support AWSJson for new SDKs and if we already support AWSQuery we need to change that

jbelkins:
Released in AWS SDK for Swift 0.32.0

PR #1233, opened 2023-11-22
feat: update SQS to latest model that uses aws json protocol + add integration test

dayaffe:
Update models for SQS and add an integration test

## Issue \#
#763 

## Description of changes
- Update SQS models which now use AWS Json instead of AWS Query. The rest of the model update will occur during the next release.
- Add an integration test to ensure that it works.

## Conventional Commits


jbelkins (approved):
Approved.  Optionally, you might want to just collapse `setUp()` and `tearDown()` into the test since there is only one test.

dayaffe:
> Approved. Optionally, you might want to just collapse `setUp()` and `tearDown()` into the test since there is only one test.

Going to leave setUp and tearDown in case we want to add additional SQS integration tests in the future

Issue #1113, opened 2023-09-05
S3 throws errors after HTTP/200


jbelkins:
### Describe the feature

Internal tracking: `SDK-153398`

### Use Case

Required for GA

### Acknowledgements

- [ ] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

dayaffe:
Discussion related to issue: https://amzn-aws.slack.com/archives/G017UPVDRAS/p1693326099352059

jbelkins:
Shipped in AWS SDK for Swift 0.34.0

PR #1266, opened 2023-12-19
feat: handle errors in 200 response from S3

dayaffe:
## Issue \#
#1113 

## Description of changes
- Add a SwiftIntegration/Middleware which handles parsing <Error> in 200 response for S3 service calls
- Code-generates the following middleware into all methods without a stream in the response output structure:
```
operation.deserializeStep.intercept(position: .after, middleware: AWSClientRuntime.AWSS3ErrorWith200StatusXMLMiddleware<OutputStructType>())
```
- Add unit test that forces <Error> tag in 200 response and checks for 500 error thrown with message

## Conventional Commits


jbelkins (changes requested):

jbelkins (changes requested):
The middleware block that is repeated for every single S3 operation should be extracted to a reusable middleware, then just inserted into code generation.

That block has no dependency on anything in S3Client, so there should be no problem having the reusable middleware live in `AWSClientRuntime`.

codegen/smithy-aws-swift-codegen/src/main/kotlin/software/amazon/smithy/aws/swift/codegen/customization/s3/S3ErrorWith200StatusIntegration.kt
```diff
@@ -0,0 +1,68 @@
...
+    override fun customizeMiddleware(
+        ctx: ProtocolGenerator.GenerationContext,
+        operationShape: OperationShape,
+        operationMiddleware: OperationMiddleware,
+    ) {
+        // we don't know for sure what operations S3 does this on. Go customized this for only a select few
+        // like CopyObject/UploadPartCopy/CompleteMultipartUpload but Rust hit it on additional operations
+        // (DeleteObjects).
+        // Instead of playing whack-a-mole broadly apply this interceptor to everything but streaming responses
+        // which adds a small amount of overhead to response processing.
```
jbelkins, line 42: Agreed with this.  I don't think it will cause any problems where it is not needed.

Sources/Core/AWSClientRuntime/Middlewares/AWSS3ErrorWith200StatusXMLMiddleware.swift
```diff
@@ -0,0 +1,50 @@
...
+            return response
+        }
+
+        // Check if the response body contains an XML Error
+        guard let data = try await response.httpResponse.body.readData() else {
+            return response
+        }
+
+        let xmlString = String(data: data, encoding: .utf8) ?? ""
+        if xmlString.contains("<Error>") {
```
jbelkins, line 37: I'd prefer not working with XML data directly.

This can stay for now, but as part of XML project I'll look at how we can do this within existing XML handling.

jbelkins (approved):

Issue #1102, opened 2023-09-05
Region Search Precedence


jbelkins:
### Describe the feature

Quip: `nY63AaBMJcoU/Cross-SDK-Feature-Matrix-Items-Deprecated-or-Without-SEPs#temp:C:UcSa0a343b6b2034712914a2965f`

### Use Case

Is a GA requirement

### Acknowledgements

- [ ] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

jbelkins:
Shipped in AWS SDK for Swift 0.34.0

PR #1290, opened 2024-01-09
fix: Delete bundle region provider.

sichanyoo:
## Issue \#

https://github.com/awslabs/aws-sdk-swift/issues/1102

## Description of changes

Deletes unnecessary region provider that fetches region from Bundle for iOS / tvOS / watchOS applications.

N/A

## Conventional Commits


dayaffe:
Happy to approve but can you provide more details on why its unnecessary? Was it unused?

jbelkins:
@dayaffe We asked both Amplify and our tester community at-large, no one seems to be using this

Sources/Core/AWSClientRuntime/Regions/DefaultRegionResolver.swift
```diff
@@ -13,7 +13,6 @@ public struct DefaultRegionResolver: RegionResolver {
 
     public init(fileBasedConfigurationProvider: @escaping FileBasedConfigurationProviding) throws {
         self.providers = [
-            BundleRegionProvider(),
             EnvironmentRegionProvider(),
             ProfileRegionProvider(fileBasedConfigurationProvider: fileBasedConfigurationProvider),
             try IMDSRegionProvider()
```
jbelkins, line 18: After removing `BundleRegionProvider`, are the rest of these region providers complete & in the right order?

sichanyoo: Yes. 
`nY63AaBMJcoU/Cross-SDK-Feature-Matrix-Items-Deprecated-or-Without-SEPs#temp:C:UcSa0a343b6b2034712914a2965f`

sichanyoo:
> Happy to approve but can you provide more details on why its unnecessary -- was it unused?

@dayaffe It's not part of the standard chain of places to resolve region from (environment, profile, then IMDS), and it's not clear whether it's even used. Checked with Amplify and they also don't specify region from application Bundle. Checked with internal slack channel and no reply from anyone. Josh also said that based on his experience, it would be unlikely that you'd want to put region info into the application Bundle.

dayaffe (approved):

PR #1294, opened 2024-01-10
chore: add smoke test traits dep

milesziemer:
## Issue \#

Adds a dependency on smithy-smoke-test-traits in the sdk code gen package so aws service models that use the smokeTests trait won't fail to build because of unknown traits. Generated a client with a smoke test and confirmed there was no diff.

Adding the dependency in the code gen package since it's the models that need the dependency, not the code generator. When smoke tests are implemented in the generator, we can add the dep there.

## Conventional Commits


dayaffe (approved):

codegen/sdk-codegen/build.gradle.kts
```diff
@@ -16,8 +16,10 @@ plugins {
     id("software.amazon.smithy") version "0.5.3"
 }
 
+val smithyVersion: String by project
 dependencies {
     implementation(project(":smithy-aws-swift-codegen"))
+    implementation("software.amazon.smithy:smithy-smoke-test-traits:$smithyVersion")
```
jbelkins, line 22: @milesziemer Will this dependency fail our build if we don't update to the Smithy version that includes it at the same time?

(We pin to specific Smithy versions in this project and in smithy-swift)

milesziemer: It looks like you're on 1.42.0 based on [gradle.properties](https://github.com/smithy-lang/smithy-swift/blob/922d0663ff94bf4895088f13c6d3afcd91c703fa/gradle.properties#L6), right? 1.42.0 has smithy-smoke-test-traits

jbelkins (approved):

jbelkins:
Merging.  Integration test failures are due to this PR being opened from a fork.

PR #1297, opened 2024-01-11
Update .spi.yml

finestructure:
Update `.spi.yml` to fix some issues (more detail inline)

I noticed that the builds are all failing in our compatibility test suite and I believe this can be fixed with a couple of tweaks to the SPI config file.

There's also a real build failure on Linux:

```
/host/spi-builder-workspace/.build/checkouts/aws-crt-swift/aws-common-runtime/aws-c-cal/include/aws/cal/private/opensslcrypto_common.h:10:10: fatal error: 'openssl/crypto.h' file not found
#include <openssl/crypto.h>
         ^~~~~~~~~~~~~~~~~~
```

because we didn't yet install `libssl-dev` in our images. I've fixed this and Linux builds now pass:

CleanShot 2024-01-11 at 11 51 01@2x

(at least for 0.33.0, there's a [genuine build failure on `main`](https://swiftpackageindex.com/builds/3E63E721-AA2F-423D-8C35-196D214DCC74):

```
error: Dependencies could not be resolved because root depends on 'aws-crt-swift' 0.20.0.
'aws-crt-swift' 0.17.0 is required because 'smithy-swift' 0.37.0 depends on 'aws-crt-swift' 0.17.0 and root depends on 'smithy-swift' 0.37.0.
BUILD FAILURE 5.9 linux
```
)

### DIFF
--- a/.spi.yml
+++ b/.spi.yml
@@ -1,20 +1,5 @@
 version: 1
 builder:
   configs:
-  - platform: macosXcodebuild
-    target: AWSSTS
-  - platform: ios
-    target: AWSSTS
-  - platform: linux
-    swift_version: '5.7'
-    image: registry.gitlab.com/finestructure/spi-images:basic-5.7-latest
-    target: AWSSTS
-  - platform: linux
-    swift_version: '5.8'
-    image: registry.gitlab.com/finestructure/spi-images:basic-5.8-latest
-    target: AWSSTS
-  - platform: linux
-    swift_version: '5.9'
-    image: registry.gitlab.com/finestructure/spi-images:basic-5.9-latest
-    target: AWSSTS
-
+  - target: AWSSTS
+    scheme: AWSSTS


.spi.yml
```diff
@@ -1,20 +1,5 @@
 version: 1
 builder:
   configs:
-  - platform: macosXcodebuild
-    target: AWSSTS
```
finestructure, line 5: `macosXcodebuild` doesn't support the `target` flag, it needs `scheme` instead. (`--target` is a `swift build` parameter, `--scheme` an `xcodebuild` parameter. We should perhaps have collapsed them into one but it wasn't clear at the time if they'd be used synonymously.)

.spi.yml
```diff
@@ -1,20 +1,5 @@
...
-    swift_version: '5.8'
-    image: registry.gitlab.com/finestructure/spi-images:basic-5.8-latest
-    target: AWSSTS
-  - platform: linux
-    swift_version: '5.9'
-    image: registry.gitlab.com/finestructure/spi-images:basic-5.9-latest
-    target: AWSSTS
-
+  - target: AWSSTS
+    scheme: AWSSTS
```
finestructure, line 5: Overall, this can just be condensed to setting both `target` and `scheme` for _all_ platform/Swift version combos because they'll be ignored if they don't apply.

dayaffe:
Thank you for your PR! In addition to reviewing this I will look into the build failure you pointed out that is effecting the rest of the platforms.

dayaffe (approved):

dayaffe:
@finestructure main is failing because aws-sdk-swift pulls the latest released version of smithy-swift which is still on 0.17. This issue will be resolved with our next release.

jbelkins (approved):

jbelkins:
@finestructure We merged this `.spi.yml` change, plus we fixed the issue that was causing `maIn` branch to fail to build.

We'll re-engage with you if there are still build failures after these changes get rebuilt by SPI.

finestructure:
Thanks @jbelkins @dayaffe , that greatly improved things!

CleanShot 2024-01-12 at 09 30 42@2x

CleanShot 2024-01-12 at 09 32 26@2x

jbelkins:
@finestructure Thanks again for your help with our SPI config.

Once we publish our next release, we should show all green across the board ✅

finestructure:
That's great, happy to help!

PR #1312, opened 2024-01-22
feat: add default client configurations

AndrewFossAWS:
## Description of changes
* See https://github.com/smithy-lang/smithy-swift/pull/650 for context on client configurations
* Added AWS specific default client configurations
* `AWSClientConfigDefaultsProvider.swift` provides all the default values for client config properties
* Code generate AWS clients configurations (align with Smithy SRA runtime-plugins design)

## Conventional Commits


Sources/Core/AWSClientRuntime/Config/AwsDefaultClientConfiguration.swift
```diff
@@ -0,0 +1,36 @@
...
+// All Rights Reserved.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+public protocol AwsDefaultClientConfiguration {
+    /// The credentials provider to be used for AWS credentials.
+    ///
+    /// If no credentials provider is supplied, the SDK will look for credentials in the environment, then in the `~/.aws/credentials` file.
+    var credentialsProvider: CredentialsProviding { get set }
```
jbelkins, line 12: Isn't credentials provider a non-AWS Smithy construct, and if so, shouldn't this go in the Smithy `DefaultClientConfiguration`?

Sources/Services/AWSACM/ACMClient.swift
```diff
@@ -51,6 +51,35 @@ extension ACMClient {
...
+
+    static func resolve(plugins: [any Plugin]) async throws -> ACMClient.ACMClientConfiguration {
+        let clientConfiguration = try await ACMClient.ACMClientConfiguration()
+        for plugin in plugins {
+            try await plugin.configureClient(clientConfiguration: clientConfiguration)
+        }
+        return clientConfiguration
+    }
+
+    public class Builder {
```
jbelkins, line 63: Wondering if this Builder can be generic, rather than code-generate it 375 times over.  Looks like every client just needs to have a `Client(config:)` initializer

Sources/Core/AWSClientRuntime/Config/AwsDefaultClientConfiguration.swift
```diff
@@ -0,0 +1,36 @@
+//
+// Copyright Amazon.com Inc. or its affiliates.
+// All Rights Reserved.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+public protocol AwsDefaultClientConfiguration {
```
jbelkins, line 8: nit: Swift convention is to leave acronyms all-caps in CamelCase.

So, `AWSDefaultClientConfiguration` not `AwsDefaultClientConfiguration`

jbelkins (dismissed):

Sources/Core/AWSClientRuntime/Config/AwsDefaultClientConfiguration.swift
AndrewFossAWS: The `CredentialsProviding` protocol (and all the credentials related classes) currently resides in aws-sdk-swift AWSClientRuntime. 
We can potentially move it into smithy-swift's ClientRuntime and then add it to `DefaultClientConfiguration` in a seperate PR

Sources/Services/AWSACM/ACMClient.swift
AndrewFossAWS: Created `ClientBuilder` class in `smithy-swift` ClientRuntime

Sources/Services/AWSACM/ACMClient.swift
```diff
@@ -26,30 +26,96 @@ public class ACMClient {
...
+            self.logger = logger
+            self.retryStrategyOptions = retryStrategyOptions
+            self.clientLogMode = clientLogMode
+            self.endpoint = endpoint
+            self.idempotencyTokenGenerator = idempotencyTokenGenerator
+            self.httpClientEngine = httpClientEngine
+            self.httpClientConfiguration = httpClientConfiguration
+        }
+
+        public convenience init(useFIPS: Bool? = nil, useDualStack: Bool? = nil, appID: String? = nil, credentialsProvider: CredentialsProviding? = nil, awsRetryMode: AWSRetryMode? = nil, region: String? = nil, signingRegion: String? = nil, endpointResolver: EndpointResolver? = nil, encoder: RequestEncoder? = nil, decoder: ResponseDecoder? = nil, logger: LogAgent? = nil, retryStrategyOptions: RetryStrategyOptions? = nil, clientLogMode: ClientLogMode? = nil, endpoint: String? = nil, idempotencyTokenGenerator: IdempotencyTokenGenerator? = nil, httpClientEngine: HTTPClient? = nil, httpClientConfiguration: HttpClientConfiguration? = nil) async throws {
```
AndrewFossAWS, line 96: We can potentially reduce the large number of parameters in the constructor if some of them are optional. I have included all of them for now. (There are also two convenience inits: `init()` and `init(region)`)

IntegrationTests/Services/AWSKinesisIntegrationTests/KinesisTests.swift
```diff
@@ -22,7 +22,7 @@ class KinesisTests: XCTestCase {
         // Resources will be cleaned up before the test concludes, pass or fail, unless the test crashes.
 
         let streamName = UUID().uuidString
-        let client = try KinesisClient(region: "us-west-2")
+        let client = try await KinesisClient(region: "us-west-2")
```
syall, line 25: Clarifying question: is the constructor become async like this the norm with these changes?

IntegrationTests/Services/AWSSQSIntegrationTests/SQSTests.swift
```diff
@@ -17,7 +17,7 @@ class SQSTests: XCTestCase {
     private var queueUrl: String?
 
     override func setUp() async throws {
-        client = try await SQSClient()
+        client = try await SQSClient.builder().build()
```
syall, line 20: Is the noargs constructor no longer available?

Sources/Core/AWSClientRuntime/UserAgent/AWSUserAgentMetadata.swift
```diff
@@ -73,10 +73,10 @@ public struct AWSUserAgentMetadata {
         self.frameworkMetadata = frameworkMetadata
     }
 
-    public static func fromConfig<ServiceSpecificConfiguration>(
+    public static func fromConfig(
         serviceID: String,
         version: String,
-        config: AWSClientConfiguration<ServiceSpecificConfiguration>
+        config: DefaultClientConfiguration & AWSDefaultClientConfiguration
```
syall, line 79: This narrows down service specific configurations to default configurations (which I think is good).

But is there a use case where we might need service specific configuration?

Sources/Services/AWSACM/ACMClient.swift
```diff
@@ -26,30 +26,96 @@ public class ACMClient {
...
+
+        public var endpoint: Swift.String?
+
+        public var idempotencyTokenGenerator: ClientRuntime.IdempotencyTokenGenerator
+
+        public var httpClientEngine: ClientRuntime.HTTPClient
+
+        public var httpClientConfiguration: ClientRuntime.HttpClientConfiguration
+
+        private init(_ useFIPS: Swift.Bool?, _ useDualStack: Swift.Bool?, _ appID: Swift.String?, _ credentialsProvider: AWSClientRuntime.CredentialsProviding, _ awsRetryMode: AWSClientRuntime.AWSRetryMode, _ region: Swift.String?, _ signingRegion: Swift.String?, _ endpointResolver: EndpointResolver, _ encoder: ClientRuntime.RequestEncoder?, _ decoder: ClientRuntime.ResponseDecoder?, _ logger: ClientRuntime.LogAgent, _ retryStrategyOptions: ClientRuntime.RetryStrategyOptions, _ clientLogMode: ClientRuntime.ClientLogMode, _ endpoint: Swift.String?, _ idempotencyTokenGenerator: ClientRuntime.IdempotencyTokenGenerator, _ httpClientEngine: ClientRuntime.HTTPClient, _ httpClientConfiguration: ClientRuntime.HttpClientConfiguration) {
```
syall, line 76: nit: is there a way to format this better, or is this convention?

I saw the other comment on the constructor, but don't have an alternative.

syall (changes requested):
One category of tests that are missing are integration tests with a custom plugin for a service (with/out the builder), and seeing if the config values after resolution are expected. Can approve after that.

IntegrationTests/Services/AWSKinesisIntegrationTests/KinesisTests.swift
AndrewFossAWS: Yes. If user doesn't provide region or other settings to the client constructor, then I think it makes sense for it to be asynchronous because it will try to fetch the settings from local config file.

IntegrationTests/Services/AWSSQSIntegrationTests/SQSTests.swift
AndrewFossAWS: It's still available. I will try updating it to the following to verify the region is updated:
```
try await SQSClient.builder().withPlugin(RegionPlugin("us-west-1")).build()
```

Sources/Core/AWSClientRuntime/UserAgent/AWSUserAgentMetadata.swift
AndrewFossAWS: I refactored the `fromConfig` function because 2) `AWSClientConfiguration ` and 2) the function only uses variables specified by the two configuration protocols. (it doesn't use any service specific stuff)

jbelkins (commented):
Please address swiftlint issues raised by CI.

To run swiftlint locally:
```
$ brew install swiftlint
```

Then, from the project root directory, on the CLI:
```
$ swiftlint
```

Sources/Services/AWSACM/ACMClient.swift
AndrewFossAWS: I can format them into seperate lines to make the `init` look prettier, though I don't think it really matters from user's perspective.

AndrewFossAWS: Actually it gets more complicated if I want to generate the following:
```
private init(
  _ useFIPS: Swift.Bool?, 
  _ useDualStack: Swift.Bool?, 
  _ appID: Swift.String?, 
  ...
) {
  // 
}
```
I have to open and join two blocks `writer.openBlock(...)`. I would prefer to keep all the constructor args in a single line since it doesn't affect the end user.

syall (approved):

IntegrationTests/Services/AWSKinesisIntegrationTests/KinesisTests.swift
```diff
@@ -22,7 +23,10 @@ class KinesisTests: XCTestCase {
         // Resources will be cleaned up before the test concludes, pass or fail, unless the test crashes.
 
         let streamName = UUID().uuidString
-        let client = try KinesisClient(region: "us-west-2")
+        let client = try await KinesisClient
+            .builder()
+            .withPlugin(RegionPlugin("us-west-2"))
+            .build()
```
jbelkins, line 29: We have to keep our client initializer non-async, at least for now.  This has caused problems with amplify-swift when we made client construction async in the past.

jbelkins: The background here is that asynchronous Swift is a relatively recent thing, and not all Swift is asynchronous.

Almost all of the Swift SDK is async at this point, but there are a couple things like client constructors that we have agreed to keep synchronous for Amplify.

jbelkins (commented):
See my comments on asynchronous constructors.

Sources/Core/AWSClientRuntime/AWSClientConfigDefaultsProvider.swift
```diff
@@ -0,0 +1,105 @@
...
+
+        // swiftlint:disable switch_case_alignment
+        switch try self.retryMode(nil) {
+            case .legacy, .standard:
+                resolvedRateLimitingMode = .standard
+            case .adaptive:
+                resolvedRateLimitingMode = .adaptive
+        }
+        // swiftlint:enable switch_case_alignment
+
```
dayaffe, line 102: nit: should remove switftlint disable switch_case_alignment and move case statements over so that they are in-line with switch

dayaffe (commented):
just one nit here I'd like addressed! I am good with the code logic

dayaffe (approved):

Issue #1189, opened 2023-10-24
Support STSWebIdentity credential provider


jbelkins:
### Describe the feature

aws-crt-swift 0.14.0 added needed parameters for the STSWebIdentity credential provider.  https://github.com/awslabs/aws-crt-swift/pull/199/files

Create a SDK wrapper for this credential provider and integration tests.

### Use Case

To support all standard AWS credential providers

### Acknowledgements

- [ ] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

smivij:
Add Integration test

sichanyoo:
Released with: 
https://github.com/awslabs/aws-sdk-swift/releases/tag/0.35.0
https://github.com/smithy-lang/smithy-swift/releases/tag/0.40.0

PR #1327, opened 2024-02-07
feat: Update STS web identity credentials provider & add integration test

sichanyoo:
## Issue \#

https://github.com/awslabs/aws-sdk-swift/issues/1189

## Description of changes

- Updates the SDK wrapper for STS web identity credentials provider following CRT changes here: https://github.com/awslabs/aws-crt-swift/pull/199/files
- Add the integration test for STS web identity credentials provider that uses STS, IAM, and Cognito.

## Integration test structure overview:
1. Create a temporary role with assume role policy that accepts unauthenticated OIDC tokens made by Cognito
2. Create a Cognito identity pool, then fetch Cognito ID from it, then fetch OIDC token from it using Cognito ID
3. Save the fetched OIDC token to a file
4. Construct STS web identity credentials provider using the role and the file path to cached token
5. Call `STS::getCallerIdentity` using a STS client that has just the STS web identity credentials provider configured
6. Confirm request succeeded by checking response fields (not nil & not empty string)
7. Clean up role policy, role, Cognito identity pool, and token file after the test

## Conventional Commits

### DIFF
--- a/IntegrationTests/Services/AWSSTSIntegrationTests/STSWebIdentityCredentialsProviderTests.swift
+++ b/IntegrationTests/Services/AWSSTSIntegrationTests/STSWebIdentityCredentialsProviderTests.swift
@@ -41,41 +41,41 @@ class STSWebIdentityCredentialsProviderTests: XCTestCase {
 
     // JSON assume role policy
     private let assumeRolePolicy = """
+    {
+      "Version": "2012-10-17",
+      "Statement": [
         {
-          "Version": "2012-10-17",
-          "Statement": [
-            {
-              "Sid": "",
-              "Effect": "Allow",
-              "Principal": {
-                "Federated": "cognito-identity.amazonaws.com"
-              },
-              "Action": [
-                "sts:AssumeRoleWithWebIdentity"
-              ],
-              "Condition": {
-                 "ForAnyValue:StringLike": {
-                   "cognito-identity.amazonaws.com:amr": "unauthenticated"
-                 }
-              }
-            }
-          ]
+          "Sid": "",
+          "Effect": "Allow",
+          "Principal": {
+            "Federated": "cognito-identity.amazonaws.com"
+          },
+          "Action": [
+            "sts:AssumeRoleWithWebIdentity"
+          ],
+          "Condition": {
+             "ForAnyValue:StringLike": {
+               "cognito-identity.amazonaws.com:amr": "unauthenticated"
+             }
+          }
         }
+      ]
+    }
     """
     // Identity policy name & JSON identity policy
     private let identityPolicyName = "allow-STS-getCallerIdentity"
     private let roleIdentityPolicy = """
-        {
-            "Version": "2012-10-17",
-            "Statement": [
-                {
-                    "Sid": "",
-                    "Effect": "Allow",
-                    "Action": "sts:GetCallerIdentity",
-                    "Resource": "*"
-                }
-            ]
-        }
+    {
+        "Version": "2012-10-17",
+        "Statement": [
+            {
+                "Sid": "",
+                "Effect": "Allow",
+                "Action": "sts:GetCallerIdentity",
+                "Resource": "*"
+            }
+        ]
+    }
     """
 
     override func setUp() async throws {
--- a/IntegrationTests/Services/AWSSTSIntegrationTests/STSWebIdentityCredentialsProviderTests.swift
+++ b/IntegrationTests/Services/AWSSTSIntegrationTests/STSWebIdentityCredentialsProviderTests.swift
@@ -127,7 +127,7 @@ class STSWebIdentityCredentialsProviderTests: XCTestCase {
             assumeRolePolicyDocument: assumeRolePolicy,
             roleName: roleName
         )).role?.arn
-        // Return after role creation finishes
+        // This wait is necessary for role creation to propagate everywhere
         let seconds = 10
         let NSEC_PER_SEC = 1_000_000_000
         try await Task.sleep(nanoseconds: UInt64(seconds * NSEC_PER_SEC))
--- a/IntegrationTests/Services/AWSSTSIntegrationTests/STSWebIdentityCredentialsProviderTests.swift
+++ b/IntegrationTests/Services/AWSSTSIntegrationTests/STSWebIdentityCredentialsProviderTests.swift
@@ -61,7 +61,7 @@ class STSWebIdentityCredentialsProviderTests: XCTestCase {
             }
           ]
         }
-        """
+    """
     // Identity policy name & JSON identity policy
     private let identityPolicyName = "allow-STS-getCallerIdentity"
     private let roleIdentityPolicy = """
@@ -76,7 +76,7 @@ class STSWebIdentityCredentialsProviderTests: XCTestCase {
                 }
             ]
         }
-        """
+    """
 
     override func setUp() async throws {
         try await super.setUp()
--- a/IntegrationTests/Services/AWSSTSIntegrationTests/STSWebIdentityCredentialsProviderTests.swift
+++ b/IntegrationTests/Services/AWSSTSIntegrationTests/STSWebIdentityCredentialsProviderTests.swift
@@ -17,8 +17,7 @@ import Foundation
 class STSWebIdentityCredentialsProviderTests: XCTestCase {
     private let region = "us-west-2"
     private var oidcToken: String!
-    // File path for the cached OIDC token.
-    private var tokenFilePath: String!
+    private var oidcTokenFilePath: String!
 
     // STS client with only the STS web identity credentials provider configured.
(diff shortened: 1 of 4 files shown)


IntegrationTests/Services/AWSSTSIntegrationTests/STSWebIdentityCredentialsProviderTests.swift
```diff
@@ -0,0 +1,206 @@
...
+        _ = try await iamClient.deleteRole(input: DeleteRoleInput(roleName: roleName))
+        // Delete Cognito identity pool
+        _ = try await cognitoIdentityClient.deleteIdentityPool(input: DeleteIdentityPoolInput(identityPoolId: identityPoolId))
+        // Delete token file
+        try deleteTokenFile()
+    }
+
+    // Helper methods for saving & deleting token file
+    private func saveTokenIntoFile() throws {
+        tokenFilePath = NSHomeDirectory() + "/token.txt"
```
dayaffe, line 162: We might want to use `FileManager.default.temporaryDirectory` instead of NSHomeDirectory()

IntegrationTests/Services/AWSSTSIntegrationTests/STSWebIdentityCredentialsProviderTests.swift
```diff
@@ -0,0 +1,206 @@
...
+        let role = try await iamClient.createRole(input: CreateRoleInput(
+            assumeRolePolicyDocument: assumeRolePolicy,
+            roleName: roleName
+        ))
+        roleArn = role.role?.arn
+
+        // This wait is necessary for role creation to propagate everywhere
+        let seconds = 10
+        let NSEC_PER_SEC = 1_000_000_000
+        try await Task.sleep(nanoseconds: UInt64(seconds * NSEC_PER_SEC))
```
dayaffe, line 95: We shouldn't introduce anything non-deterministic in our integration tests if we can help it. A better approach would be to poll IAM to see if the role was created successfully.

Example using `getRole`:
```
func isRoleReady(iamClient: IAMClient, roleName: String) async -> Bool {
    do {
        let response = try await iamClient.getRole(input: GetRoleInput(roleName: roleName))
        return response.role != nil
    } catch {
        return false
    }
}
```

Then you can setup a function that calls isRoleReady every X seconds using Task.sleep until a configured timeout (if after X seconds/minutes the role is still not there, throw an error).

dayaffe (changes requested):
Overall looks good! just a comment to use getRole instead of sleeping to make the test deterministic and some clean up

IntegrationTests/Services/AWSSTSIntegrationTests/STSWebIdentityCredentialsProviderTests.swift
sichanyoo: It definitely would be ideal to have some polling mechanism like the one you suggested instead of a fixed wait behavior, but `getRole` doesn't serve our purpose here (nor does any other IAM API). That's because changes to IAM entities become available to IAM API immediately BUT it takes time for that info to propagate everywhere. With 7 second wait it reliably passed every time on local machine, so I made it 10 just to be safe. Here's a [relevant stackoverflow discussion](https://stackoverflow.com/questions/20156043/how-long-should-i-wait-after-applying-an-aws-iam-policy-before-it-is-valid/20157294#20157294) if you'd like to read more about it.

IntegrationTests/Services/AWSSTSIntegrationTests/STSWebIdentityCredentialsProviderTests.swift
```diff
@@ -0,0 +1,206 @@
...
+import XCTest
+import AWSSTS
+import AWSIAM
+import AWSCognitoIdentity
+import ClientRuntime
+import AWSClientRuntime
+import Foundation
+
+/// Tests STS web identity credentials provider using STS::getCallerIdentity.
+class STSWebIdentityCredentialsProviderTests: XCTestCase {
```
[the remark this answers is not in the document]
sichanyoo: I'd rather keep stuff for this integration test confined to one file. I'll go ahead and refactor stuff into more helper methods and locate them all at bottom of file for better readability.

IntegrationTests/Services/AWSSTSIntegrationTests/STSWebIdentityCredentialsProviderTests.swift
sichanyoo: Changed as suggested.

IntegrationTests/Services/AWSSTSIntegrationTests/STSWebIdentityCredentialsProviderTests.swift
```diff
@@ -0,0 +1,206 @@
...
+    private func deleteTokenFile() throws {
+        do {
+            try FileManager.default.removeItem(atPath: tokenFilePath)
+        } catch {
+            throw TokenFileError.TokenFileDeletionFailed("Failed to delete token text file.")
+        }
+    }
+
+    // Confirm STS web identity credentials provider gave valid credentials
+    // ...by checking response was returned successfully.
```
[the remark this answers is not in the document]
sichanyoo: Changed as suggested.

IntegrationTests/Services/AWSSTSIntegrationTests/STSWebIdentityCredentialsProviderTests.swift
```diff
@@ -0,0 +1,227 @@
...
+        stsConfig = try await STSClient.STSClientConfiguration(
+            credentialsProvider: webIdentityCredentialsProvider,
+            region: region
+        )
+    }
+
+    /* /////////////////////////
+     * TEARDOWN HELPER FUNCTIONS
+     * /////////////////////////
+     */
```
jbelkins, line 196: Xcode has a "mark" comment directive that can be used to delineate sections of code, and the marks you create show up in the list of symbols in the dropdown above the editor.

See: https://www.avanderlee.com/xcode/xcode-mark-line-comment/

IntegrationTests/Services/AWSSTSIntegrationTests/STSWebIdentityCredentialsProviderTests.swift
```diff
@@ -0,0 +1,227 @@
...
+     * ENUMS USED IN SETUP & TEARDOWN
+     * //////////////////////////////
+     */
+
+    enum TokenFileError: Error {
+        case TokenFileCreationFailed(String)
+        case TokenFileDeletionFailed(String)
+    }
+
+    enum IdentityPoolStatus {
```
jbelkins, line 223: Swift convention is that enum cases should be lowerCamelCase.

IntegrationTests/Services/AWSSTSIntegrationTests/STSWebIdentityCredentialsProviderTests.swift
jbelkins: Also see if there is a waiter defined on the service for the resource you are waiting on.

Check the Kinesis integration test: it uses a predefined waiter and manual waiting in two stages of the test to ensure conditions are met before proceeding.

jbelkins (changes requested):

IntegrationTests/Services/AWSSTSIntegrationTests/STSWebIdentityCredentialsProviderTests.swift
sichanyoo: Changed as suggested.

IntegrationTests/Services/AWSSTSIntegrationTests/STSWebIdentityCredentialsProviderTests.swift
sichanyoo: Changed as suggested.

IntegrationTests/Services/AWSSTSIntegrationTests/STSWebIdentityCredentialsProviderTests.swift
sichanyoo: Tried the waiters and they don't serve the polling purpose in this instance for the same reasons explained above, as expected. Unless IAM provides some sort of API where it shows IAM identity's propagation status, I don't think it's possible to poll this info.

jbelkins: Okay, we can try this live in CI for a while, but we'll have to revisit this test if it causes spurious failures in practice.

With the size & complexity of our test suite, we can't afford even a small number of flaky tests.

jbelkins (approved):
Approved, but note my comment on the time delay issue.

dayaffe (approved):

PR #1345, opened 2024-02-12
feat: add integration tests for normal payloads

dayaffe:
## Conventional Commits


IntegrationTests/Services/AWSS3IntegrationTests/S3FlexibleChecksumsTests.swift
```diff
@@ -0,0 +1,95 @@
...
+
+        let getInput = GetObjectInput(
+            bucket: bucketName,
+            checksumMode: S3ClientTypes.ChecksumMode.enabled,
+            key: objectName
+            )
+        let responseGet = try await client.getObject(
+            input: getInput
+            ) // will error for normal payloads if checksum mismatch
+        XCTAssertNotNil(responseGet.body) // Ensure there's a body in the response.
```
sichanyoo, line 41: In addition to checking it's not nil, I think it's important to ensure the returned object really is the one that was just uploaded. Could we add a check for it? Something like `XCTAssertEqual(<response>, upload)` where `<response>` is the corresponding counterpart in `responseGet`.

sichanyoo (changes requested):
A comment on checking returned object

IntegrationTests/Services/AWSS3IntegrationTests/S3FlexibleChecksumsTests.swift
dayaffe: They have to be equal if we don't get a checksum mismatch exception! Since ChecksumMode = `ENABLED` we will calculate checksum of the response body and compare it to the one sent to us by S3. If the data is the same then checksum will be the same and we'll be good. If not then we won't. But actually, Ill add a check to make sure validation occurred.

sichanyoo: Oh I see , that makes sense 👍

sichanyoo (approved):
Lgtm!

dayaffe:
This will be merged with #1329

PR #1362, opened 2024-02-22
feat: add endpoint plugins

AndrewFossAWS:
## Description of changes
* Add endpoint plugin
* Because endpoint resolve is client dependent, it needs to code generated.
* See https://github.com/smithy-lang/smithy-swift/pull/671 for non code generated plugins


codegen/smithy-aws-swift-codegen/src/main/kotlin/software/amazon/smithy/aws/swift/codegen/plugins/EndpointPlugin.kt
```diff
@@ -0,0 +1,35 @@
...
+            writer.openBlock("public init(endpointResolver: \$L) {", "}", AWSServiceTypes.EndpointResolver) {
+                writer.write("self.endpointResolver = endpointResolver")
+            }
+            writer.openBlock("public func configureClient(clientConfiguration: ClientRuntime.ClientConfiguration) throws {", "}") {
+                writer.openBlock("if var config = clientConfiguration as? ${serviceConfig.typeName} {", "}") {
+                    writer.write("config.endpointResolver = self.endpointResolver")
+                }
+            }
+        }
+    }
```
sichanyoo, line 34: Why does the endpoint resolver plugin have to be code-generated? Can we maybe add the property `endpointResolver` to the base client config protocol in smithy-swift and have a plugin written in AWSClientRuntime that works the same as other plugins?

Sources/Services/AWSXRay/XRayClient.swift
```diff
@@ -127,6 +127,18 @@ public struct XRayClientLogHandlerFactory: ClientRuntime.SDKLogHandlerFactory {
...
+    public init(endpointResolver: EndpointResolver) {
+        self.endpointResolver = endpointResolver
+    }
+    public func configureClient(clientConfiguration: ClientRuntime.ClientConfiguration) throws {
+        if var config = clientConfiguration as? XRayClient.XRayClientConfiguration {
+            config.endpointResolver = self.endpointResolver
+        }
+    }
+}
+
```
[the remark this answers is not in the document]
AndrewFossAWS: End users can pass in their own customized endpoint resolver via the EndpointPlugin constructor, or they can just implement their own plugin similar to this.

codegen/smithy-aws-swift-codegen/src/main/kotlin/software/amazon/smithy/aws/swift/codegen/plugins/EndpointPlugin.kt
sichanyoo: Oh, so the base protocol itself is code-generated for each service. Maybe we can delete the codegen for that protocol given it seems to be same thing for every service, move it to AWSClientRuntime, and do the approach I mentioned in original comment. That way we can reduce 5k lines of fundamentally duplicate codegen.

sichanyoo: Hmm.. forgot that endpoint params for each service has different values, and maybe that makes the approach I suggested be not feasible.

codegen/smithy-aws-swift-codegen/src/main/kotlin/software/amazon/smithy/aws/swift/codegen/plugins/EndpointPlugin.kt
```diff
@@ -0,0 +1,35 @@
...
+class EndpointPlugin(private val serviceConfig: ServiceConfig) : Plugin {
+    override val className: Symbol
+        get() = buildSymbol {
+            this.name = "EndpointPlugin"
+            this.namespace = AWSSwiftDependency.AWS_CLIENT_RUNTIME.target
+            dependency(AWSSwiftDependency.AWS_CLIENT_RUNTIME)
+        }
+
+    override val isDefault: Boolean
+        get() = false
```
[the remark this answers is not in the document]
AndrewFossAWS: The `Plugin` interface requires the implementations to specify whether it's default or not.

PR #1370, opened 2024-02-28
chore: Improve AWSS3 SigV4A integration tests to use preexisting MRAP if applicable.

sichanyoo:
## Issue \#

https://github.com/awslabs/aws-sdk-swift/issues/1371

## Conventional Commits

### DIFF
--- a/IntegrationTests/Services/AWSS3IntegrationTests/S3SigV4ATests.swift
+++ b/IntegrationTests/Services/AWSS3IntegrationTests/S3SigV4ATests.swift
@@ -27,25 +27,28 @@ class S3SigV4ATests: S3XCTestCase {
     private var mrapArnFormat = "arn:aws:s3::%@:accesspoint/%@"
     private var mrapArn: String!
     private var mrapAlias: String!
+    private let mrapNamePrefix = "aws-sdk-s3-integration-test-"
     private let mrapName = "aws-sdk-s3-integration-test-" + UUID().uuidString.split(separator: "-").first!.lowercased()
+    private var mrapConfig: S3ControlClientTypes.CreateMultiRegionAccessPointInput!
 
     // The S3 control client used to create and delete MRAP
     private var s3ControlClient: S3ControlClient!
-    
+
     // The STS client for fetching AWS account ID
     private var stsClient: STSClient!
     private var accountId: String!
 
     // Key string used for putting object in tests
-    private let key = "text.txt"
+    private let key = UUID().uuidString.split(separator: "-").first!.lowercased()
 
     private let NSEC_PER_SEC = 1_000_000_000
 
-    override func setUp() async throws {
-        // Create a bucket
-        try await super.setUp()
+    // MARK: - SETUP & TEARDOWN
 
-        // Create sigv4a-only S3 client
+    override func setUp() async throws {
+        s3ControlClient = try S3ControlClient(region: region)
+
+        // Create sigv4a-only S3 client to be used for tests.
         sigv4aConfig = try await S3Client.S3ClientConfiguration(region: region)
         sigv4aConfig.authSchemes = [SigV4AAuthScheme()]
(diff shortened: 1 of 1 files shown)


IntegrationTests/Services/AWSS3IntegrationTests/S3SigV4ATests.swift
```diff
@@ -199,4 +146,59 @@ class S3SigV4ATests: S3XCTestCase {
...
+        var status: S3ControlClientTypes.MultiRegionAccessPointStatus? = .creating
+        repeat {
+            let seconds = 20.0
+            try await Task.sleep(nanoseconds: UInt64(seconds * Double(NSEC_PER_SEC)))
+
+            status = try await s3ControlClient.getMultiRegionAccessPoint(input: GetMultiRegionAccessPointInput(
+                accountId: accountId,
+                name: mrapName
+            )).accessPoint?.status
+        } while status == .creating
```
jbelkins, line 194: Do we need to have some sort of timeout mechanism for this `repeat` loop?

IntegrationTests/Services/AWSS3IntegrationTests/S3SigV4ATests.swift
```diff
@@ -199,4 +146,59 @@ class S3SigV4ATests: S3XCTestCase {
...
+            accountId: accountId,
+            clientToken: UUID().uuidString.split(separator: "-").first!.lowercased(),
+            details: mrapConfig
+        )
+        _ = try await s3ControlClient.createMultiRegionAccessPoint(input: createMRAPInput)
+
+        // Wait until MRAP creation finishes
+        var status: S3ControlClientTypes.MultiRegionAccessPointStatus? = .creating
+        repeat {
+            let seconds = 20.0
```
jbelkins, line 187: nit: use `TimeInterval` type for a number of seconds

IntegrationTests/Services/AWSS3IntegrationTests/S3SigV4ATests.swift
```diff
@@ -54,84 +57,28 @@ class S3SigV4ATests: S3XCTestCase {
...
+            // Construct MRAP and save its ARN for use in tests.
+            _ = try await constructMRAP()
+        }
     }
 
     override func tearDown() async throws {
-        // MRAP must be deleted before deleting the linked bucket
-        try await deleteMRAP()
-        // Empty & delete the bucket
-        try await super.tearDown()
```
jbelkins, line 105: This test class inherits `S3XCTestCase`, which does have a tear-down... we do call `super.setUp()` at least in some cases, which creates a test bucket... do we need to call `super.tearDown()` as well?

jbelkins (changes requested):
See questions.  I suspect changes are required but have a look.

IntegrationTests/Services/AWSS3IntegrationTests/S3SigV4ATests.swift
sichanyoo: The `super.tearDown()` deletes the bucket that was created in `super.setUp()`, but since we're keeping the MRAP, we can't delete the bucket. To delete the bucket, we have to delete MRAP first. Otherwise it would result in an error. That's why the it's a no-op now instead of deleting MRAP then the bucket.

IntegrationTests/Services/AWSS3IntegrationTests/S3SigV4ATests.swift
sichanyoo: Added a simple polling limit of 20 with poll count incrementing by 1 each time.

IntegrationTests/Services/AWSS3IntegrationTests/S3SigV4ATests.swift
sichanyoo: Changed as suggested.

jbelkins (approved):

PR #1378, opened 2024-03-05
feat: add service specific AuthSchemePlugin and DefaultAWSAuthSchemePlugin

AndrewFossAWS:
## Description of changes
Add service specific AuthSchemePlugin and DefaultAWSAuthSchemePlugin


Sources/Services/AWSACM/Plugins.swift
```diff
@@ -0,0 +1,59 @@
...
+            }
+            if (self.authSchemeResolver != nil) {
+                config.authSchemeResolver = self.authSchemeResolver!
+            }
+            if (self.awsCredentialIdentityResolver != nil) {
+                config.awsCredentialIdentityResolver = self.awsCredentialIdentityResolver!
+            }
+        }
+    }
+}
```
sichanyoo, line 59: Q: Why do we have two plugins for auth scheme related things? Could these maybe be combined into one plugin `ACMClientAuthSchemePlugin` by checking if set value is nil (like it's doing rn) and if it is nil, setting it to default value (like its done in `DefaultAWSAuthSchemePlugin`?

codegen/smithy-aws-swift-codegen/src/main/kotlin/software/amazon/smithy/aws/swift/codegen/plugins/AuthSchemePlugin.kt
```diff
@@ -0,0 +1,63 @@
...
+
+            writer.openBlock(
+                "public init(authSchemes: \$N = nil, authSchemeResolver: \$N = nil, awsCredentialIdentityResolver: \$N = nil) {", "}",
+                ClientRuntimeTypes.Auth.AuthSchemes.toOptional(),
+                AuthSchemeResolverGenerator.getServiceSpecificAuthSchemeResolverName(ctx).toOptional(),
+                AWSClientRuntimeTypes.Core.AWSCredentialIdentityResolver.toGeneric().toOptional()
+            ) {
+                writer.write("self.authSchemeResolver = authSchemeResolver")
+                writer.write("self.authSchemes = authSchemes")
+                writer.write("self.awsCredentialIdentityResolver = awsCredentialIdentityResolver")
```
sichanyoo, line 45: Q: Why is credential identity resolver part of the service specific auth scheme plugin, but not the default AWS auth scheme plugin? Is this overlap between auth scheme plugin & credential identity resolver plugin intentional?

sichanyoo (commented):
Left a couple questions.

codegen/smithy-aws-swift-codegen/src/main/kotlin/software/amazon/smithy/aws/swift/codegen/plugins/AuthSchemePlugin.kt
AndrewFossAWS: > Is this overlap between auth scheme plugin & credential identity resolver plugin intentional?
Yes, I thought they are all related Identity and authentication so I decided to group in the same plugin.

sichanyoo: I see, that's fine, but should it be consistent between default plugin and service-specific plugin? Default one does not have identity resolver but service-specific one does. Also, does this make identity resolver plugin redundant?

syall: I think the identity resolver plugin doesn't exist in the PR anymore, but the title and description need to be updated.

syall (approved):

sichanyoo (approved):

PR #1395, opened 2024-03-08
fix: Add overwrite behavior for endpoints 2.0 for signing properties

sichanyoo:
## Description of changes

- Add codegen for overwriting signing property value for `signingName` and `signingRegion` if auth scheme is resolved from the resolved endpoint and it has custom values for these.
- Update a codegen test

## Conventional Commits


jbelkins (approved):
Approved pending green build

Issue #1400, opened 2024-03-11
Implement "S3 Expires Shape Customization" SEP


smivij:

sichanyoo:
IF Smithy team makes the changes before we go GA, no action should be needed from Swift SDK team.
Needs further investigation on:
- Whether protocol tests have been written / wip by Smithy team for this
- When the Smithy team is planning to change the type of `Expires` shape from `timestamp` to `string` (as of 3/11/24, it is still `timestamp`)

sichanyoo:
AWS Swift SDK has a `SwiftIntegration` in place that "re-shapes" Expires member shape in S3 model to target `smithy.api#String` instead of `smithy.api#Timestamp` IF it is nested under `XYZOutput` shapes.

[This](https://github.com/awslabs/aws-sdk-swift/blob/main/codegen/smithy-aws-swift-codegen/src/main/kotlin/software/amazon/smithy/aws/swift/codegen/customization/s3/S3Expires.kt) is the integration, and it was added with [this PR](https://github.com/awslabs/aws-sdk-swift/pull/709).

This means that even if Smithy team changes Expires shape to be String in C2J / Smithy model resulting from C2J tool, the behavior of Swift SDK **will remain the same for deserialization** because we already treat it as string and copy the string from HTTP header into string property of the output type.

However, the planned changes by Smithy team outlined by `--/seps/accepted/shared/s3-expires-shape-customization.md` SEP includes `Expires` shape nested under input shapes as well (SIMs: `SMITHY-2535` & `SMITHY-2536`). I'll be putting up a PR soon that links this issue, that expands pre-existing integration to include `Expires` nested under input shapes as well.

dayaffe:
Released in 0.43

PR #1466, opened 2024-04-29
feat!: Make Expires member in S3 always be a String

sichanyoo:
## Issue \#

GH issue: https://github.com/awslabs/aws-sdk-swift/issues/1400
SEP: `--/seps/accepted/shared/s3-expires-shape-customization.md`

## Description of changes

- Resulting codegen change in S3 model is included in PR for quick reference.
- Expand pre-existing model preprocessing where `Expires` member shape gets changed from targeting timestamp to string shape if it were nested under output shapes, to change it regardless of where it is.
- This change future-proofs our SDK for this particular SEP. Regardless of whether Smithy team makes the change to the C2J tool to remove the reshaping logic before we GA or not, our SDK's behavior will stay the same (treat Expires as string).

## Conventional Commits

### DIFF
--- a/Sources/Services/AWSS3/models/Models.swift
+++ b/Sources/Services/AWSS3/models/Models.swift
@@ -1911,7 +1911,7 @@ extension CopyObjectInput {
             items.add(Header(name: "x-amz-source-expected-bucket-owner", value: Swift.String(expectedSourceBucketOwner)))
         }
         if let expires = value.expires {
-            items.add(Header(name: "Expires", value: Swift.String(TimestampFormatter(format: .httpDate).string(from: expires))))
+            items.add(Header(name: "Expires", value: Swift.String(expires)))
         }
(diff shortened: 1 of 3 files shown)


Sources/Services/AWSS3/models/Models.swift
```diff
@@ -1911,7 +1911,7 @@ extension CopyObjectInput {
             items.add(Header(name: "x-amz-source-expected-bucket-owner", value: Swift.String(expectedSourceBucketOwner)))
```
sichanyoo, line 1911: S3 model codegen output; included in PR for quick reference to reviewers.

syall (approved):

jbelkins (approved):
Discussed offline with Chan.  Primary effect is to ensure that abnormal string timestamp values in the `Expires` field do not fail an otherwise valid request.

Approved.

PR #1498, opened 2024-05-13
feat: Require Swift DocC only when env var is set

jbelkins:
## Description of changes
Removes DocC as a SDK dependency unless the env var `AWS_SWIFT_SDK_ENABLE_DOCC` is set, since it is essentially unused unless documentation is actually being generated.  There is generally not a reason to enable it by default for customers.

`AWS_SWIFT_SDK_ENABLE_DOCC` is set for the "generate documentation" job on Github, where DocC is actually used.

### DIFF
--- a/.github/workflows/generate-documentation.yml
+++ b/.github/workflows/generate-documentation.yml
@@ -17,6 +17,7 @@ jobs:
     runs-on: ubuntu-latest
     container: swift:5.10-jammy
     env:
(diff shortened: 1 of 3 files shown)


dayaffe (approved):

sichanyoo (approved):
There's a swiftlint warning that got past review from some other PR. Address before merge if possible

Issue #874, opened 2023-02-10
OSMetadata doesn't differentiate between Simulator versus Real device


phantumcode:
### Describe the feature

[OSMetadata](https://github.com/awslabs/aws-sdk-swift/blob/6d7c4a1ce4038149fc13b77aa648d33e9048a9c8/Sources/Core/AWSClientRuntime/Http/OSMetadata.swift) used to calculate the [AWSUserAgentMetadata](https://github.com/awslabs/aws-sdk-swift/blob/6d7c4a1ce4038149fc13b77aa648d33e9048a9c8/Sources/Core/AWSClientRuntime/Http/OSMetadata.swift) does not differentiate when the AWS Swift SDK is being executed in an iOS/MacOS Simulator versus a real device.

As a result, the AWS UserAgent header on request has inaccurate OS information and does include detail that the environment is on a simulator.

i.e.
```
User-Agent: aws-sdk-swift/1.0 api/cognito-identity-provider/1.0 os/iOS/16.2.0 lang/swift/5.7 lib/amplify-ios/2.2.0
```

### Use Case

Amplify Swift library would like the User Agent to differentiate when running on iOS simulator and macOS simulator to differentiate between developer debug/simulation versus when running in real device for more accurate and improved granularity for calculating Amplify Swift library and Swift SDK usage metrics.

### Proposed Solution

**Option 1**: Update [PlatformOperatingSystem](https://github.com/awslabs/smithy-swift/blob/1fbd5f7bebd024e29c577240a960cd011ce71ac4/Sources/ClientRuntime/Util/PlatformOperatingSystem.swift) in smithy-swift repo with new enum cases of iOSSimulator, watchOSSimulator, tvOS Simulator, and macOSSImulator

```swift
public enum PlatformOperatingSystem: String {
    case windows
    case linux
    case iOS
    case macOS
    case watchOS
    case tvOS
    case iOSSimulator
    case macOSSimulator
    case watchOSSimulator
    case tvOSSimulator
    case unknown
}

public var currentOS: PlatformOperatingSystem {
    #if os(Linux)
    return .linux
    #elseif os(macOS)
         #if targetEnvironment(simulator)
         return .macOSSimulator
         #else
         return .macOS
         #endif
    #elseif os(iOS)
          #if targetEnvironment(simulator)
          return .iOSSimulator
          #else
          return .iOS
          #endif
    #elseif os(watchOS)
         #if targetEnvironment(simulator)
[...]

PR #1523, opened 2024-05-22
feat: Add simulator as additional metadata of OS metadata in user agent

sichanyoo:
## Issue \#

https://github.com/awslabs/aws-sdk-swift/issues/874

## Description of changes

Using compiler directive, add "simulator" as additional metadata to OS metadata in user agent header if SDK is being used within simulators.

Also, make the `value` of additional metadata optional to follow the SEP.

The change was tested using a simple iOS application & confirming that user agent header header is changed as expected:
```
User-Agent: aws-sdk-swift/1.0 ua/2.0 api/s3#1.0 os/ios#17.4.0 md/simulator lang/swift#5.10 cfg/retry-mode#legacy
```

## Conventional Commits

### DIFF
--- a/Sources/Core/AWSClientRuntime/UserAgent/AdditionalMetadata.swift
+++ b/Sources/Core/AWSClientRuntime/UserAgent/AdditionalMetadata.swift
@@ -9,12 +9,21 @@ import Foundation
 
 struct AdditionalMetadata {
     let name: String
-    let value: String
+    let value: String?
+
+    init(name: String, value: String? = nil) {
+        self.name = name
+        self.value = value
+    }
 }
 
 extension AdditionalMetadata: CustomStringConvertible {
 
     var description: String {
-        "md/\(name.userAgentTokenNoHash)#\(value.userAgentToken)"
+        if let value {
+            "md/\(name.userAgentTokenNoHash)#\(value.userAgentToken)"
+        } else {
+            "md/\(name.userAgentTokenNoHash)"
+        }
     }
 }
--- a/Sources/Core/AWSClientRuntime/UserAgent/AdditionalMetadata.swift
+++ b/Sources/Core/AWSClientRuntime/UserAgent/AdditionalMetadata.swift
@@ -21,9 +21,9 @@ extension AdditionalMetadata: CustomStringConvertible {
 
     var description: String {
         if let value {
-            "md/\(name.userAgentTokenNoHash)#\(value.userAgentToken)"
+            return "md/\(name.userAgentTokenNoHash)#\(value.userAgentToken)"
         } else {
-            "md/\(name.userAgentTokenNoHash)"
+            return "md/\(name.userAgentTokenNoHash)"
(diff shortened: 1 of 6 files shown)


Sources/Core/AWSClientRuntime/UserAgent/ExecutionEnvMetadata.swift
```diff
@@ -18,6 +18,9 @@ extension ExecutionEnvMetadata: CustomStringConvertible {
     }
 
     static func detectExecEnv() -> ExecutionEnvMetadata? {
+        #if targetEnvironment(simulator)
+        return ExecutionEnvMetadata(name: "simulator")
```
jbelkins, line 22: From the user-agent SEP:
> Execution environment name can be retrieved from the environment variable AWS_EXECUTION_ENV as defined in this [SEP](https://code.amazon.com/packages/AwsDrSeps/blobs/ca26ae803530fb10429a218c1c0a3e25ee07cceb/--/seps/accepted/shared/user_agent.md). The SDK MUST include this in the header if present.

I don't think simulator is the intended use of this UA field, and I'm afraid we will mess up data collection if we put other values into this field.

jbelkins (changes requested):

Sources/Core/AWSClientRuntime/UserAgent/ExecutionEnvMetadata.swift
sichanyoo: That's fair, I think what I could do is append this as additional metadata to OS metadata then.

Sources/Core/AWSClientRuntime/UserAgent/AdditionalMetadata.swift
```diff
@@ -9,12 +9,21 @@ import Foundation
```
sichanyoo, line 9: The `value` is optional according to SEP. Changed accordingly.

Sources/Core/AWSClientRuntime/UserAgent/OSMetadata.swift
```diff
@@ -11,11 +11,18 @@ struct OSMetadata {
     let family: PlatformOperatingSystem
```
sichanyoo, line 11: If compiled for simulator target, add `md/simulator` additional metadata.

Tests/Core/AWSClientRuntimeTests/UserAgent/AWSUserAgentMetadataTests.swift
```diff
@@ -17,6 +17,15 @@ class AWSUseragentMetadataTests: XCTestCase {
     let executionEnvMetadata = ExecutionEnvMetadata(name: "e123")
```
sichanyoo, line 17: Add a test for simulator case.

jbelkins (approved):
Approved after consulting with SDK data team about how to implement.

Issue #1201, opened 2023-10-27
Conceal "unknown error" types


jbelkins:
### Describe the feature

After a conversation with Smithy, it was suggested that SDKs structure their code and/or guide customers to write their error handling code in a way that is compatible with adding modeled errors after services have launched.

For Swift, this is challenging because Swift developers' first instinct (and community best practice) is to catch errors by type.  If a Swift customer catches an "unknown error" by its type, the error cannot be converted to a modeled error later because changing the concrete type of an existing error is a breaking change.

Because both modeled and unmodeled errors conform to several of the same protocols, i.e. `AWSServiceError`, new error types can be added safely if non-modeled errors are caught by interface rather than by the unknown error type, i.e.
```swift
do {
    let output = try await client.getObject(input: input)
} catch ModeledError1 {
    print("Modeled Error 1 occurred")
} catch ModeledError2 {
    print("Modeled Error 2 occurred")
} catch let serviceError as AWSServiceError {
    switch serviceError.errorCode {
    case "UnmodeledError1":
        // This error could be of any type that conforms to AWSServiceError and has the error code "UnmodeledError1".
        // This error could be of a concrete modeled error type or of a "unknown error" type meant to represent
        // unmodeled errors.
        print("A known/expected but unmodeled error occurred here")
    default:
        print("An unknown error was returned by the service with code: \(serviceError.errorCode)")
    }
} catch {
    print("An unknown non-service error occurred")
}
```

We can encourage customers to catch errors by code and not by the unknown error type by directing them to do so in the docs, but since this approach is not a Swift developer's first instinct and many devs might write code without reading docs, many developers will end up with code that breaks when an error is converted from an unmodeled to modeled type.  There is no clear way that we can have the Swift compiler or other dev tools enforce this preference either.

A better approach than to make recommendations in the docs would be to make the unknown error type not public - if it's not public, you can't catch it by type.  This would force developers to use a mechanism such as the one above which does not rely on the existence of any concrete error type that is not already a modeled type.  This would also allow the compiler to steer customers toward the best means of error handling, by not exposing the concrete error type that we don't want customers to match.

Errors that change from one modeled type to another, or from a modeled to unmodeled type, are much rarer cases and would still remain breaking changes.

### Use Case

To prevent breaking customer error handling code when a new modeled error type is introduced to a service or operation.

### Proposed Solution

[...]

PR #1528, opened 2024-05-24
feat!: Conceal unknown error type

sichanyoo:
Companion PR: https://github.com/smithy-lang/smithy-swift/pull/740

## Issue \#

https://github.com/awslabs/aws-sdk-swift/issues/1201

## Description of changes

Changes needed to make `UnknownAWSHTTPServiceError` runtime struct internal with `@_spi(UnknownAWSHTTPServiceError)`.
Updates codegen & tests accordingly.

## Conventional Commits


Sources/Core/AWSClientRuntime/Errors/UnknownAWSHTTPServiceError.swift
```diff
@@ -6,7 +6,8 @@
 import ClientRuntime
 
 /// AWS specific Service Error structure used when exact error could not be deduced from the `HttpResponse`
-public struct UnknownAWSHTTPServiceError: AWSServiceError, HTTPError, Error {
+/// Developers should catch unknown errors by the interface `AWSServiceError`, then use the `errorCode` to determine & handle each type of error.
+@_spi(Internal) public struct UnknownAWSHTTPServiceError: AWSServiceError, HTTPError, Error {
```
sichanyoo, line 10: Makes this struct part of SPI, with "Internal" tag.

Sources/Services/AWSS3/models/Models.swift
```diff
@@ -1,4 +1,5 @@
 // Code generated by smithy-swift-codegen. DO NOT EDIT!
+@_spi(Internal) import struct AWSClientRuntime.UnknownAWSHTTPServiceError
 import AWSClientRuntime
```
sichanyoo, line 3: Example generated model output.

codegen/smithy-aws-swift-codegen/src/main/kotlin/software/amazon/smithy/aws/swift/codegen/AWSClientRuntimeTypes.kt
```diff
@@ -43,7 +45,7 @@ object AWSClientRuntimeTypes {
         val Partition = runtimeSymbol("Partition")
         val ServiceEndpointMetadata = runtimeSymbol("ServiceEndpointMetadata")
         val CredentialScope = runtimeSymbol("CredentialScope")
-        val UnknownAWSHTTPServiceError = runtimeSymbol("UnknownAWSHTTPServiceError")
+        val UnknownAWSHTTPServiceError = runtimeSymbol("UnknownAWSHTTPServiceError").toInternalSPI(SwiftKind.STRUCT)
         val AWSServiceError = runtimeSymbol("AWSServiceError")
```
sichanyoo, line 49: Mark this symbol with properties used by `SwiftWriter` to import it as SPI.

scripts/mergeModels.sh
```diff
@@ -24,8 +24,8 @@ MODELS="Models.swift"
 mergeFiles() {
     TEMP=`mktemp`
     cat *.swift | grep -e "//\ Code\ generated" | sort | uniq >> ${TEMP}
-    cat *.swift | grep -e "^import\ " | sort | uniq  >> ${TEMP}
-    cat *.swift | grep -ve "^import\ " | grep -ve "//\ Code\ generated" >> ${TEMP}
+    cat *.swift | grep -e "^import\ " -e "@_spi" | sort | uniq  >> ${TEMP}
+    cat *.swift | grep -ve "^import\ " -ve "@_spi" -ve "//\ Code\ generated" >> ${TEMP}
```
sichanyoo, line 28: Without this change, in the merged model file result, there would be dozens of `@_spi` statements. This ensures there's no no duplicate SPI imports in merged file.

sichanyoo (commented):
Explanation comments for reviewers

Sources/Core/AWSClientRuntime/Errors/UnknownAWSHTTPServiceError.swift
jbelkins: This needs a more specific name than "Internal".  Reason is that a single module may have multiple SPI's for different functions, so each function needs a specific name that allows the importer to only import the SPI they need or care about.

I'd suggest the type name `UnknownAWSHTTPServiceError` as your SPI identifier, since that's basically what you're concealing.

scripts/mergeModels.sh
jbelkins: This is great.  When I implemented `@_spi()` in codegen a year ago, I kicked the can on this.  Thanks for mopping it up!

jbelkins (changes requested):
Minor stuff, mostly all good

Sources/Core/AWSClientRuntime/Errors/UnknownAWSHTTPServiceError.swift
sichanyoo: Changed as suggested.

jbelkins (approved):

Issue #1532, opened 2024-05-30
Modularize Checksums


jbelkins:
### Describe the feature

The checksums feature needs to be split from ClientRuntime to its own module in smithy-swift.
- Checksums interfaces & data types should go into a `SmithyChecksumsAPI` module.
- CRT-dependent checksum implementation code should go into a `SmithyChecksums` module.

There is checksum-related logic in the signer code as well; we should investigate separating those two functions.

### Use Case

To support reference architecture and to allow for future flexible use of Smithy components.

### Acknowledgements

- [ ] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

PR #1568, opened 2024-06-13
feat: modularize checksums

dayaffe:
New modularization package:
- AWSSDKChecksums

## Issue \#
#1532 

## Description of changes
- AWSSDKChecksums contains details specific to the AWS part of checksums. As of now this includes the method to add the aws-chunked header. AWSSDKChecksums can only contain checksum and chunking related methods that are specific to AWS and are not dependencies of other parts of the smithy-swift package.
- Adds additional dependencies from smithy-swift including SmithyChecksums and SmithyStreams
- Protocol tests now depend on SmithyStreams and a symbol has been added for SmithyStreams.BufferedStream.
- FlexibleChecksums Request/Response middlewares have been moved from smithy-swift to aws-sdk-swift

## Conventional Commits


Sources/Core/AWSSDKHTTPAuth/AWSSigV4Signer.swift
```diff
@@ -255,3 +256,14 @@ public class AWSSigV4Signer: SmithyHTTPAuthAPI.Signer {
         }
     }
 }
+
+extension SigningConfig {
```
dayaffe, line 260: this method doesnt rely on checksums

Sources/Core/AWSSDKHTTPAuth/AWSSigV4Signer.swift
```diff
@@ -30,6 +30,7 @@ import struct SmithyHTTPAuthAPI.SigningFlags
 import struct Foundation.Date
 import struct Foundation.TimeInterval
 import struct Foundation.URL
+import AWSSDKChecksums
```
dayaffe, line 33: the signer still needs to depend on AWSSDKChecksums for setChunkedBody method

AWSSDKSwiftCLI/Sources/AWSSDKSwiftCLI/Resources/Package.Base.swift
```diff
@@ -22,6 +22,7 @@ extension Target.Dependency {
     static var awsSDKEventStreamsAuth: Self { "AWSSDKEventStreamsAuth" }
     static var awsSDKHTTPAuth: Self { "AWSSDKHTTPAuth" }
     static var awsSDKIdentity: Self { "AWSSDKIdentity" }
+    static var awsSDKChecksums: Self { "AWSSDKChecksums" }
```
sichanyoo, line 25: Question: Is a separate module actually necessary for just a single extension with two methods?

sichanyoo (commented):
Question on need for AWSSDKChecksums module

AWSSDKSwiftCLI/Sources/AWSSDKSwiftCLI/Resources/Package.Base.swift
dayaffe: I wish more could be moved to AWSSDKChecksums but that's dependent on CRT HTTP Client being refactored in the future. I'd say even if its only two methods its good to have but I don't feel strongly. `aws-chunked` is pretty explicitly aws-only though and one of those methods adds it to the headers

sichanyoo: I see. We could keep it the way it is in this PR, or move it into AWSClientRuntime. For both cases we would revisit this down the line. I don't feel strongly about either option. Unless Josh has preference for one over the other I suppose we can keep `AWSSDKChecksums`.

jbelkins (approved):

Issue #1550, opened 2024-06-06
Create free-standing Integration Test package


jbelkins:
### Describe the feature

Create a package within the aws-sdk-swift repository for AWS integration tests.
It should contain a Package.swift and all of the current integration tests.

All integration tests should run when `swift test` is executed in the package root directory.

There should also be test plans, schemes, etc. provided to allow integration tests to run using Xcode and on Mac plus on Apple platform simulators.

### Use Case

To be able to run integration tests after the project is modularized.

### Acknowledgements

- [ ] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

PR #1585, opened 2024-06-20
feat!: Standalone integration test package

sichanyoo:
## Issue \#

https://github.com/awslabs/aws-sdk-swift/issues/1550

## Description of changes

- Add a new `Package.swift` under `IntegrationTests/` to separate out integration test from main package under `aws-sdk-swift/`.
- Remove integration test things from package codegen, `Package.Base.swift`, and CI.

## Conventional Commits


AWSSDKSwiftCLI/Sources/AWSSDKSwiftCLI/Commands/GeneratePackageManifest.swift
```diff
@@ -32,8 +32,8 @@ struct GeneratePackageManifestCommand: ParsableCommand {
     @Option(help: "The names of the services to include in the package manifest. This defaults to all services located in aws-sdk-swift/Sources/Services")
     var services: [String] = []
 
-    @Flag(help: "If the package manifest should include the integration tests.")
-    var includeIntegrationTests: Bool = false
+    @Flag(help: "If the package manifest should include services with integration tests.")
+    var includeServicesWithIntegTests: Bool = false
```
jbelkins, line 36: Why do we need this option at all in the main Package.swift if we have a whole separate Package.swift for integration tests now?

If we code-generate with the special sdk.properties file for integration tests, shouldn't that ensure that all the needed services for integration tests are present?

jbelkins (commented):
A question about why integration test related stuff remains in main Package.swift.

AWSSDKSwiftCLI/Sources/AWSSDKSwiftCLI/Commands/GeneratePackageManifest.swift
sichanyoo: That's a good point, for some reason I thought service targets will need to be available from the OG Package.swift for it to be available for integration tests, but as you said the source files are already there and I can just add targets to them in the new Package.swift anyways. Will update accordingly.

sichanyoo: Actually, it seems I can't add targets to generated SDK services from within `IntegrationTests/Package.swift`, because apparently you can't refer to files outside the package root (where Package.swift is located). It seems enabling services for integration tests must be done as it is by the OG Package manifest.

jbelkins: Can we use symlinks to link the code?  Git will commit symlinks IIRC.

jbelkins: @sichanyoo Confirmed, you don't need any of the integration test logic in the main Package.swift anymore.
See changes in https://github.com/awslabs/aws-sdk-swift/pull/1591 .

jbelkins: #1591 doesw as I described at top: generate only the services used for integraion testing by using the special `sdk.properties` file to pick enabled services.  The logic in main Package.swift becomes superfluous as long as that properties file is maintained as tests and/or dependencies are changed.

IntegrationTests/Package.swift
```diff
@@ -83,9 +70,25 @@ func addClientRuntimeDependency() {
 }
 
 func addAWSClientRuntimeDependency() {
-    package.dependencies += [
-        .package(path: "../../aws-sdk-swift")
-    ]
+    let smithySwiftURL = "https://github.com/awslabs/aws-sdk-swift"
```
jbelkins, line 73: This local var is called `smithySwiftURL` but it contains the aws-sdk-swift URL

jbelkins: Also, can we combine this logic with the smithy-swift logic? (i.e. one case statement where two package dependencies are added in each case)

jbelkins (changes requested):

IntegrationTests/Package.swift
sichanyoo: Oof nice catch, and sure, one minute

jbelkins (approved):

PR #1590, opened 2024-06-25
feat: add ability to run protocol tests as unit tests of a service

dayaffe:
## Description of changes
- SignerMiddleware will check and update signing properties with X-Amz-Content-Sha256 if it is pre-computed by SHA256 middleware
- SHA256 Middleware now occurs before signing rather than before transmit. It also no longer adds the X-Amz-Content-Sha256 header and instead computes the value and passes it to CRT to add during signing. (Glacier fix)
- Create s3-tests.smithy with protocol tests that require using the actual service client. These tests will get generated and copied into the relevant service client test folder.
## Conventional Commits


Sources/Core/AWSClientRuntime/Middlewares/Sha256TreeHashMiddleware.swift
```diff
@@ -110,10 +112,10 @@ extension Sha256TreeHashMiddleware: HttpInterceptor {
     public typealias InputType = OperationStackInput
     public typealias OutputType = OperationStackOutput
 
-    public func modifyBeforeTransmit(context: some MutableRequest<Self.InputType, Self.RequestType>) async throws {
+    public func modifyBeforeSigning(context: some MutableRequest<Self.InputType, Self.RequestType>) async throws {
```
dayaffe, line 115: SHA256 middleware should occur prior to signing rather than adding headers aftwards

codegen/XCTestPlans/ProtocolTestPlan.xctestplan
```diff
@@ -0,0 +1,143 @@
...
+      {
+        "key" : "AWS_SECRET_ACCESS_KEY",
+        "value" : "your-secret-access-key"
+      },
+      {
+        "key" : "AWS_SESSION_TOKEN",
+        "value" : "your-session-token"
+      },
+    ]
+  },
```
dayaffe, line 34: Protocol Test Plan contains fake credentials. This is required for github actions simulators

codegen/smithy-aws-swift-codegen/src/main/kotlin/software/amazon/smithy/aws/swift/codegen/protocols/restxml/RestXMLProtocolGenerator.kt
```diff
@@ -14,19 +14,18 @@ class RestXMLProtocolGenerator : AWSHTTPBindingProtocolGenerator(RestXMLCustomiz
...
+        "S3VirtualHostDualstackAddressing", // can leave disabled, pre-endpoints 2.0
+        "S3VirtualHostAccelerateAddressing", // can leave disabled, pre-endpoints 2.0
+        "S3VirtualHostDualstackAccelerateAddressing", // can leave disabled, pre-endpoints 2.0
+        "S3OperationAddressingPreferred", // can leave disabled, pre-endpoints 2.0
+        "S3EscapeObjectKeyInUriLabel", // moved to s3-tests.smithy
+        "S3EscapePathObjectKeyInUriLabel", // moved to s3-tests.smithy
+        "SDKAppliedContentEncoding_restXml", // not implemented yet (request compression)
+        "SDKAppendedGzipAfterProvidedEncoding_restXml", // not implemented yet (request compression)
+        "S3PreservesEmbeddedDotSegmentInUriLabel", // moved to s3-tests.smithy
+        "S3PreservesLeadingDotSegmentInUriLabel", // moved to s3-tests.smithy
```
dayaffe, line 28: Tests that have been moved to s3-tests.smithy would still remain ignored here for when they are pulled from smithy. They have been copied and renamed.

.github/workflows/codegen-build-test.yml
```diff
@@ -61,5 +64,16 @@ jobs:
         run: ./scripts/ci_steps/codegen_sdk.sh
       - name: Build SDK with Unit Tests
         run: swift build --build-tests
+      - name: Create AWS config directory
```
jbelkins, line 67: Let's add a comment here about why we're going to the trouble of setting up these credentials in files, so future us doesn't forget about the manipulation of env vars in the credential resolver tests.

.github/workflows/codegen-build-test.yml
```diff
@@ -6,6 +6,9 @@ on:
   workflow_dispatch:
   pull_request:
 
+permissions:
+  id-token: write
+
```
jbelkins, line 11: Do we need this?  Doesn't look like we use `write` in this job.

.github/workflows/continuous-integration.yml
```diff
@@ -173,6 +178,11 @@ jobs:
         run: ./scripts/ci_steps/install_native_linux_dependencies.sh
       - name: Tools Versions
         run: ./scripts/ci_steps/log_tool_versions.sh
+      - name: Configure AWS Credentials for Integration Tests
+        uses: aws-actions/configure-aws-credentials@v3
+        with:
+          role-to-assume: ${{ secrets.INTEGRATION_TEST_ROLE_ARN }}
+          aws-region: us-west-2
```
jbelkins, line 185: Do we need to configure AWS credentials in CI?

.github/workflows/codegen-build-test.yml
jbelkins, line 67: Let's add a comment here about why we're going to the trouble of setting up these credentials in files, so future us doesn't forget about the manipulation of env vars in the credential resolver tests.

Sources/Services/AWSS3/Tests/AWSS3Tests/DeleteObjectTaggingRequestTest.swift
```diff
@@ -0,0 +1,73 @@
...
+class DeleteObjectTaggingRequestTest: HttpRequestTestBase {
+    /// S3 clients should escape special characters in Object Keys when the Object Key is used as a URI label binding.
+    func testProtocol_S3EscapeObjectKeyInUriLabel() async throws {
+        let urlPrefix = urlPrefixFromHost(host: "s3.us-west-2.amazonaws.com")
+        let hostOnly = hostOnlyFromHost(host: "s3.us-west-2.amazonaws.com")
+        let expected = buildExpectedHttpRequest(
+            method: .delete,
+            path: "/my%20key.txt",
+            headers: [
+                "Content-Type": "application/json"
```
jbelkins, line 19: Does this PR need to have its codegen updated?

codegen/.swiftpm/xcode/xcshareddata/xcschemes/aws-sdk-swift-protocol-tests-Package.xcscheme
jbelkins, line 1: This name matches the scheme that Xcode auto-generates for a package.  Are you sure we need this to be committed explicitly?

.github/workflows/continuous-integration.yml
```diff
@@ -109,8 +112,10 @@ jobs:
         run: |
           cd codegen/
           set -o pipefail && \
-          NSUnbufferedIO=YES xcodebuild \
+          xcodebuild -showTestPlans -scheme aws-sdk-swift-protocol-tests-Package && \
```
jbelkins, line 115: Was this here just for troubleshooting?  Can we get rid of it?

jbelkins (changes requested):
Overall good, but it appears there are several things in test setup & config that can be eliminated.

codegen/.swiftpm/xcode/xcshareddata/xcschemes/aws-sdk-swift-protocol-tests-Package.xcscheme
dayaffe: it doesnt need to be committed as it will get generated, but I thought we keep Sources/Services/ codegen committed and up to date?

.github/workflows/codegen-build-test.yml
dayaffe: I'll try removing it

.github/workflows/continuous-integration.yml
dayaffe: Yeah CI needs credentials though they can be fake, this was just the easiest way I got it to work.

Sources/Services/AWSS3/Tests/AWSS3Tests/DeleteObjectTaggingRequestTest.swift
dayaffe: oh yes it does

.github/workflows/continuous-integration.yml
```diff
@@ -109,8 +112,9 @@ jobs:
         run: |
           cd codegen/
           set -o pipefail && \
-          NSUnbufferedIO=YES xcodebuild \
+          xcodebuild \
```
jbelkins, line 115: is there a reason `NSUnbufferedIO=YES` was removed here?

.github/workflows/continuous-integration.yml
```diff
@@ -7,6 +7,9 @@ on:
 env:
   AWS_SWIFT_SDK_USE_LOCAL_DEPS: 1
 
+permissions:
+  id-token: write
+
```
jbelkins, line 12: are you sure we can't get rid of these extra permissions in CI?  IIRC this is needed to get AWS credentials through OpenID which we don't do in the CI workflow.

codegen/smithy-aws-swift-codegen/src/main/kotlin/software/amazon/smithy/aws/swift/codegen/protocols/restxml/RestXMLProtocolGenerator.kt
jbelkins: If tests that have been copied over to `s3-tests.smithy` are modified or added to, how will we know (other than manually checking those test models)?

.github/workflows/continuous-integration.yml
dayaffe: On the linux side I'm using `aws-actions/configure-aws-credentials@v3` still. Would rather I try to change this one to fake credentials?

codegen/smithy-aws-swift-codegen/src/main/kotlin/software/amazon/smithy/aws/swift/codegen/protocols/restxml/RestXMLProtocolGenerator.kt
dayaffe: they'll continue to be marked as disabled here with a comment noting they've been moved. Another place to check is in AdditionalServiceTests/<service> folder for the smithy file

.github/workflows/continuous-integration.yml
jbelkins: Yes, let's not get real credentials where fake ones will do

.github/workflows/continuous-integration.yml
jbelkins: As mentioned elsewhere, let's change this back to fake credentials.  I don't want to pull real credentials into a job that doesn't need them

jbelkins (approved):

PR #1620, opened 2024-07-10
chore: Separate Registry CLI tools from AWSSDKSwiftCLI

jbelkins:
## Issue \#
#1621

## Description of changes
Having the package manifest generator, located in `AWSSDKSwiftCLI`, depend on the SDK itself has proven problematic:
- The `AWS_SWIFT_SDK_USE_LOCAL_DEPS` env var, which must be set in most situations in development & CI, prevents the `AWSSDKSwiftCLI` package from building with stable aws-sdk-swift as a dependency.
- Using the local copy of aws-sdk-swift causes problems when `AWSSDKSwiftCLI` is run to update the manifest, but the manifest being updated is also part of the code being built.

The problem is addressed by separating the Registry CLI tools into a separate Swift package, allowing the `aws-sdk-swift` dependency to be removed from `AWSSDKSwiftCLI`.  `AWSSDKSwiftCLI` publishes its reusable `AWSCLIUtils` as a library, so it may be used by the Registry CLI as a dependency.


AWSSDKSwiftCLI/Package.swift
```diff
@@ -7,12 +7,14 @@ let package = Package(
     platforms: [
         .macOS(.v13)
     ],
+    products: [
+        .library(name: "AWSCLIUtils", targets: ["AWSCLIUtils"]),
+    ],
```
jbelkins, line 12: This product is published so `AWSCLIUtils` can be shared with other packages (like the new `SPRCLI` package.)

AWSSDKSwiftCLI/Package.swift
```diff
@@ -27,31 +29,6 @@ let package = Package(
...
-        ),
-        .target(
-            name: "SPR",
-            dependencies: [
-                "AWSCLIUtils",
-                .product(name: "ArgumentParser", package: "swift-argument-parser"),
-                .product(name: "AWSS3", package: "aws-sdk-swift"),
-                .product(name: "AWSCloudFront", package: "aws-sdk-swift"),
-            ]
-        ),
```
jbelkins, line 54: All the SPR targets in `AWSSDKSwiftCLI`, along with the SDK dependency, are moved to the new SPRCLI package defined below.

SPRCLI/Package.swift
```diff
@@ -0,0 +1,41 @@
...
+
+let package = Package(
+    name: "SPRCLI",
+    platforms: [
+        .macOS(.v13)
+    ],
+    dependencies: [
+        .package(url: "https://github.com/apple/swift-argument-parser", from: "1.2.0"),
+        .package(url: "https://github.com/awslabs/aws-sdk-swift", from: "0.46.0"),
+        .package(path: "../AWSSDKSwiftCLI"),
```
jbelkins, line 13: The `AWSCLIUtils` are accessed by adding the local copy of `AWSSDKSwiftCLI` as a dependency.

SPRCLI/Package.swift
```diff
@@ -0,0 +1,41 @@
...
+            ]
+        ),
+        .target(
+            name: "SPR",
+            dependencies: [
+                .product(name: "AWSCLIUtils", package: "AWSSDKSwiftCLI"),
+                .product(name: "AWSS3", package: "aws-sdk-swift"),
+                .product(name: "AWSCloudFront", package: "aws-sdk-swift"),
+            ]
+        ),
```
jbelkins, line 39: The three targets above were copied in from `AWSSDKSwiftCLI`.

scripts/ci_steps/codegen_sdk.sh
```diff
@@ -12,7 +12,6 @@ rm -rf Tests/Services/*
 ./gradlew --stop
 
 # Regenerate the SDK Package.swift with all services
-unset AWS_SWIFT_SDK_USE_LOCAL_DEPS
```
jbelkins, line 15: Un-setting this env var is no longer needed when building `AWSSDKSwiftCLI` since it does not depend on the SDK anymore.

sichanyoo (commented):
👍

PR #1630, opened 2024-07-13
feat: finish interceptors migration

milesziemer:
## Description of changes

smithy-swift pr: https://github.com/smithy-lang/smithy-swift/pull/781

This commit includes the codegen changes that move all services to use orchestrator + interceptors, instead of the operation stack using middleware. It also removes all references to the old types related to Middleware, in Swift libraries and in the code generator.

Since this is already a huge PR, I tried to limit any refactoring, which may have made it harder to review. For example, I didn't change the names of middleware or their file names.

This commit also adds two new interceptors, AmzSdkRequestMiddleware and AmzSdkInvocationIdMiddleware, which add corresponding headers to requests. They are added in codegen through a SwiftIntegration. Previously, these interceptors were implemented as a part of RetryMiddleware in smithy-swift. But unlike all other middleware, RetryMiddleware does not have an interceptor implementation - its logic is inlined within Orchestrator - so the logic for adding these amz-sdk headers had to be extracted. Plus, they seem to be aws headers, so I moved them here into the SDK. I also copied the RetryIntegrationTests test, which was testing the headers' functionality.

Other than that, there were some codegen changes I made which impacted the ordering that interceptors are rendered into operation bodies, which meant I needed to fix some codegen tests. The order interceptors are added to orchestrator shouldn't matter.


milesziemer:
Closing for https://github.com/awslabs/aws-sdk-swift/pull/1631 so CI will run

Issue #1629, opened 2024-07-11
Check implementation against S3 200 error SEP


sichanyoo:
S3 200 SEP has been written since S3 200 error implementation.
Check that the implementation matches up with SEP

PR #1633, opened 2024-07-15
fix: S3 200 error customization should only not apply to streaming + blob shapes

dayaffe:
## Issue \#
#1629 

## Description of changes
- S3 200 Error customization should only not apply to streaming + blob shapes not all streaming shapes.
- This change only effects one method `selectObjectContent` which didn't have the customization applied but now does. The response of this method contains an event stream.
- We should only check the root of the response for an <Error> node not if <Error> is present anywhere in the response

## Conventional Commits


jbelkins:
so event streams DO get the s3/200 customization applied to them?

Package.swift
```diff
@@ -44,6 +44,7 @@ extension Target.Dependency {
     static var smithyWaitersAPI: Self { .product(name: "SmithyWaitersAPI", package: "smithy-swift") }
     static var smithyTestUtils: Self { .product(name: "SmithyTestUtil", package: "smithy-swift") }
     static var smithyStreams: Self { .product(name: "SmithyStreams", package: "smithy-swift") }
+    static var smithyXml: Self { .product(name: "SmithyXML", package: "smithy-swift") }
```
sichanyoo, line 47: nit: Convention for acronyms; use `XML` instead of `Xml`

Sources/Core/AWSClientRuntime/Sources/AWSClientRuntime/Middlewares/AWSS3ErrorWith200StatusXMLMiddleware.swift
```diff
@@ -5,30 +5,25 @@
...
-        guard let data = try await response.body.readData() else {
-            return false
-        }
+    private func isRootErrorElement(data: Data) throws -> Bool {
+        let reader = try Reader.from(data: data)
 
-        response.body = .data(data)
-        let xmlString = String(decoding: data, as: UTF8.self)
-        return xmlString.contains("<Error>")
+        // Use findNode to check if there's an "Error" node in the XML
```
sichanyoo, line 25: nit: `findNode` isn't used here, is it?

Sources/Core/AWSClientRuntime/Sources/AWSClientRuntime/Middlewares/AWSS3ErrorWith200StatusXMLMiddleware.swift
```diff
@@ -40,9 +35,43 @@ extension AWSS3ErrorWith200StatusXMLMiddleware: HttpInterceptor {
...
+        // Determine if there is an error in result (500) or keep status (200)
+        let statusCode = try isRootErrorElement(data: data) ? errorStatusCode : response.statusCode
+
+        // For event streams the body needs to be copied as buffered streams are non-seekable
+        let updatedBody = response.body.copy(data: data)
+
+        let updatedResponse = response.copy(
+            body: updatedBody,
+            statusCode: statusCode
+        )
```
sichanyoo, line 58: Q: Am I understanding correctly: since the body can't be read twice, the read body is copied back in to the response for it to be available down the line in operation stack?

Sources/Core/AWSClientRuntime/Sources/AWSClientRuntime/Middlewares/AWSS3ErrorWith200StatusXMLMiddleware.swift
```diff
@@ -40,9 +35,43 @@ extension AWSS3ErrorWith200StatusXMLMiddleware: HttpInterceptor {
...
-        if try await isErrorWith200Status(response: response) {
-            response.statusCode = errorStatusCode
-            context.updateResponse(updated: response)
+
+        // Check if the status code is OK (200)
+        guard response.statusCode == .ok else {
+            return
+        }
+
+        // Check if the response body contains an XML Error
```
sichanyoo, line 44: nit: This isn't checking for error, it's just reading body

Sources/Core/AWSClientRuntime/Sources/AWSClientRuntime/Middlewares/AWSS3ErrorWith200StatusXMLMiddleware.swift
```diff
@@ -40,9 +35,43 @@ extension AWSS3ErrorWith200StatusXMLMiddleware: HttpInterceptor {
...
+        guard let data = try await response.body.readData() else {
+            return
+        }
+
+        // Determine if there is an error in result (500) or keep status (200)
```
sichanyoo, line 49: I think comments in this highlighted block can just be deleted without harm since what's being done is pretty straightforward; unnecessary comments may actually confuse if it ever gets outdated in future.

codegen/smithy-aws-swift-codegen/src/main/kotlin/software/amazon/smithy/aws/swift/codegen/customization/s3/S3ErrorWith200StatusIntegration.kt
```diff
@@ -39,10 +38,14 @@ class S3ErrorWith200StatusIntegration : SwiftIntegration {
         // Instead of playing whack-a-mole broadly apply this interceptor to everything but streaming responses
         // which adds a small amount of overhead to response processing.
         val output = ctx.model.expectShape(operationShape.output.get())
-        val outputIsNotStreaming = output.members().none {
-            it.hasTrait<StreamingTrait>() || ctx.model.expectShape(it.target).hasTrait<StreamingTrait>()
+        val outputIsNotStreamingAndBlobShape = output.members().none {
```
sichanyoo, line 41: nit: Better name might be `outputIsNotAStreamingBlobShape`. As it stands, it reads to denote a shape that is `Not Streaming Shape And Is A Blob Shape` unless I read the next 4 lines of code

sichanyoo (changes requested):
Re: comments and namings

Sources/Core/AWSClientRuntime/Sources/AWSClientRuntime/Middlewares/AWSS3ErrorWith200StatusXMLMiddleware.swift
dayaffe: Yes otherwise it will be `nil` when the user tries to look at the events

Sources/Core/AWSClientRuntime/Sources/AWSClientRuntime/Middlewares/AWSS3ErrorWith200StatusXMLMiddleware.swift
```diff
@@ -5,30 +5,25 @@
...
-            return false
-        }
+    private func isRootErrorElement(data: Data) throws -> Bool {
+        let reader = try Reader.from(data: data)
 
-        response.body = .data(data)
-        let xmlString = String(decoding: data, as: UTF8.self)
-        return xmlString.contains("<Error>")
+        // Check if there's an "Error" node at the root of the XML response
+        return reader.nodeInfo.name == "Error"
```
jbelkins, line 26: I like this, much more solid than to search a string for XML

jbelkins: But, RestXML supports error wrapping.
https://smithy.io/2.0/aws/protocols/aws-restxml-protocol.html#aws-protocols-restxml-trait

Will this code detect the error whether or not it is wrapped?

Sources/Core/AWSClientRuntime/Sources/AWSClientRuntime/Middlewares/AWSS3ErrorWith200StatusXMLMiddleware.swift
```diff
@@ -40,9 +35,41 @@ extension AWSS3ErrorWith200StatusXMLMiddleware: HttpInterceptor {
...
+        // Check if the status code is OK (200)
+        guard response.statusCode == .ok else {
+            return
+        }
+
+        guard let data = try await response.body.readData() else {
+            return
+        }
+
+        let statusCode = try isRootErrorElement(data: data) ? errorStatusCode : response.statusCode
```
jbelkins, line 48: How do we know we will have enough data here to determine if the body returns an error?
`readData()` could return zero bytes if the data is not yet available, and the stream could be filled with error data after this step is complete.

jbelkins (commented):
Questions inline

Sources/Core/AWSClientRuntime/Sources/AWSClientRuntime/Middlewares/AWSS3ErrorWith200StatusXMLMiddleware.swift
dayaffe: we don't get the stream back all at once, each response is a separate action so it should be checking each streamed message for <Error> at root

Sources/Core/AWSClientRuntime/Sources/AWSClientRuntime/Middlewares/AWSS3ErrorWith200StatusXMLMiddleware.swift
dayaffe: In short, no if it is not Error at the root it won't detect it.

sichanyoo (approved):
should be good to go pending approval from Josh also

Sources/Core/AWSClientRuntime/Sources/AWSClientRuntime/Middlewares/AWSS3ErrorWith200StatusXMLMiddleware.swift
jbelkins: OK.  Reading the SEP, it sounds pretty clear that the affected errors will be in the root.

Sources/Core/AWSClientRuntime/Sources/AWSClientRuntime/Middlewares/AWSS3ErrorWith200StatusXMLMiddleware.swift
```diff
@@ -40,9 +35,41 @@ extension AWSS3ErrorWith200StatusXMLMiddleware: HttpInterceptor {
...
+        )
+
+        context.updateResponse(updated: updatedResponse)
+    }
+}
+
+extension ByteStream {
+
+    // Copy an existing ByteStream, optionally with new data
+    public func copy(data: Data?) -> ByteStream {
```
jbelkins, line 65: Does this need to be public?  Is it used anywhere other than in the middleware above?

Sources/Core/AWSClientRuntime/Sources/AWSClientRuntime/Middlewares/AWSS3ErrorWith200StatusXMLMiddleware.swift
```diff
@@ -40,9 +35,41 @@ extension AWSS3ErrorWith200StatusXMLMiddleware: HttpInterceptor {
...
+        switch self {
+        case .data(let existingData):
+            return .data(data ?? existingData)
+        case .stream(let existingStream):
+            return .stream(data != nil ? BufferedStream(data: data, isClosed: true) : existingStream)
```
jbelkins, line 70: Shouldn't the copied stream be created un-closed?  At this point we don't know if there will be more data in the stream in the future.

Sources/Core/AWSClientRuntime/Sources/AWSClientRuntime/Middlewares/AWSS3ErrorWith200StatusXMLMiddleware.swift
jbelkins: Also, can this param be non-nil `Data`?  In the code above, `data` is safe-unwrapped before this function is called

dayaffe: sure I can make it private and non-nil unless we want to make it generally available and move it to a place like SmithyStreams or ClientRuntime. Only issue with that is neither has both ByteStream and the various types of streams

Sources/Core/AWSClientRuntime/Sources/AWSClientRuntime/Middlewares/AWSS3ErrorWith200StatusXMLMiddleware.swift
dayaffe: If its created unclosed then the AsyncThrowingStream will never complete when looping through it. It yields each event until an end event and then hangs forever.

jbelkins (approved):

PR #1650, opened 2024-07-26
chore: Remove dependency on AWSSDKIdentity unless necessary

sichanyoo:
Companion PR: https://github.com/smithy-lang/smithy-swift/pull/789

## Description of changes

- Use resolver chain for bearer token only if service actually uses HTTP bearer auth. For services that do not use bearer auth, use static token resolver instead, which is part of `SmithyIdentity` module.
- This allows removal of unnecessary dependency on `AWSSDKIdentity`.
- Also, make `DefaultBearerTokenIdentityResolverChain` be imported as `struct`

## Conventional Commits


jbelkins (commented):
When I code generate with this branch, every service still has a `AWSSDKIdentity.DefaultBearerTokenIdentityResolverChain`.  Do you see the same?

(also, a nit: modify the `DefaultBearerTokenIdentityResolverChain` symbol so that it imports as a struct.)

jbelkins (approved):
Pending green build.  Re-perform codegen in separate PR after merge.

PR #1667, opened 2024-08-07
feat: use sdk-partitions.json from trebuchet

dayaffe:
## Conventional Commits


jbelkins:
@dayaffe I have reservations about performance on this, since a file must be read from disk for every API request.  File system caching might minimize the hit though.  This might be something we can look at once performance monitoring is in place.

Package.swift
```diff
@@ -83,9 +83,13 @@ let package = Package(
                 .awsSDKHTTPAuth,
                 .awsSDKIdentity
             ],
-            path: "Sources/Core/AWSClientRuntime/Sources",
+            path: "Sources/Core/AWSClientRuntime/Sources/AWSClientRuntime",
             resources: [
+<<<<<<< HEAD
+                .process("Resources"),
```
jbelkins, line 89: lol pick this

Sources/Core/AWSClientRuntime/Sources/AWSClientRuntime/Endpoints/AWSPartitionDefinition.swift
```diff
@@ -7,232 +7,47 @@
...
-          "description": "US West (N. California)"
-        },
-        "us-west-2": {
-          "description": "US West (Oregon)"
+enum PartitionLoadingError: Error {
+    case fileNotFound
+    case readError(String)
+}
+
+public class PartitionData {
```
jbelkins, line 15: I think if you just make this type an actor, it will enforce exclusive calls to `loadAWSPartitionJSON()`.

We should not be using DispatchQueue & other GCD types in new code, except where necessary to interact with pre-Swift concurrency code.

jbelkins (changes requested):
You have unresolved merge conflicts in this branch.

Also, better to use `actor` to prevent simultaneous access than `DispatchQueue`.

Sources/Core/AWSClientRuntime/Sources/AWSClientRuntime/Endpoints/AWSPartitionDefinition.swift
dayaffe: I made PartitionDataCache an actor which is used by this and removed DispatchQueue

Issue #1683, opened 2024-08-16
Missing folder during dependency install on build system


jbelkins:
### Describe the bug

Preview builds have intermittent build failures when building `AWSSDKSwiftCLI` due to missing folders, i.e.:
```swift
warning: 'swift-llbuild': ignoring broken symlink /codebuild/output/src3290271464/src/aws-sdk-swift/AWSSDKSwiftCLI/.build/checkouts/swift-llbuild/lib/llvm/Demangle/include/llvm/Demangle/include/llvm/Demangle/include/llvm/Demangle/include/llvm/Demangle/include/llvm/Demangle/include/llvm/Demangle/include/llvm/Demangle/include/llvm/Demangle/include/llvm/Demangle/include/llvm/Demangle/include/llvm/Demangle/include/llvm/Demangle/include/llvm/Demangle/include/llvm/Demangle/include/llvm/Demangle/include/llvm/Demangle/include/llvm/Demangle/include/llvm/Demangle/include/llvm/Demangle/include/llvm/Demangle/include/llvm/Demangle/include/llvm/Demangle/include/llvm/Demangle/include/llvm/Demangle/include/llvm/Demangle/include/llvm/Demangle/include/llvm/Demangle/include/llvm/Demangle/include/llvm/Demangle/include/llvm/Demangle/include/llvm/Demangle/include/llvm/Demangle/include/llvm/Demangle/include/llvm/Demangle/include/llvm/Demangle/include/llvm/Demangle/include/llvm/Demangle/include/llvm/Demangle/include/llvm/Demangle/include/llvm/Demangle/include/llvm
error: 'swift-llbuild': The folder “ADT” doesn’t exist.

[Container] 2024/08/15 23:13:18.317969 Phase context status code: COMMAND_EXECUTION_ERROR Message: Error while executing command: swift run AWSSDKSwiftCLI test-aws-sdk ../ --batches 10. Reason: exit status 1
```

It is noted that the `swift-llbuild` dependency version in use does some weird stuff with symlinks, which are corrected in later versions.

### Expected Behavior

`AWSSDKSwiftCLI` tools should build & run every time and on all platforms.

### Current Behavior

`AWSSDKSwiftCLI` fails intermittently as described.

### Reproduction Steps

Look at build logs from previous preview builds.

### Possible Solution

`swift-llbuild` seems to be linked only because it is a dependency of `AWSSDKSwiftCLI` dependency `SwiftPM`.
[...]

PR #1684, opened 2024-08-16
fix: Replace SwiftPM dependency with local Version type

jbelkins:
## Issue \#
https://github.com/awslabs/aws-sdk-swift/issues/1683

## Description of changes
The `AWSSDKSwiftCLI` project requires the `SwiftPM` package as a dependency.  One of `SwiftPM`'s dependencies, `swift-llbuild`, has been causing failures of preview builds because of issues with symlinked folders in the project.

A review of the `SwiftPM` dependency shows that is is used exclusively for access to the `Version` structure, which represents a semantic version specifier.

To simplify the project and lighten its dependencies (including the specific dependency that is causing preview build failures), the `SwiftPM` package is removed as a dependency and a `Version` structure is provided in its place.

The `Version` structure represents a major, minor, and patch version.  It does not accept versions that do not include all three values, and it does not accept preprelease identifiers that are allowed by the semver standard.  However, this is sufficient for the current manner in which the project is versioned.

All `AWSSDKSwiftCLI` tests pass with this change.  (They previously froze because of an issue with testing Process invocations.  This PR fixes the `_runReturningStdOut()` function so that it may be mocked with a test runner.)

I have also tested this change by running the `prepare-release` CLI command on my desktop, and observing the changes made to the project and the change log that was generated.


AWSSDKSwiftCLI/Package.resolved
```diff
@@ -18,15 +18,6 @@
         "version" : "1.3.1"
       }
     },
```
jbelkins, line 20: Bye bye, useless dependencies

AWSSDKSwiftCLI/Sources/AWSCLIUtils/Process+Utils.swift
```diff
@@ -49,6 +49,14 @@ public func _run(_ process: Process) throws {
...
 
 public func _runReturningStdOut(_ process: Process) throws -> String? {
+    // If debug and we have a non-nil test runner, then use that
+    // This allows developers to intercept processes when they run to assert that it is the expected process
+    #if DEBUG
+    if let testRunner = ProcessRunner.testRunner {
+        try testRunner.run(process)
+        return nil
+    }
+    #endif
```
jbelkins, line 59: Adding this code fixes this process runner so that it can be mocked with a test runner.  Note that when the behavior is mocked, the returned output is `nil`; output may be mocked as well in the future if desired.

AWSSDKSwiftCLI/Sources/AWSCLIUtils/Version.swift
```diff
@@ -7,15 +7,46 @@
...
+            throw Error("Version does not have three components")
+        }
+        guard let major = Int(components[0]), let minor = Int(components[1]), let patch = Int(components[2]) else {
+            throw Error("Version components are not all Int")
+        }
+        self.init(major, minor, patch)
+    }
+}
+
+extension Version: CustomStringConvertible {
```
jbelkins, line 36: Like the previous `Version`, this struct provides the version string when printed.

AWSSDKSwiftCLI/Sources/AWSCLIUtils/Version.swift
```diff
@@ -7,15 +7,46 @@
 
 import Foundation
 import ArgumentParser
-import PackageDescription
 
-// MARK: - Decodable
```
jbelkins, line 12: The struct below replaces the one that was previously imported from the `SwiftPM` package.

AWSSDKSwiftCLI/Sources/AWSCLIUtils/Version.swift
```diff
@@ -7,15 +7,46 @@
...
+    }
+}
+
+extension Version: CustomStringConvertible {
+    public var description: String { versionString }
+}
+
+// MARK: - Codable
+
+extension Version: Codable {
```
jbelkins, line 42: `Version` encodes & decodes itself by encoding/decoding its version string.

AWSSDKSwiftCLI/Sources/AWSCLIUtils/Version.swift
```diff
@@ -55,9 +82,7 @@ public extension Version {
        Version(
             self.major + 1,
             0,
-            0,
-            prereleaseIdentifiers: prereleaseIdentifiers,
-            buildMetadataIdentifiers: buildMetadataIdentifiers
+            0
```
jbelkins, line 85: Prerelease & build metadata identifiers are removed, since our project does not use them.

AWSSDKSwiftCLI/Sources/AWSCLIUtils/Version.swift
```diff
@@ -7,15 +7,46 @@
...
+    public let patch: Int
+
+    private var versionString: String { "\(major).\(minor).\(patch)" }
+
+    init(_ major: Int, _ minor: Int, _ patch: Int) {
+        self.major = major
+        self.minor = minor
+        self.patch = patch
+    }
+
```
jbelkins, line 23: Note that creating a `Version` from a string is now a non-optional initializer that throws if the string is not a valid semantic version identifier.

AWSSDKSwiftCLI/Tests/AWSSDKSwiftCLITests/Commands/GeneratePackageManifestTests.swift
```diff
@@ -16,10 +16,10 @@ class GeneratePackageManifestTests: CLITestCase {
     func createPackageDependencies(
         crtVersion: String,
         clientRuntimeVersion: String
```
jbelkins, line 18: Below, many tests have been marked with `throws` since creating a `Version` from a string is now a throwing operation.

sichanyoo (approved):

PR #1699, opened 2024-08-29
feat!: Conceal serde types with Swift SPI

jbelkins:
## Issue \#
https://github.com/awslabs/aws-sdk-swift/issues/1702

## Description of changes
Conceals the public interfaces related to serde using the Swift `@_spi()` annotation.
This makes clear to customers that though public, the serde types are not intended for them to interact with directly.

Types that are now SPI:
- `SmithyReader`, `SmithyWriter` and all implementations of those types.
- All reading & writing closures, and their typealias definitions.
- All document reading & writing closures, and their typealias definitions.
- The entire `SmithyTimestamps` module.
- A couple middlewares that have serde types in their interfaces.

Also:
- The `protogen.sh` script is modified to not fail when `Package.swift` is not present in generated protocol tests.
- A module that was missing from the protocol test scheme has been added.


codegen/.swiftpm/xcode/xcshareddata/xcschemes/aws-sdk-swift-protocol-tests-Package.xcscheme
```diff
@@ -245,6 +245,20 @@
                ReferencedContainer = "container:">
             </BuildableReference>
          </BuildActionEntry>
```
jbelkins, line 247: This target was added to protocol tests in a previous change

scripts/protogen.sh
```diff
@@ -24,25 +24,25 @@ rm -rf codegen/protocol-test-codegen-local/build
 ./gradlew -p codegen/protocol-test-codegen-local build
 
 # Delete the generated Package.swift for protocol test packages so they may be seen in Xcode
```
jbelkins, line 26: `rm -f` is used below so the script does not fail if `Package.swift` is not present.

dayaffe (approved):

sichanyoo (approved):

Issue #1695, opened 2024-08-27
Add regex validation for resolvers that use session name


sichanyoo:
### Background

A ticket raised regarding STS web identity AWS credential identtiy resolver ([ticket](https://github.com/awslabs/aws-sdk-swift/issues/1694)) showed that it's unreasonably difficult to track down cause of the failure when it is caused by the user using a session name that does not match the [required regex](https://docs.aws.amazon.com/STS/latest/APIReference/API_AssumeRoleWithWebIdentity.html), as CRT does not do input parameter validation and ends up just throwing a bad request error.

### To-Do

For credential identity resolvers that use session name (SSO, STSWebIdentity, and STSAssumeRole), add input parameter validation to make it much easier to catch this for SDK users.

PR #1700, opened 2024-08-30
chore: Add regex validation for session name to relevant resolvers

sichanyoo:
## Issue \#

- https://github.com/awslabs/aws-sdk-swift/issues/1695

## Description of changes

- Add regex validation for session name to STSAssumeRole and STSWebIdentity credential resolver initializers so that error gets caught and thrown wayyyy before it does without this change (i.e., bad request exception returned from STS server)

## Conventional Commits


Sources/Core/AWSSDKIdentity/Sources/AWSSDKIdentity/AWSCredentialIdentityResolvers/STSAssumeRoleAWSCredentialIdentityResolver.swift
```diff
@@ -48,3 +50,9 @@ public struct STSAssumeRoleAWSCredentialIdentityResolver: AWSCredentialIdentityR
 }
 
 // swiftlint:enable type_name
+
+func validateString(name: String, regex: String) throws {
+    guard let _ = name.range(of: regex, options: .regularExpression) else {
```
jbelkins, line 55: `name(range:options:)` is an old method inherited from ObjC.

See here for Swift-native, modern regex support:
https://developer.apple.com/documentation/swift/regex

jbelkins (changes requested):
See my comment, you're actually using Objective-C regex support.

Sources/Core/AWSSDKIdentity/Sources/AWSSDKIdentity/AWSCredentialIdentityResolvers/STSAssumeRoleAWSCredentialIdentityResolver.swift
jbelkins: After review, Swift regex features require platform support beyond our minimums.  It's okay to keep Foundation-based regex support in place for now.

jbelkins (approved):
After reviewing docs, Swift regex support requires macOS platform support above our minimum version.

Approving with Foundation regex support as written.

Issue #1677, opened 2024-08-13
Integration test for unauthenticated AWS API calls


jbelkins:
### Describe the feature

A test is needed for unauthenticated calls to AWS services.

For an example of an unauthenticated call, see AWS Cognito Identity Provider:
https://github.com/awslabs/aws-sdk-swift/blob/main/codegen/sdk-codegen/aws-models/cognito-identity-provider.json

### Use Case

To continuously, automatically verify that unauthenticated API calls are successful.

### Acknowledgements

- [ ] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

PR #1703, opened 2024-08-30
chore: Add unauthenticated API integration test

sichanyoo:
## Issue \#

- https://github.com/awslabs/aws-sdk-swift/issues/1677

## Description of changes

- Adds an integration test for an API call that doesn't use authentication

## Conventional Commits


IntegrationTests/Services/AWSCognitoIdentityIntegrationTests/UnauthenticatedAPITests.swift
```diff
@@ -0,0 +1,51 @@
...
+        _ = try await cognitoIdentityClient.deleteIdentityPool(input: DeleteIdentityPoolInput(
+            identityPoolId: identityPoolID
+        ))
+    }
+
+    func testUnauthenticatedAPI() async throws {
+        _ = try await cognitoIdentityClient.getId(
+            input: GetIdInput(accountId: accountID, identityPoolId: identityPoolID)
+        )
+        // API call completed successfully
```
dayaffe, line 49: should we check that the deserialized response contains expected results?

sichanyoo: sure, though reaching line 49 would mean the operation succeeded without throwing error, it wouldn't hurt to check it. Adopting suggestion.

dayaffe (approved):

jbelkins (commented):
The point of this is to prove that unauthenticated requests can be performed successfully.  But the machine performing integration tests will have a set of AWS credentials configured for use in all the other tests.

How do we show that this request was generated unauthenticated, and isn't just succeeding because there happen to be credentials on the machine?

sichanyoo:
@jbelkins 

We know for sure that the operation used in this new integration test doesn't use authentication and is sending an unauthenticated request because the generated auth scheme resolver for `CognitoIdentity`'s `getId` operation returns only a single auth scheme `smithy.api#noAuth`. (see [here](https://github.com/awslabs/aws-sdk-swift/blob/c223c13e0f49693398733ae0689cbaf3e870295b/Sources/Services/AWSCognitoIdentity/Sources/AWSCognitoIdentity/AuthSchemeResolver.swift#L39-L40)). 

This means when the auth scheme gets resolved during runtime, it resolves as a "no-auth" auth scheme in `AuthSchemeMiddleware` [here](https://github.com/smithy-lang/smithy-swift/blob/5f6949b2677130730ed0062835b6e9e57e8a288c/Sources/ClientRuntime/Networking/Http/Middlewares/AuthSchemeMiddleware.swift#L55-L62), which then subsequently results in no signing occurring during `SignerMiddleware` [here](https://github.com/smithy-lang/smithy-swift/blob/5f6949b2677130730ed0062835b6e9e57e8a288c/Sources/ClientRuntime/Networking/Http/Middlewares/SignerMiddleware.swift#L31-L34).

jbelkins (approved):

jbelkins:
A check was added to the test to ensure that the HTTP `Authentication:` header is not set when making an unauthenticated request.  In response, I have approved.

Issue #1735, opened 2024-09-12
Add back maxAttempts config back to AWS config


sichanyoo:
### Background

During the client configuration rework for SRA compliance, a previously exposed optional config value for number of retries to do for a failed request got scrapped. It was part of `AWSClientConfiguration` and had the variable name `maxAttempts`, and it was added with [this PR](https://github.com/awslabs/aws-sdk-swift/pull/1062). 

But the `maxAttempts` config option did not transfer over to the new corresponding config `AWSDefaultClientConfiguration` added with [this rework PR](https://github.com/awslabs/aws-sdk-swift/pull/1312).

The developer guide for the SDK currently advises SDK users to directly manipulate the [`retryStrategyOptions`](https://github.com/smithy-lang/smithy-swift/blob/29c86d14cf1a0063593bafc55139164a80289af3/Sources/ClientRuntime/Config/DefaultClientConfiguration.swift#L14) option, but that is NOT what we want to encourage the users do. Instead, users should generally only use the [`awsRetryMode`](https://github.com/awslabs/aws-sdk-swift/blob/47d93969bcd393505c7cb757bfa3d9821da4b5de/Sources/Core/AWSClientRuntime/Sources/AWSClientRuntime/Config/AWSDefaultClientConfiguration.swift#L38) and the `maxAttempts` options. The values configured on these two fields get used automatically to resolve the `retryStrategyOptions` used by the client.

But `maxAttempts` got scrapped, which is an issue.

### To-do

Add back `maxAttempts` config by adding it to `AWSDefaultClientConfiguration`. 
Update codegen / runtime default provider logic as needed to have the configured value for `maxAttempts` take effect.

### Note

(Sorta) Tracks: https://github.com/smithy-lang/smithy-swift/issues/818
- After this config is added back, developer guide should scrap its current configure retry section and re-write it based on the two config fields (`awsRetryMode` and `maxAttempts` options) that we actually want to encourage users to use.

PR #1744, opened 2024-09-13
fix!: Retry config

sichanyoo:
## Issue \#

- https://github.com/awslabs/aws-sdk-swift/issues/1735

## Description of changes

- Adds back the previously lost config `maxAttempts` that allows SDK users to configure number of retries the SDK should make without having to directly manipulate the `retryStategyOptions` config option.
- Also: Fixes the config defaults provider and codegen for it, such that the `maxAttempts` and `awsRetryMode` config values set by the user actually get used when we resolve the default `retryStrategyOptions` config.

## Conventional Commits


Sources/Core/AWSClientRuntime/Sources/AWSClientRuntime/AWSClientConfigDefaultsProvider.swift
```diff
@@ -99,17 +99,30 @@ public class AWSClientConfigDefaultsProvider {
         return resolvedRetryMode!
```
sichanyoo, line 99: GH diff is being a bit wonky here. I just added `private static func maxAttempts`, and changed the function signature of `func retryStrategyOptions` to take in arguments (the top level config `awsRetryMode` and `maxAttempts`) at call site.

Sources/Core/AWSClientRuntime/Sources/AWSClientRuntime/Config/AWSDefaultClientConfiguration.swift
```diff
@@ -31,9 +31,19 @@ public protocol AWSDefaultClientConfiguration {
```
sichanyoo, line 31: Adds back the lost `maxAttempts` option back to config. Welcome back to our poor option that finally found its way back home.

Sources/Core/AWSClientRuntime/Sources/AWSClientRuntime/Plugins/DefaultAWSClientPlugin.swift
```diff
@@ -15,10 +15,13 @@ public class DefaultAWSClientPlugin: Plugin {
     }
```
sichanyoo, line 15: Now that defaults provider for retry strategy options actually takes in user-exposed config now, let's use em.

codegen/smithy-aws-swift-codegen/src/main/kotlin/software/amazon/smithy/aws/swift/codegen/AWSHttpProtocolServiceClient.kt
```diff
@@ -67,7 +67,7 @@ class AWSHttpProtocolServiceClient(
                     ConfigProperty(
```
sichanyoo, line 67: In initializers that take in arguments for `awsRetryMode` and `maxAttempts`, default value for retry strategy options should be defaults provider function call using those arguments. 

In initializer that doesn't take in arguments for `awsRetryMode` and `maxAttempts` but needs to set retry strategy options value to something that isn't nil (e.g., the region initializer), we can't pass in any values to defaults provider method call.

codegen/smithy-aws-swift-codegen/src/main/kotlin/software/amazon/smithy/aws/swift/codegen/config/AWSDefaultClientConfiguration.kt
```diff
@@ -39,5 +39,6 @@ class AWSDefaultClientConfiguration : ClientConfiguration {
             { it.format("\$N.retryMode()", AWSClientRuntimeTypes.Core.AWSClientConfigDefaultsProvider) },
```
sichanyoo, line 39: Adds the new config value in codegen side.

sichanyoo (commented):
Comments for reviewers

jbelkins (approved):

PR #1774, opened 2024-09-24
fix: Make manifest Swift 6 Language Mode compatible

jbelkins:
## Description of changes
Modifies the AWS SDK package manifest to not raise Swift 6 concurrency warning/errors.
- Generated code is now at the top of the manifest, right under the headers & imports.
- Generated code is now all immutable, `Sendable` globally-spaced variables.  There are no more mutable variables in the manifest.
- The package manifest is constructed by calling functions that transform the generated code to create & return the manifest components.  Since the generated code components are all immutable & `Sendable`, there is no concurrency issue with calling these functions.

The `Package.swift` in this PR is the new, re-rendered version.

Also:
- Eliminated the ability to use `main` branch of `smithy-swift` which is unused.
- Eliminated the manifest generator's `--exclude-aws-services` CLI option, which no longer has utility since integration tests & protocol tests were moved to their own Swift packages.


scripts/ci_steps/prepare_protocol_and_unit_tests.sh
```diff
@@ -2,8 +2,12 @@
 
 set -e
 
+# Get rid of the generated services
+rm -rf Sources/Services/*
+
+# Regenerate the SDK manifest, with only runtime included
 cd AWSSDKSwiftCLI
-swift run AWSSDKSwiftCLI generate-package-manifest --exclude-aws-services ..
+swift run AWSSDKSwiftCLI generate-package-manifest ..
```
jbelkins, line 10: Simply deleting the services & regenerating the manifest has the same effect as the now-deleted CLI option.

AWSSDKSwiftCLI/Sources/AWSSDKSwiftCLI/Resources/Package.Prefix.txt
```diff
@@ -0,0 +1,14 @@
+// swift-tools-version:5.9
```
jbelkins, line 1: Generated code now goes above the static content, so this separate "prefix" file is needed to provide the copyright notice, imports, etc.

AWSSDKSwiftCLI/Sources/AWSSDKSwiftCLI/Resources/Package.Base.txt
```diff
@@ -1,19 +1,6 @@
-// swift-tools-version:5.9
+// MARK: - Static Content
```
jbelkins, line 1: This static file gets rendered into the manifest after the dynamic content.

AWSSDKSwiftCLI/Sources/AWSSDKSwiftCLI/Resources/Package.Base.txt
```diff
@@ -56,15 +43,50 @@ let package = Package(
...
+        runtimeProducts +
+        serviceTargets.map(productForService(_:)),
+    dependencies:
+        [clientRuntimeDependency, crtDependency, doccDependencyOrNil].compactMap { $0 },
+    targets:
+        runtimeTargets +
+        runtimeTestTargets +
+        serviceTargets.map(target(_:)) +
+        serviceTargets.map(unitTestTarget(_:))
+)
```
jbelkins, line 56: The `Package` instance for this manifest is above.  Note that its params are either a literal or some combination of function or read-only var results.

sichanyoo (approved):

PR #1775, opened 2024-09-25
fix: Fix SDK version metadata in UserAgent header

sichanyoo:
## Issue \#

1729

## Description of changes

- During prepare release, write the new current version to `AWSClientRuntime/Resources/sdk-version.txt` as well, and use that to initialize user agent middleware during runtime similar to how partitions data is loaded.

## Conventional Commits


AWSSDKSwiftCLI/Sources/AWSSDKSwiftCLI/Commands/AWSSDKSwiftCLI/Subcommands/PrepareRelease.swift
```diff
@@ -165,6 +165,14 @@ struct PrepareRelease {
...
             try "\(newVersion)".write(toFile: "Package.version" , atomically: true, encoding: .utf8)
             log("Updated Package.version: \(newVersion)")
+            if (repoType == .awsSdkSwift) {
+                try "\(newVersion)".write(
+                    toFile: "Sources/Core/AWSClientRuntime/Sources/AWSClientRuntime/Resources/sdk-version.txt",
+                    atomically: true,
+                    encoding: .utf8
+                )
+                log("Updated AWSClientRuntime/Resources/Package.version: \(newVersion)")
+            }
```
dayaffe, line 175: **Non-Blocking:** This may be the best solution but it gives me pause that we will be storing a duplicate of `package.version` in a `sdk-version.txt`. I understand the reasons for doing so but I wonder if there is either a better way or an opportunity for refactoring so that we only have a single file with the version.

jbelkins (changes requested):
I don't want to mix auto-generated files into `AWSClientRuntime`.  That complicates our build/release pipeline.

I thought we had discussed using the contents of `Package.version.next` to write the version string into the client during codegen.  Did that approach not work?

sichanyoo:
@jbelkins 
I decided to take this route instead of using `Package.version.next` during codegen, after finding out that reading a file in Kotlin that's not placed under `resources/` is tricky / dangerous. You can give it relative path or an absolute path of the filet o read. Relative path gets resolved against where-ever the program was initiated from (e.g., from `~` if it was initiated from there, instead of project root). And as for absolute path, we don't have that available during codegen.

sichanyoo:
Work taken over by @jbelkins. Closing PR after offline discussion.

PR #1780, opened 2024-09-27
feat: Service client release notes

sichanyoo:
## Issue \#

1734

### Current
Currently, the GitHub release notes that get generated with each new release only contains information about commits that got merged to main since last release. In addition, it aggregates all commits into release notes, even ones that are irrelevant to customers (such as chore commits or commits from automated user for updating models / endpoints).
### Goal
We want to filter those commits out from the release notes, as well as include specific service client changes in the release notes. 
### Changes
In the build system, at any point in time, we have access to two JSON files that we can extract this info from.
This PR's changes use those two JSON files to generate enhanced release notes for GitHub releases.

The changes were tested by running an E2E test against dev stack.

*See comments on GH diff for more details.

## Conventional Commits


sichanyoo (commented):
Comments for reviewers

AWSSDKSwiftCLI/Sources/AWSSDKSwiftCLI/Models/Features.swift
```diff
@@ -0,0 +1,72 @@
...
+        var fileContents: Data
+        do {
+            fileContents = try FileManager.default.loadContents(atPath: filePath)
+        } catch {
+            throw Error("Failed to load build-request.json.")
+        }
+        do {
+            return try JSONDecoder().decode(Self.self, from: fileContents)
+        } catch {
+            throw Error("Failed to decode build-request.json.")
```
jbelkins, line 24: Instead of throwing this error which hides the underlying problem, why not just let
```swift
try JSONDecoder().decode(Self.self, from: fileContents)
```
throw its own error?

AWSSDKSwiftCLI/Sources/AWSSDKSwiftCLI/Models/Features.swift
```diff
@@ -0,0 +1,72 @@
...
+        enum FeatureMetadataKeys: String, CodingKey {
+            case trebuchet
+
+            enum TrebuchetKeys: String, CodingKey {
+                case featureId, featureType
+            }
+        }
+    }
+
+    init(from decoder: Decoder) throws {
```
jbelkins, line 62: Rather than write custom encoders/decoders, I'd just create Swift structures that match the JSON schema and let auto-Codable conformance do the hard work.  The sample JSON:
```json
    {
        "releaseNotes": "Doc update description C.",
        "featureMetadata": {
            "trebuchet": {
                "featureId": "feature-id-c",
                "featureType": "DOC_UPDATE",
            }
        }
    }
```
translates to (any of these fields can be marked optional if warranted):
```swift
struct Feature: Codable {
    let releaseNotes: String
    let featureMetadata: FeatureMetadata
    
    struct FeatureMetadata: Codable {
        let trebuchet: Trebuchet

        struct Trebuchet: Codable {
            let featureId: String
            let featureType: String
        }
    }
}
```
Swift will do all the encoding/decoding for you by generating `init(from:) throws` and `func encode(to:) throws` for each of these structs.

AWSSDKSwiftCLI/Tests/AWSSDKSwiftCLITests/Models/PackageManifestBuilderTests.swift
```diff
@@ -12,7 +12,6 @@ class PackageManifestBuilderTests: XCTestCase {
 
     let expected = """
 <contents of prefix>
-
```
jbelkins, line 15: Why is this test expectation changed?  This feature doesn't involve the manifest.

AWSSDKSwiftCLI/Sources/AWSSDKSwiftCLI/Models/ReleaseNotesBuilder.swift
```diff
@@ -15,24 +15,73 @@ struct ReleaseNotesBuilder {
...
-    func build() -> String {
-        let contents = [
-            "## What's Changed",
-            buildCommits(),
-            .newline,
-            "**Full Changelog**: https://github.com/\(repoOrg.rawValue)/\(repoType.rawValue)/compare/\(previousVersion)...\(newVersion)"
+    func build() throws -> String {
+        let baseContent: [String] = (
+            ["## What's Changed"] +
+            buildSDKChangeSection()
```
jbelkins, line 25: Should we maybe put SDK changes below service-related changes?

AWSSDKSwiftCLI/Sources/AWSSDKSwiftCLI/Models/ReleaseNotesBuilder.swift
```diff
@@ -15,24 +15,73 @@ struct ReleaseNotesBuilder {
...
+        return []
+    }
+
+    func buildServiceChangeSection() throws -> [String] {
+        var features: Features
+        var mapping: [String: String]
+        // At this point, FileManager.default's current working directory is aws-sdk-swift/.
+        // The JSON files we need are located one level above it, in the workspace directory.
+        // For tests, due to sandboxing, the dummy files are created in current directory instead of
+        //  in parent directory. Therefore, fetch them from current directory for tests.
```
jbelkins, line 52: Better way to do this is with dependency injection: there's a "live" object that creates the file locations for when ReleaseNotesBuilder and a "test" object that provides file locations for tests.

One or the other of these objects is passed into ReleaseNotesBuilder at creation (or, the "live" one can be used as a default param.)

These would take the place of the static methods `Features.fromFile(_:)` and `Features.mappingFromFile(_:)` below.

jbelkins (changes requested):

AWSSDKSwiftCLI/Sources/AWSSDKSwiftCLI/Models/Features.swift
sichanyoo: Good point; checked the original error thrown does print exact issue. Will do this for other catch clauses as well.

AWSSDKSwiftCLI/Sources/AWSSDKSwiftCLI/Models/Features.swift
sichanyoo: Changed as suggested; regression tests pass with this change.

AWSSDKSwiftCLI/Tests/AWSSDKSwiftCLITests/Models/PackageManifestBuilderTests.swift
sichanyoo: The test was failing; I don't think this test was being run as part of CI with the package refactor task from before.

AWSSDKSwiftCLI/Sources/AWSSDKSwiftCLI/Models/ReleaseNotesBuilder.swift
sichanyoo: Sure, that might make more sense since most customers would just care about service client changes. Changed as suggested.

AWSSDKSwiftCLI/Sources/AWSSDKSwiftCLI/Models/ReleaseNotesBuilder.swift
sichanyoo: Adopted suggestion; the behavior is determined by paths saved in-memory of an instance of `FeaturesReader` object that gets passed to `ReleaseNotesBuilder` initializer, with default behavior using non-test file locations.

jbelkins (approved):

PR #1803, opened 2024-10-24
feat!: Flexible checksum v2

sichanyoo:
Corresponding PR: https://github.com/smithy-lang/smithy-swift/pull/885

## Issue \#

1516

## Conventional Commits


Sources/Core/AWSClientRuntime/Sources/AWSClientRuntime/AWSClientConfigDefaultsProvider.swift
```diff
@@ -84,6 +85,40 @@ public class AWSClientConfigDefaultsProvider {
         return resolvedAppID
     }
 
+    public static func requestChecksumCalculation(
```
dayaffe, line 88: would be good to add comment strings explaining what these functions do / how theyre used

Sources/Core/AWSSDKChecksums/Sources/AWSSDKChecksums/AWSChecksumCalculationMode.swift
```diff
@@ -0,0 +1,24 @@
+//
+// Copyright Amazon.com Inc. or its affiliates.
+// All Rights Reserved.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+public enum AWSChecksumCalculationMode: String {
```
dayaffe, line 8: I like this idea of whenRequired vs whenSupported

sichanyoo: This addition to config is part of the V2 SEP 👍

Sources/Core/AWSClientRuntime/Sources/AWSClientRuntime/Middlewares/FlexibleChecksumsResponseMiddleware.swift
```diff
@@ -31,13 +32,15 @@ public struct FlexibleChecksumsResponseMiddleware<OperationStackInput, Operation
...
-        if !validationMode {
+        if validationMode != "ENABLED" && attributes.responseChecksumValidation == .whenRequired {
             logger.info("Checksum validation should not be performed! Skipping workflow...")
             return
         }
 
         let checksumHeaderIsPresent = priorityList.first {
-            response.headers.value(for: "x-amz-checksum-\($0)") != nil
+            response.headers.value(for: "x-amz-checksum-\($0)") != nil  &&
+            // Checksum of checksums has "-#" at the end and should be ignored.
```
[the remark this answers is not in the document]
sichanyoo: Not necessarily a new requirement, it was just something that was left out from V1 SEP. S3 can return checksum of checksum, which ends with `-#` and should be ignored when validating response checksum.

Sources/Core/AWSClientRuntime/Sources/AWSClientRuntime/AWSClientConfigDefaultsProvider.swift
sichanyoo: These are essentially for default behavior in SDK side; I don't think any SDK user would have to / want to use these directly.

PR #1818, opened 2024-11-12
chore: Role for STS web identity integration test is assumed based on identity pool ID

jbelkins:
## Description of changes
Requester must be associated with the identity pool for this test in order to assume the role.

### DIFF
--- a/IntegrationTests/Services/AWSSTSIntegrationTests/STSWebIdentityAWSCredentialIdentityResolverTests.swift
+++ b/IntegrationTests/Services/AWSSTSIntegrationTests/STSWebIdentityAWSCredentialIdentityResolverTests.swift
@@ -42,13 +42,33 @@ class STSWebIdentityAWSCredentialIdentityResolverTests: XCTestCase {
     private let roleName = "aws-sts-integration-test-\(UUID().uuidString.split(separator: "-").first!.lowercased())"
     private let roleSessionName = "aws-sts-integration-test-\(UUID().uuidString.split(separator: "-").first!.lowercased())"
     private var roleArn: String!
+
     // JSON assume role policy
-    private let assumeRolePolicy = """
-    {"Version": "2012-10-17","Statement": [{"Sid": "","Effect": "Allow",
-    "Principal": {"Federated": "cognito-identity.amazonaws.com"},"Action": [
-    "sts:AssumeRoleWithWebIdentity"],"Condition": {"ForAnyValue:StringLike": {
-    "cognito-identity.amazonaws.com:amr": "unauthenticated"}}}]}
-    """
+    private var assumeRolePolicy: String {
+        return """
+        {
+            "Version": "2012-10-17",
+            "Statement": [
+                {
+                    "Sid": "",
+                    "Effect": "Allow",
+                    "Principal": {
+                        "Federated": "cognito-identity.amazonaws.com"
+                    },
+                    "Action": [
+                        "sts:AssumeRoleWithWebIdentity"
+                    ],
+                    "Condition": {
+                        "ForAnyValue:StringLike": {
+                            "cognito-identity.amazonaws.com:aud": "\(identityPoolId!)"
+                        }
+                    }
+                }
+            ]
+        }
+        """
+    }
+
     private let identityPolicyName = "allow-STS-getCallerIdentity"
     // JSON identity policy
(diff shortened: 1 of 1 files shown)


IntegrationTests/Services/AWSSTSIntegrationTests/STSWebIdentityAWSCredentialIdentityResolverTests.swift
```diff
@@ -42,13 +42,33 @@ class STSWebIdentityAWSCredentialIdentityResolverTests: XCTestCase {
...
+                    "Effect": "Allow",
+                    "Principal": {
+                        "Federated": "cognito-identity.amazonaws.com"
+                    },
+                    "Action": [
+                        "sts:AssumeRoleWithWebIdentity"
+                    ],
+                    "Condition": {
+                        "ForAnyValue:StringLike": {
+                            "cognito-identity.amazonaws.com:aud": "\(identityPoolId!)"
```
jbelkins, line 63: Besides formatting, the only change to this policy is the line immediately above, where audience is used as the condition for assuming the role instead of authenticated / unauthenticated status.

IntegrationTests/Services/AWSSTSIntegrationTests/STSWebIdentityAWSCredentialIdentityResolverTests.swift
```diff
@@ -61,15 +81,15 @@ class STSWebIdentityAWSCredentialIdentityResolverTests: XCTestCase {
     override func setUp() async throws {
         try await super.setUp()
 
+        // Create the Cognito identity pool that allows unauthenticated identities
+        try await createCognitoIdentityPool()
+
```
jbelkins, line 86: The identity pool has to be created before the role is created, so that the identity pool ID can be specified in the assume role policy.  Hence, the order of steps is changed.

IntegrationTests/Services/AWSSTSIntegrationTests/STSWebIdentityAWSCredentialIdentityResolverTests.swift
```diff
@@ -82,24 +102,26 @@ class STSWebIdentityAWSCredentialIdentityResolverTests: XCTestCase {
...
-        _ = try await cognitoIdentityClient.deleteIdentityPool(
+        _ = try? await cognitoIdentityClient.deleteIdentityPool(
             input: DeleteIdentityPoolInput(identityPoolId: identityPoolId)
         )
 
         // Delete token file
-        try deleteTokenFile()
+        try? deleteTokenFile()
+
+        try await super.tearDown()
```
jbelkins, line 118: `super` is called during tear down.

Also, `try?` is used on throwing teardown operations so that the remainder of teardown may proceed if one part fails.

IntegrationTests/Services/AWSSTSIntegrationTests/STSWebIdentityAWSCredentialIdentityResolverTests.swift
```diff
@@ -82,24 +102,26 @@ class STSWebIdentityAWSCredentialIdentityResolverTests: XCTestCase {
...
     }
 
     // MARK: - TEST CASE
 
     // Confirm STS web identity credentials provider works by validating response.
-    func xtestGetCallerIdentity() async throws {
+    func testGetCallerIdentity() async throws {
```
jbelkins, line 124: Test is now re-enabled

IntegrationTests/Services/AWSSTSIntegrationTests/STSWebIdentityAWSCredentialIdentityResolverTests.swift
sichanyoo: Question: What's the point of `super.tearDown()` here? Isn't it just a no-op in `XCTestCase`?

jbelkins: In general I call `super` in overrides unless I'm explicitly choosing to not call `super`; i.e. replacing & not just augmenting the current behavior.

(FWIW not calling `super` isn't wrong, but it's saved me trouble on refactors in the past)

sichanyoo (approved):

PR #1822, opened 2024-11-14
feat: add toggle to SigV4AuthScheme to turn off body signing

dayaffe:
## Issue \#
SWIFT-1849, PR: https://github.com/smithy-lang/smithy-swift/pull/857

## Description of changes
- Add signing property `requestUnsignedBody`. Note this wont effect middlewares or anything else on purpose until we go to actually sign
- In AWSSigV4Signer before we go to sign, check for requestUnsignedBody and if its true, overrule isUnsignedBody, setting header `x-amz-content-sha256` to `UNSIGNED-PAYLOAD`

## Usage
```
let config = try await S3Client.S3ClientConfiguration(
    authSchemes: [SigV4AuthScheme(requestUnsignedBody: true)]
)
let client = S3Client(config: config)
```

## Conventional Commits


IntegrationTests/Services/AWSS3IntegrationTests/S3ToggleUnsignedTests.swift
```diff
@@ -0,0 +1,45 @@
+//
```
sichanyoo, line 1: Please add:
- One more test case that has payload over the chunked threshold
- An interceptor that asserts that:
  - `Authorization` header isn't present as expected
  -  `x-amz-content-sha256` header values match expected values (`UNSIGNED-PAYLOAD` for single chunk request, and `STREAMING-UNSIGNED-PAYLOAD-TRAILER` for aws chunked request)

sichanyoo (changes requested):
Comments re: test

IntegrationTests/Services/AWSS3IntegrationTests/S3ToggleUnsignedTests.swift
dayaffe: I believe `Authorization` header is still expected to be there. We are still signing the request just not the body. Will do the rest

sichanyoo: That's true, and gotcha

sichanyoo (changes requested):
nit: Change integration test file name and test cases to mean unsigned payload, not unsigned request

Sources/Core/AWSSDKHTTPAuth/Tests/AWSSDKHTTPAuthTests/SigV4AuthSchemeTests.swift
```diff
@@ -283,4 +283,17 @@ class SigV4AuthSchemeTests: XCTestCase {
...
+        let customSigV4AuthScheme = SigV4AuthScheme(requestUnsignedBody: true)
+        let context = contextBuilder
+            .withBidirectionalStreamingEnabled(value: true)
+            .withServiceName(value: "filler")
+            .withFlowType(value: .NORMAL)
+            .withOperation(value: "filler")
+            .withUnsignedPayloadTrait(value: false)
+            .build()
+        let updatedProperties = try customSigV4AuthScheme.customizeSigningProperties(signingProperties: Attributes(), context: context)
+        XCTAssertTrue(try XCTUnwrap(updatedProperties.get(key: SigningPropertyKeys.requestUnsignedBody)))
```
jbelkins, line 297: Maybe break the XCTUnwrap & XCAssert into two lines for clarity

Sources/Core/AWSSDKHTTPAuth/Sources/AWSSDKHTTPAuth/SigV4AuthScheme.swift
```diff
@@ -15,9 +15,14 @@ import struct Smithy.Attributes
 public struct SigV4AuthScheme: AuthScheme {
     public let schemeID: String = "aws.auth#sigv4"
     public let signer: Signer = AWSSigV4Signer()
+    public var requestUnsignedBody: Bool?
```
jbelkins, line 18: The possible values for `Bool?` are `true`, `false`, and `nil`.

`true` means unsigned body is requested.
`nil` means unsigned body was not specified in the auth scheme resolver.

So what is `false`?  I can certainly create a auth scheme resolver with `requestUnsignedBody: false`.  Would that ever be used?  Is it undefined?  (If it doesn't, we should consider a different data type than `Bool?` for this setting.)

jbelkins (changes requested):

IntegrationTests/Services/AWSS3IntegrationTests/S3ToggleUnsignedTests.swift
```diff
@@ -0,0 +1,107 @@
...
+        let fetchedObject = try await client.getObject(input: getObjectInput)
+
+        XCTAssertNotNil(fetchedObject.metadata)
+        let metadata = try XCTUnwrap(fetchedObject.metadata)
+        XCTAssertEqual(metadata["filename"], key)
+    }
+
+    func testS3ToggleUnsignedRequestStreaming() async throws {
+        let key = "test-streaming.txt"
+        let data = Data((0..<(1024 * 1024)).map { _ in UInt8.random(in: UInt8.min...UInt8.max) })
```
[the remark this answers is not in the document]
dayaffe: Sure I can move this out

IntegrationTests/Services/AWSS3IntegrationTests/S3ConcurrentTests.swift
```diff
@@ -18,38 +18,27 @@ final class S3ConcurrentTests: S3XCTestCase {
     // Payload just below chunked threshold
     // Tests concurrent upload of simple data payloads
     func test_10x_1MB_getObject() async throws {
-        fileData = try generateDummyTextData(count: CHUNKED_THRESHOLD - 1)
+        fileData = S3ConcurrentTests.generateRandomTextData(ofSizeInMB: 1)
```
jbelkins, line 21: There is a folder `IntegrationTests/AWSIntegrationTestUtils` where I think this method belongs

IntegrationTests/Services/AWSS3IntegrationTests/S3ConcurrentTests.swift
```diff
@@ -18,38 +18,27 @@ final class S3ConcurrentTests: S3XCTestCase {
...
-        fileData = try generateDummyTextData(count: CHUNKED_THRESHOLD)
+        fileData = S3ConcurrentTests.generateRandomTextData(ofSizeInMB: 1.5)
         try await repeatConcurrentlyWithArgs(count: 10, test: getObject, args: fileData!)
     }
 
     // Payload 256 bytes with 200 concurrent requests, sends as simple data
     // Tests very high concurrency with small data payloads
     func test_200x_256B_getObject() async throws {
-        fileData = try generateDummyTextData(count: 256)
+        let bytes_256 = Double(256) / (1024 * 1024)
```
jbelkins, line 35: Any Swift literal with a decimal point is a Double.  So you can just use `256.0` or `256.` instead of the Double initializer.

IntegrationTests/Services/AWSS3IntegrationTests/S3XCTestCase.swift
```diff
@@ -116,4 +116,12 @@ class S3XCTestCase: XCTestCase {
         let input = DeleteBucketInput(bucket: bucketName)
         _ = try await client.deleteBucket(input: input)
     }
+
+    static func generateRandomTextData(ofSizeInMB megabytes: Double) -> Data {
```
jbelkins, line 120: This method should take a precise integer number of bytes.

I'm okay with it if we have a convenience form that takes a Double number of MB then converts that to bytes.

Also, as mentioned above, let's move this to integration test utils.

PR #1839, opened 2024-12-03
feat: Add integration test for STS assume role credentials provider.

sichanyoo:
## Issue \#

2140

## Description of changes

- Adds integration test for `STSAssumeRoleAWSCredentialIdentityResolver`

## Conventional Commits


IntegrationTests/Services/AWSSTSIntegrationTests/STSAssumeRoleAWSCredentialIdentityResolverTests.swift
```diff
@@ -0,0 +1,130 @@
...
+                    "Effect": "Allow",
+                    "Principal": {
+                        "AWS": "arn:aws:iam::\(acctID):root"
+                    },
+                    "Action": "sts:AssumeRole"
+                }
+            ]
+        }
+        """
+    }
```
dayaffe, line 46: nit: consider extract this out to a helper function since assume role is also used in IMDS integration test

IntegrationTests/Services/AWSSTSIntegrationTests/STSAssumeRoleAWSCredentialIdentityResolverTests.swift
```diff
@@ -0,0 +1,130 @@
...
+
+    private let identityPolicyName = "temp-allow-STS-getCallerIdentity"
+    // JSON identity policy
+    private let roleIdentityPolicy = """
+    {"Version": "2012-10-17","Statement": [{"Sid": "","Effect": "Allow",
+    "Action": "sts:GetCallerIdentity","Resource": "*"}]}
```
dayaffe, line 52: nit: consider creating a constant and moving these to another file

IntegrationTests/Services/AWSSTSIntegrationTests/STSAssumeRoleAWSCredentialIdentityResolverTests.swift
```diff
@@ -0,0 +1,130 @@
...
+
+        // Ensure returned caller info aren't nil
+        let account = try XCTUnwrap(response.account)
+        let userId = try XCTUnwrap(response.userId)
+        let arn = try XCTUnwrap(response.arn)
+
+        // Ensure returned caller info aren't empty strings
+        XCTAssertNotEqual(account, "")
+        XCTAssertNotEqual(userId, "")
+        XCTAssertNotEqual(arn, "")
```
dayaffe, line 71: Question: Can we do better and assert what they should be?

dayaffe (approved):

IntegrationTests/Services/AWSSTSIntegrationTests/STSAssumeRoleAWSCredentialIdentityResolverTests.swift
sichanyoo: [One in IMDS integ test](https://github.com/awslabs/aws-sdk-swift/blob/143d6938a2ec4997efd764daca2f57d70933a6cd/IntegrationTests/Services/AWSEC2IntegrationTests/IMDSAWSCredentialIdentityResolverTests.swift#L105-L110) is slightly different bc that one allows EC2 principal rather than account principal. I think it probably will require a function that needs multiple arguments with messy string concat logic just to handle the two cases.

But if more and more integration tests have this sort of duplicates, we can then consider refactoring integration test suite a bit and adding some utility class that can be used by any test for common uses like these.

IntegrationTests/Services/AWSSTSIntegrationTests/STSAssumeRoleAWSCredentialIdentityResolverTests.swift
sichanyoo: For now this identity policy is only used by this test, so it makes sense to have it in same file as the tests for cohesion. When new integration tests down the line use identical policies, then we can consider refactoring it.

PR #1845, opened 2024-12-10
fix: Fix signing region source

sichanyoo:
## Description of changes

- Get signing region from resolved auth option first, instead of getting it from the context first. This removes the need to override signing region in the endpoint resolver middleware as well.
- Below outlines how signing region is resolved and used for rules based auth:
  - `signingRegionSet` property is taken from the endpoint resolved from endpoint rules engine
  - If the property was missing, use signing region configured on request context instead (middleware context / interceptor context)
  - If the property was present, use the first value in that list to sign the request

## Conventional Commits


dayaffe:
Confirmed this solves a known issue in multi-region authentication with sigv4a

dayaffe (approved):

jbelkins (approved):

PR #1850, opened 2024-12-24
feat: Support RPCv2 CBOR wire protocol

dayaffe:
All protocol tests passing ✅

## Conventional Commits


Sources/Core/AWSClientRuntime/Sources/AWSClientRuntime/Protocols/RpcV2Cbor/RpcV2CborError.swift
```diff
@@ -0,0 +1,85 @@
...
+    public let requestID: String?
+    @_spi(SmithyReadWrite) public var errorBodyReader: Reader { responseReader }
+
+    public let httpResponse: HTTPResponse
+    private let responseReader: Reader
+
+    @_spi(SmithyReadWrite)
+    public init(httpResponse: HTTPResponse, responseReader: Reader, noErrorWrapping: Bool, code: String? = nil) throws {
+
+        // Improve this code
```
dayaffe, line 25: will take a second look at this section and see if code can be improved

Sources/Core/AWSClientRuntime/Sources/AWSClientRuntime/UserAgent/BusinessMetrics.swift
```diff
@@ -103,4 +106,8 @@ private func setFlagsIntoContext(
     if context.selectedAuthScheme?.schemeID == "aws.auth#sigv4a" {
         context.businessMetrics = ["SIGV4A_SIGNING": "S"]
     }
+    // Handle M
+    if headers.value(for: "smithy-protocol") == "rpc-v2-cbor" {
```
dayaffe, line 110: enable business metrics tracking based on header in request

codegen/smithy-aws-swift-codegen/src/main/kotlin/software/amazon/smithy/aws/swift/codegen/AWSServiceUtils.kt
```diff
@@ -7,24 +7,25 @@ package software.amazon.smithy.aws.swift.codegen
...
-import software.amazon.smithy.swift.codegen.model.expectTrait
+import software.amazon.smithy.swift.codegen.model.getTrait
 
 /**
  * Get the [sdkId](https://smithy.io/2.0/aws/aws-core.html#sdkid) from the (AWS) service shape
+ * or return a default value if the trait is not present.
  */
 val ServiceShape.sdkId: String
-    get() = expectTrait<ServiceTrait>().sdkId
+    get() = getTrait<ServiceTrait>()?.sdkId ?: "defaultSdkId"
```
dayaffe, line 17: default values are necessary since not all protocol tests will be tied to a particular AWS service. This is the first set of smithy files we need to codegen where there is no service trait

codegen/smithy-aws-swift-codegen/src/main/kotlin/software/amazon/smithy/aws/swift/codegen/protocols/rpcv2cbor/RPCV2CBORHttpBindingResolver.kt
```diff
@@ -0,0 +1,34 @@
...
+            .method("POST")
+            .uri(UriPattern.parse("/"))
+            .build()
+    }
+
+    override fun httpTrait(operationShape: OperationShape): HttpTrait = HttpTrait
+        .builder()
+        .code(200)
+        .method("POST")
+        .uri(UriPattern.parse("/service/${getServiceName()}/operation/${operationShape.id.name}"))
```
dayaffe, line 32: SEP requires particular URI pattern

codegen/smithy-aws-swift-codegen/src/main/kotlin/software/amazon/smithy/aws/swift/codegen/AWSServiceUtils.kt
sichanyoo: Is this because RPCv2 CBOR protocol tests are run in aws-sdk-swift instead of smithy-swift? Should these be included in the RPC protocol relocation ticket you create as changes that need to be reverted?

codegen/smithy-aws-swift-codegen/src/main/kotlin/software/amazon/smithy/aws/swift/codegen/AWSSmokeTestGenerator.kt
```diff
@@ -26,11 +26,15 @@ class AWSSmokeTestGenerator(
...
+        val serviceTrait = ctx.service.getTrait(ServiceTrait::class.java).orElse(null)
+        val sdkId = serviceTrait?.sdkId?.toUpperCamelCase() ?: "DefaultService"
+        return "AWS$sdkId"
     }
 
     override fun getClientName(): String {
-        return ctx.service.getTrait(ServiceTrait::class.java).get().sdkId.toUpperCamelCase().removeSuffix("Service") + "Client"
+        val serviceTrait = ctx.service.getTrait(ServiceTrait::class.java).orElse(null)
+        val sdkId = serviceTrait?.sdkId?.toUpperCamelCase()?.removeSuffix("Service") ?: "Default"
+        return "${sdkId}Client"
```
sichanyoo, line 37: I think it's far from ideal to have these changes just to accommodate codegen'ing for protocol test model that doesn't have sdkId. Did you create the follow-up ticket for re-locating RPCv2 CBOR protocol as discussed last week btw?

codegen/smithy-aws-swift-codegen/src/main/kotlin/software/amazon/smithy/aws/swift/codegen/protocols/rpcv2cbor/RPCV2CBORHttpBindingResolver.kt
sichanyoo: SEP includes `{prefix?}` portion before `/service`; it doesn't affect routing, but nevertheless can be configured by services for internal use / metrics. Might need to look into where that comes from in the model and add accordingly.

codegen/smithy-aws-swift-codegen/src/main/kotlin/software/amazon/smithy/aws/swift/codegen/protocols/rpcv2cbor/RpcV2CborCustomizations.kt
```diff
@@ -0,0 +1,15 @@
...
+import software.amazon.smithy.aws.swift.codegen.swiftmodules.AWSClientRuntimeTypes
+import software.amazon.smithy.codegen.core.Symbol
+import software.amazon.smithy.model.traits.TimestampFormatTrait
+
+class RpcV2CborCustomizations : AWSHTTPProtocolCustomizations() {
+
+    override val baseErrorSymbol: Symbol = AWSClientRuntimeTypes.RpcV2Cbor.RpcV2CborError
+
+    // Timestamp format is not used in RpcV2Cbor since it's a binary protocol. We seem to be missing an abstraction
+    // between text-based and binary-based protocols
```
sichanyoo, line 13: Based on SEP wording "`The [timestampFormat](https://smithy.io/2.0/spec/protocol-traits.html#timestampformat-trait) MUST NOT be respected to customize timestamp serialization.`", I assume this is the intended implementation

codegen/smithy-aws-swift-codegen/src/main/kotlin/software/amazon/smithy/aws/swift/codegen/protocols/rpcv2cbor/RpcV2CborProtocolGenerator.kt
```diff
@@ -0,0 +1,90 @@
...
+    override val shouldRenderEncodableConformance = true
+
+    override fun getProtocolHttpBindingResolver(ctx: ProtocolGenerator.GenerationContext, defaultContentType: String):
+        HttpBindingResolver = RPCV2CBORHttpBindingResolver(ctx, defaultContentType)
+
+    override fun addProtocolSpecificMiddleware(ctx: ProtocolGenerator.GenerationContext, operation: OperationShape) {
+        super.addProtocolSpecificMiddleware(ctx, operation)
+
+        // Original instance of OperationInputBodyMiddleware checks if there is an HTTP Body, but for AWSJson protocols
+        // we always need to have an InputBodyMiddleware
```
sichanyoo, line 33: Possibly irrelevant comment from copy-paste?

codegen/smithy-aws-swift-codegen/src/main/kotlin/software/amazon/smithy/aws/swift/codegen/protocols/rpcv2cbor/RpcV2CborProtocolGenerator.kt
```diff
@@ -0,0 +1,90 @@
...
+        // Every request for the rpcv2Cbor protocol MUST contain a smithy-protocol header with the value of rpc-v2-cbor
+        val smithyProtocolRequestHeaderMiddleware = MutateHeadersMiddleware(
+            overrideHeaders = mapOf(
+                "smithy-protocol" to "rpc-v2-cbor",
+                "Accept" to acceptHeaderValue
+            )
+        )
+
+        operationMiddleware.appendMiddleware(operation, smithyProtocolRequestHeaderMiddleware)
+        operationMiddleware.appendMiddleware(operation, CborValidateResponseHeaderMiddleware)
```
sichanyoo, line 64: I think it would be great to have general purpose middleware for verifying headers instead of RPCv2 CBOR specific middleware. Similar to how the `MutateHeadersMiddleware` allows passing in multiple headers to add to request, there could be a middleware called something along the lines of `VerifyHeadersMiddleware` that verifies passed in headers; you could pass in `smithy-protocol : rpc-v2-cbor` to it to serve the specific purpose here.

Sources/Core/AWSClientRuntime/Sources/AWSClientRuntime/Middlewares/RpcV2CborValidateResponseHeaderMiddleware.swift
```diff
@@ -0,0 +1,45 @@
+//
```
sichanyoo, line 1: Why does this only check `smithy-protocol` header and not `Content-Type`, `Content-Length` and `x-amzn-ErrorType` headers? Not suggesting they need to be checked, but rather asking why only one header was singled out to require verification on SDK side

sichanyoo (commented):
Mostly Qs

codegen/smithy-aws-swift-codegen/src/main/kotlin/software/amazon/smithy/aws/swift/codegen/AWSServiceUtils.kt
dayaffe: nope nothing to do with that -- the protocol tests themselves (smithy files) do not tie these tests to a particular AWS service (no service trait). Smithy team did that on purpose because it is not an "aws" protocol

codegen/smithy-aws-swift-codegen/src/main/kotlin/software/amazon/smithy/aws/swift/codegen/AWSSmokeTestGenerator.kt
dayaffe: relocating RPCv2 CBOR to smithy-swift wouldn't change the need for this. Its because the protocol tests themselves do not have a service trait defined. There is a ticket for relocating RPC to smithy-swift but that change should be purely cosmetic as I mentioned.

codegen/smithy-aws-swift-codegen/src/main/kotlin/software/amazon/smithy/aws/swift/codegen/protocols/rpcv2cbor/RPCV2CBORHttpBindingResolver.kt
dayaffe: It's not part of what we need to worry about constructing in the URI - it's things like if the endpoint is http://example.com/baz, the URL is http://example.com/baz/service/...

(Confirmed with Smithy team)

codegen/smithy-aws-swift-codegen/src/main/kotlin/software/amazon/smithy/aws/swift/codegen/protocols/rpcv2cbor/RpcV2CborCustomizations.kt
dayaffe: will change or remove comment

Sources/Core/AWSClientRuntime/Sources/AWSClientRuntime/Middlewares/RpcV2CborValidateResponseHeaderMiddleware.swift
dayaffe: This section of the SEP:

```
Operation response deserialization
Clients MUST use the following rules to interpret responses and determine how to deserialize them:

1. If the response does not have the same smithy-protocol header as the request, it MUST be considered malformed. Clients who receive a malformed response MUST handle it (i.e. throw a reasonable exception) based solely on the HTTP response code. No attempt should be made to interpret the response body or headers.
2. If the response code is 200, the request is successful and the response payload SHALL be deserialized as the output shape defined on the operation.
3. Finally, if the response code is not 200, the response payload is an exception and SHALL be deserialized according to [Operation error deserialization](https://code.amazon.com/packages/AwsDrSeps/blobs/main/--/seps/accepted/shared/rpcv2-cbor/protocol-selection_smithy-rpc-v2.md#operation-error-deserialization).
```

This middleware validates only smithy-protocol header and continues on to deserialize the request if and only if its correctly set. I did not check other headers due to the part "No attempt should be made to interpret the response body or headers." which I assumed to mean "other" headers.

codegen/smithy-aws-swift-codegen/src/main/kotlin/software/amazon/smithy/aws/swift/codegen/protocols/rpcv2cbor/RpcV2CborProtocolGenerator.kt
dayaffe: This is a great thought but I'm not sure this pattern is very generalizable. In this case we want to ensure we're validating a header **prior to deserialization** and we don't want anything else but this one specific header to be deserialized before continuing, even to check the other headers. It also gets tricky when there may be more than one correct value for a header with custom error messages, leading to a lot of function input being generated into our clients.

codegen/smithy-aws-swift-codegen/src/main/kotlin/software/amazon/smithy/aws/swift/codegen/AWSServiceUtils.kt
sichanyoo: Hmm I see; even if the RPC v2 CBOR generator was located in smithy-swift, we would be running RPCv2 protocol tests along with other protocol tests in aws-sdk-swift side.

codegen/smithy-aws-swift-codegen/src/main/kotlin/software/amazon/smithy/aws/swift/codegen/protocols/rpcv2cbor/RpcV2CborProtocolGenerator.kt
sichanyoo: By the time `readBeforeDeserialization` interceptor hook is reached in operation flow, unless I'm mistaken, all response headers are already deserialized; i.e., headers are already deserialized into `context.getResponse().header`. I think what's not deserialized yet is the payload.

Reason for suggestion was for future extensibility; off top of my head, other place in our codebase where we validate on response headers is for flexible checksum, but I think that one should stay where it is in its own checksum middleware. 

If generic method of validating on response header is set up as part of RPCv2 CBOR, then future extension would be easy. Otherwise, someone would need to look into RPCv2 CBOR SEP to refactor existing code; more likely, just add another specialized middleware for future use case, not knowing this should probably be refactored.

It depends on how big of a work it would be though; if it's more than just quick refactoring, I'd be fine with leaving this out of this task for now.

sichanyoo: Discussed offline; and we concluded that keeping this header presence check specific to RPCv2 CBOR middleware makes sense, given checking header presence just to determine go/no-go for body deserialization is probably going to be unique to RPCv2 CBOR flow.

codegen/smithy-aws-swift-codegen/src/main/kotlin/software/amazon/smithy/aws/swift/codegen/protocols/rpcv2cbor/RpcV2CborCustomizations.kt
sichanyoo: Nvm about my previous comment, on second look pretty sure the SEP is just saying disable timestamp format customization done via `@timestampFormat` trait.

dayaffe: CBOR doesnt effect headers and timestampFormat trait is now bypassed when generating reading and writing closures: https://github.com/smithy-lang/smithy-swift/pull/887/files#diff-90d7c290bf90fafa9b697105ab1d98dda1e098df1a8db4d80c616efdb88df518R72

sichanyoo: Concluded offline that new commits to smithy-swift PR addresses the concern. Now timestamp members (regardless of whether it's top level or nested like in a list) will be serde'd using epoch seconds correctly without throwing an error

sichanyoo (approved):
LGTM; should be good to go after Josh's review as well

Package.swift
```diff
@@ -16,7 +16,7 @@ import PackageDescription
 // MARK: - Dynamic Content
 
 let clientRuntimeVersion: Version = "0.106.0"
-let crtVersion: Version = "0.40.0"
+let crtVersion: Version = "0.42.0"
```
jbelkins, line 19: FYI, we updated `main` to this CRT version today.

Sources/Core/AWSClientRuntime/Sources/AWSClientRuntime/Middlewares/RpcV2CborValidateResponseHeaderMiddleware.swift
```diff
@@ -0,0 +1,47 @@
...
+
+    public func readBeforeDeserialization(
+        context: some BeforeDeserialization<InputType, RequestType, ResponseType>
+    ) async throws {
+        let response = context.getResponse()
+        let smithyProtocolHeader = response.headers.value(for: "smithy-protocol")
+
+        guard let smithyProtocolHeader else {
+            throw ServiceResponseError.missingHeader(
+                "smithy-protocol header is missing from a response over RpcV2 Cbor!"
```
jbelkins, line 37: Do we need to provide this string at the call site?  Does `.missingHeader()` ever mean anything but this?

jbelkins: (Compare to `.badHeaderValue()` below where the error message includes a dynamically provided element)

codegen/smithy-aws-swift-codegen/src/main/kotlin/software/amazon/smithy/aws/swift/codegen/AWSServiceUtils.kt
```diff
@@ -7,24 +7,25 @@ package software.amazon.smithy.aws.swift.codegen
...
 
 /**
  * Get the [arnNamespace](https://smithy.io/2.0/aws/aws-core.html#arnnamespace)
- * from the (AWS) service shape
+ * from the (AWS) service shape or return a default value if the trait is not present.
```
jbelkins, line 21: Seems like the loss of safety from not throwing when these values are absent is minimal because I'm presuming that Smithy validators already enforce these values for services, correct?

jbelkins (approved):
Just a few Q's mostly for my info, see comments

codegen/smithy-aws-swift-codegen/src/main/kotlin/software/amazon/smithy/aws/swift/codegen/AWSServiceUtils.kt
dayaffe: An AWS service is defined by including the service trait (as compared to a non-AWS service). So all the services coming to us through Trebuchet will have it. https://smithy.io/2.0/aws/aws-core.html

Sources/Core/AWSClientRuntime/Sources/AWSClientRuntime/Middlewares/RpcV2CborValidateResponseHeaderMiddleware.swift
dayaffe: Dont need to I could move it elsewhere but thought this was the most readable / easy for debugging, do you have a preferred place to define this like implementing description in the enum

PR #1852, opened 2024-12-26
chore: Add an integration test for SigV4A signing region

sichanyoo:
## Issue \#

2238

## Description of changes

- Adds an integration test that checks that signing region for rules-based SigV4A operations (all SigV4A operations atm use rules-based auth scheme resolution) resolve signing region correctly to `*` (all SigV4A operations atm just have `*` for region field for SigV4A auth scheme property in the endpoint)

## Conventional Commits


IntegrationTests/Services/AWSS3IntegrationTests/SigV4ASigningRegionTest.swift
```diff
@@ -0,0 +1,64 @@
...
+class SigningRegionAssertInterceptor<InputType, OutputType>: Interceptor {
+    typealias RequestType = HTTPRequest
+    typealias ResponseType = HTTPResponse
+
+    func readBeforeSigning(context: some AfterSerialization<InputType, RequestType>) async throws {
+        let resolvedSigningRegion = context.getAttributes().selectedAuthScheme?.signingProperties?.get(
+            key: SigningPropertyKeys.signingRegion
+        )
+        XCTAssertNotNil(resolvedSigningRegion)
+        XCTAssertEqual(resolvedSigningRegion!, "*") // Assert that the resolved signing region is "*", as expected.
```
dayaffe, line 56: I think we either (or both) need to 1) re-write this test to use a real endpoint that resolves to 2 different regions and then make requests to both regions (as in the sev2 test code) 2) change this to a unit test since we don't actually need to make a real request to check if the resolvedSigningRegion is *

dayaffe (changes requested):

IntegrationTests/Services/AWSS3IntegrationTests/SigV4ASigningRegionTest.swift
sichanyoo: 1) There's no real endpoint that resolves to 2 different regions atm. Every endpoint that uses SigV4A uses `*` for `signingRegionSet` value.
2) I think using integration test using the actual S3 operation flow is best for testing this, given doing so actually uses the typical auth flow for rules-based auth scheme resolution. Also, since it doesn't make actual network calls, there isn't a downside either for having the test.

sichanyoo: After discussion offline that maybe let's just use SESv2's multiregion (global) endpoint to test it in real-service scenario, but scrapped due to only SESv2 API using global endpoints (`sendEmail` and `sendBulkEmail`) both require email verification process that either requires manual verification via clicking a button in the actual email account, or providing a key pair for a domain name.

sichanyoo: After offline discussion, having the test separated out like this vs. including it into S3 MRAP test is pending tie-breaker from Josh.

dayaffe (approved):
Discussed on stand, we're fine with having non-network tests in this folder

jbelkins (commented):
Approved

PR #1910, opened 2025-03-26
feat: Add PerformanceTests

dayaffe:
## Description of changes
- Add PerformanceTests to the SDK, starting with a latency test on STS. Execute using `swift run`
- `XCTest` was considered but instead this is an executable for fine grained control over the output
- Test results are printed to standard out

Example Output:
```
{
  "productId" : "AWS SDK for Swift",
  "commitId" : "597be9d3fc7286807fd7a75c2153f6374eb9ef0c",
  "results" : [
    {
      "date" : 1743017237,
      "name" : "sts.getcalleridentity.latency",
      "unit" : "Milliseconds",
      "description" : "The total time between initiating a GetCallerIdentity and reading the last byte of the object.",
      "dimensions" : [
        {
          "name" : "OS",
          "value" : "macOS"
        }
      ],
      "measurements" : [
        358.86406898498535,
        276.1859893798828,
        285.6259346008301,
        274.4629383087158,
        279.0100574493408
      ]
    }
  ],
  "sdkVersion" : "1.2.36"
}
```

## Conventional Commits


PerformanceTests/Sources/PerformanceTestScript/Util/Benchmark.swift
```diff
@@ -0,0 +1,69 @@
...
+    let dimensions: [Dimension]?
+}
+
+struct Dimension: Codable {
+    let name: String
+    let value: String
+}
+
+struct BenchmarkReport: Codable {
+    let productId: String
```
sichanyoo, line 25: Should we just fix the value of `productId` to a string constant like `AWS SDK for Swift`?

PerformanceTests/Sources/PerformanceTestScript/Util/Benchmark.swift
```diff
@@ -0,0 +1,69 @@
...
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import Foundation
+
+struct BenchmarkResult: Codable {
+    let name: String
+    let description: String
+    let unit: String
```
sichanyoo, line 13: Should we maybe add an enum for the supported unit types that map to raw string values so it's easier to maintain going forward? Atm supported cases are `Operations/Second, Bits/Second, Megabits/Second, Gigabits/Second, Seconds, Milliseconds, Microseconds, Nanoseconds, Bytes, Megabytes, Gigabytes, Percentage`

PerformanceTests/Sources/PerformanceTestScript/Util/Benchmark.swift
```diff
@@ -0,0 +1,69 @@
+//
```
sichanyoo, line 1: I think the structs here should be separated into their own standalone files. Also, instead of having `Util/` directory contain everything, might be good to separate it out into `Model/`, `Runners/`, etc. for better organization.

PerformanceTests/Package.swift
```diff
@@ -0,0 +1,19 @@
...
+
+let package = Package(
+    name: "SDKWorkbench",
+    platforms: [.macOS(.v12), .iOS(.v15)],
+    dependencies: [
+        .package(name: "aws-sdk-swift", path: "../../aws-sdk-swift"),
+    ],
+    targets: [
+        .executableTarget(
+            name: "PerformanceTestScript",
```
sichanyoo, line 13: I think `PerformanceTestRunner` might be a better module name that fits same naming pattern used for smoke tests + it's more indicative that it's a Swift executable rather than a normal script file like `test.sh`.

PerformanceTests/Sources/PerformanceTestScript/Util/Benchmark.swift
```diff
@@ -0,0 +1,69 @@
...
+// Copyright Amazon.com Inc. or its affiliates.
+// All Rights Reserved.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import Foundation
+
+struct BenchmarkResult: Codable {
```
sichanyoo, line 10: I think using `PerformanceXYZ` throughout the entire module would be easier to digest, instead of using `Performance` and `Benchmark` interchangeably . E.g., this could be named `PerformanceTestResult`.

PerformanceTests/Sources/PerformanceTestScript/PerformanceTestRun.swift
```diff
@@ -0,0 +1,38 @@
...
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import Foundation
+import ClientRuntime
+
+// MARK: - Main Entry Point
+@main
+struct PerformanceTestRun {
```
sichanyoo, line 13: This could just be named `Main` maybe

PerformanceTests/Sources/PerformanceTestScript/Util/Benchmark.swift
```diff
@@ -0,0 +1,69 @@
...
+}
+
+struct BenchmarkReport: Codable {
+    let productId: String
+    let sdkVersion: String?
+    let commitId: String
+    let results: [BenchmarkResult]
+}
+
+struct BenchmarkRunner {
```
sichanyoo, line 31: Same as above, `PerformanceTestRunner` might be more coherent name. Won't comment further on usage of "benchmark".

sichanyoo (changes requested):
Comments mostly around naming & organization.

sichanyoo (approved):

PerformanceTests/Package.swift
```diff
@@ -0,0 +1,19 @@
+// swift-tools-version: 5.9
+
+import PackageDescription
+
+let package = Package(
+    name: "SDKWorkbench",
+    platforms: [.macOS(.v12), .iOS(.v15)],
+    dependencies: [
+        .package(name: "aws-sdk-swift", path: "../../aws-sdk-swift"),
```
jbelkins, line 9: I can't recall, is `../../aws-sdk-swift` required (instead of just `..`) to make the dependency resolve?

PerformanceTests/Sources/PerformanceTestRunner/Core/Model/Dimension.swift
```diff
@@ -0,0 +1,11 @@
+//
+// Copyright Amazon.com Inc. or its affiliates.
+// All Rights Reserved.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+struct Dimension: Codable {
+    let name: String
+    let value: String
```
jbelkins, line 10: Should we maybe comment these types & properties with their description from the source doc?

PerformanceTests/Sources/PerformanceTestRunner/PerformanceMain.swift
```diff
@@ -0,0 +1,37 @@
...
+        // Add more tests here
+        let performanceTests = [
+            AWSSTSGetCallerIdentity()
+        ]
+
+        var results: [PerformanceTestResult] = []
+        for test in performanceTests {
+            let result = try await runner.runTest(test)
+            results.append(result)
+        }
```
jbelkins, line 28: nit: Map list of tests directly into list of results.
```suggestion
        let results = performanceTests.map { try await runner.runTest($0) }
```

PerformanceTests/Sources/PerformanceTestRunner/Core/Util/OperatingSystem.swift
```diff
@@ -0,0 +1,22 @@
...
+// Copyright Amazon.com Inc. or its affiliates.
+// All Rights Reserved.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+enum OperatingSystem: String {
+    case macOS = "macOS"
+    case Linux = "Linux"
+    case Unknown = "Unknown"
```
jbelkins, line 11: nit: cases of a Swift enum should start with lower case (`macOS`, `linux`, `unknown`)

PerformanceTests/Sources/PerformanceTestRunner/Core/Runners/ProcessRunner.swift
```diff
@@ -0,0 +1,43 @@
...
+    }
+
+    /// Retrieves the current Git commit ID.
+    static func getGitCommitId() -> String {
+        return ProcessRunner.runProcess(arguments: ["git", "rev-parse", "HEAD"])
+    }
+
+    /// Retrieves the SDK version from a file.
+    static func getSdkVersion() -> String {
+        return ProcessRunner.runProcess(arguments: ["cat", "../../aws-sdk-swift/Package.version"])
```
jbelkins, line 41: I think we do it elsewhere, but for safety maybe trim leading & trailing whitespace before returning version string

PerformanceTests/Sources/PerformanceTestRunner/Core/Runners/PerformanceTestRunner.swift
```diff
@@ -0,0 +1,34 @@
...
+
+    /// Executes the given test multiple times
+    func runTest(
+        _ perfTest: PerformanceTest
+    ) async throws -> PerformanceTestResult {
+        var measurements: [Double] = []
+
+        for _ in 0..<iterations {
+            measurements.append(try await perfTest.test())
+        }
```
jbelkins, line 22: See my comment elsewhere about directly mapping one list to another

PerformanceTests/Sources/PerformanceTestRunner/Core/Model/Unit.swift
```diff
@@ -0,0 +1,21 @@
+//
+// Copyright Amazon.com Inc. or its affiliates.
+// All Rights Reserved.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+enum Unit: String {
```
jbelkins, line 8: Do you think it's okay to keep units for rate, time, and byte count all in one type?

(Not sure and it looks okay to me the way `Unit` is used, just wondering if you considered it)

PerformanceTests/Sources/PerformanceTestRunner/Tests/AWSSTSTests.swift
```diff
@@ -0,0 +1,30 @@
...
+
+    let description = "The total time between initiating a GetCallerIdentity and reading the last byte of the object."
+
+    let unit = Unit.milliseconds
+
+    let dimensions = [
+        Dimension(name: "OS", value: OperatingSystem.current.rawValue),
+    ]
+
+    let test = getCallerIdentity
```
jbelkins, line 22: lol took me a second to figure out what you did here

PerformanceTests/Package.swift
jbelkins: (IIRC weird things happen if the path to a dependency doesn't end with the name of the dependency, due to SPM weirdness)

PerformanceTests/Sources/PerformanceTestRunner/Core/Runners/PerformanceTestRunner.swift
jbelkins: `let measurements = (0..<iterations).map { _ in try await perfTest.test() }`

PerformanceTests/Package.swift
dayaffe: Just tested and `..` works!

jbelkins: Might be an old issue that's now resolved, IDK

PerformanceTests/Sources/PerformanceTestRunner/Core/Model/Unit.swift
dayaffe: I think it should be ok since it should always just result in a String we pass to Catapult side of things

PerformanceTests/Sources/PerformanceTestRunner/Tests/AWSSTSTests.swift
jbelkins: I think this is a fine way to satisfy the protocol closure property BTW

PerformanceTests/Sources/PerformanceTestRunner/PerformanceMain.swift
dayaffe: oh actually I remember I tried that first but I get an error:
`Cannot pass function of type '(AWSSTSGetCallerIdentity) async throws -> PerformanceTestResult' to parameter expecting synchronous function type`

PerformanceTests/Sources/PerformanceTestRunner/Core/Model/Dimension.swift
```diff
@@ -0,0 +1,11 @@
+//
+// Copyright Amazon.com Inc. or its affiliates.
+// All Rights Reserved.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+struct Dimension: Codable {
+    let name: String  // The name of the dimension. MUST comply with CloudWatch constraints on dimension name.
+    let value: String // The value of the dimension. MUST comply with CloudWatch constraints on dimension value.
```
jbelkins, line 10: These may as well be doc comments (three slashes on line before property) so that they show in tooltip help

Also, the same sort of comments are appropriate on `PerformanceTestReport` and `PerformanceTestResult` which also form the test report structure.

jbelkins (changes requested):

jbelkins (approved):

PR #1917, opened 2025-04-14
chore: Add integration test for AWSBedrockRuntime's InvokeModelWithBidirectionalStream

sichanyoo:
## Issue \#

N/A

## Description of changes

- Adds an integration test for [AWSBedrockRuntime's InvokeModelWithBidirectionalStream API](https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_InvokeModelWithBidirectionalStream.html) that released last week.

## Conventional Commits


IntegrationTests/Services/AWSBedrockRuntimeIntegrationTests/InvokeModelWithBidirectionalStreamTest.swift
```diff
@@ -0,0 +1,233 @@
...
+        let resp = try await bedrock.invokeModelWithBidirectionalStream(input: InvokeModelWithBidirectionalStreamInput(
+            body: inputStream,
+            modelId: modelID
+        ))
+
+        // Iterate on the returned output stream for output events.
+        for try await output in resp.body! {
+            switch output {
+            case .chunk(let output):
+                XCTAssertTrue(output.bytes != nil && !output.bytes!.isEmpty)
```
jbelkins, line 227: is there anything more that we can verify for test success in the response stream, or is the fact that it contains non-zero audio data the best we can do?

jbelkins: (i.e. in the transcribe tests we check that the transcription streamed back matches what we expected from the audio we sent)

dayaffe (approved):

jbelkins (commented):
One general question, otherwise good.

IntegrationTests/Services/AWSBedrockRuntimeIntegrationTests/InvokeModelWithBidirectionalStreamTest.swift
sichanyoo: Given the length and variability of what exactly the model says in response, I think asserting on byte being non-null might be the best we can do for now.

sichanyoo:
Pending CRT investigation on why server complains about frames in requests and closes the connection prematurely.

jbelkins (approved):

PR #1921, opened 2025-04-18
chore: Run Linux integration tests on x86 & ARM

jbelkins:
## Description of changes
Run Linux integration tests on both the x86 and ARM architectures


.github/workflows/integration-test.yml
```diff
@@ -138,11 +138,14 @@ jobs:
   linux:
```
dayaffe, line 138: while we're here, why dont we run macos-14 with xcode 16.1? github's runner is macos 14.7 which is compatible

dayaffe (approved):

.github/workflows/integration-test.yml
jbelkins: (I'll have to update iOS/tvOS/visionOS sim versions etc. as well)

sichanyoo (approved):

PR #1923, opened 2025-04-22
fix!: Represent TCP port as UInt16

jbelkins:
## Description of changes

**[BREAKING] This change breaks public interfaces.  SDK minor version is updated to 1.3.0 accordingly.**

Changes the type for port number from `Int16` to `UInt16`.  TCP port may be from 0-65535 which is beyond the range of `Int16`.

In this project, this affects the interfaces for the RDS auth token generator. The rest of the changes to port type are in companion PR https://github.com/smithy-lang/smithy-swift/pull/924.

The change is breaking because port number is exposed as a property of several public types and parameters to public functions.  The change is being made nonetheless because it is a significant impediment to not allow use of all possible TCP ports.


codegen/smithy-aws-swift-codegen/src/main/kotlin/software/amazon/smithy/aws/swift/codegen/customization/rds/AuthTokenGeneratorIntegration.kt
```diff
@@ -66,7 +66,7 @@ class AuthTokenGeneratorIntegration : SwiftIntegration {
                     ///   - expiration: The expiration for the token in seconds. Default is 900 seconds (15 minutes).
                     public func generateAuthToken(
                         endpoint: String,
-                        port: Int16,
+                        port: UInt16,
```
dayaffe, line 69: Changing Int16 to UInt16 will certainly solve the issue so I am approving, but since this function isnt directly interfacing with C I am of the opinion that this should be `Int` everywhere in Swift side until the moment it needs to be another type.

dayaffe (approved):

sichanyoo (approved):

codegen/smithy-aws-swift-codegen/src/main/kotlin/software/amazon/smithy/aws/swift/codegen/customization/rds/AuthTokenGeneratorIntegration.kt
dayaffe: Found this method while googling around that can potentially take an Int and tell you if its a valid port https://developer.apple.com/documentation/kernel/in_port_t

jbelkins: Sticking with `UInt16` because it matches exactly the permitted values for TCP port.  Using `Int` would allow customers to specify invalid port values.

I'm aware that Foundation's `URLComponents` uses `Int` for port.  I don't know their rationale, probably because that is how it bridged over from the original ObjC interface, but I'd prefer to match the TCP allowed values more closely.

PR #1926, opened 2025-04-28
feat: Decompose CRT default chain into individual wrappers

sichanyoo:
## Issue \#

2380

## Description of changes

- Instead of using CRT's `.defaultChain` credential resolver, just directly use SDK-side wrappers for each individual CRT credential provider in the same order that `.defaultChain` uses. This sets the stage for moving off of CRT credential provider one by one, by replacing individual provider's internals that currently uses CRT, with SDK side implementation.

## Conventional Commits


Sources/Core/AWSSDKIdentity/Sources/AWSSDKIdentity/AWSCredentialIdentityResolvers/DefaultAWSCredentialIdentityResolverChain.swift
```diff
@@ -24,21 +26,33 @@ import protocol SmithyIdentity.AWSCredentialIdentityResolvedByCRT
...
+            { try IMDSAWSCredentialIdentityResolver() }
+        ]
+
+        let lastIndex = resolverFactories.count - 1
+        for index in 0..<lastIndex {
+            do {
+                let resolver = try resolverFactories[index]()
+                return try await resolver.getIdentity(identityProperties: identityProperties)
+            } catch {
+                // Continue to the next resolver factory.
```
dayaffe, line 50: Should we do anything with the caught errors? We could at least log them

Sources/Core/AWSSDKIdentity/Sources/AWSSDKIdentity/AWSCredentialIdentityResolvers/DefaultAWSCredentialIdentityResolverChain.swift
```diff
@@ -24,21 +26,33 @@ import protocol SmithyIdentity.AWSCredentialIdentityResolvedByCRT
...
+        }
 
-    @_spi(DefaultAWSCredentialIdentityResolverChain)
-    public init(fileBasedConfig: CRTFileBasedConfiguration) throws {
-        self.crtAWSCredentialIdentityResolver = try AwsCommonRuntimeKit.CredentialsProvider(source: .defaultChain(
-            bootstrap: SDKDefaultIO.shared.clientBootstrap,
-            fileBasedConfiguration: fileBasedConfig
-        ))
+        // The error thrown from the last resolver is not caught and instead gets thrown to caller.
+        return try await resolverFactories[lastIndex]().getIdentity(identityProperties: identityProperties)
```
dayaffe, line 55: will this always throw the IMDS error if all resolver factories fail? If so this was a huge pain from the user perspective when we used CRT so we should take this opportunity to throw a more descriptive error

dayaffe (changes requested):

Sources/Core/AWSSDKIdentity/Sources/AWSSDKIdentity/AWSCredentialIdentityResolvers/DefaultAWSCredentialIdentityResolverChain.swift
```diff
@@ -24,21 +26,33 @@ import protocol SmithyIdentity.AWSCredentialIdentityResolvedByCRT
...
+        let resolverFactories: [ResolverFactory] = [
+            { try EnvironmentAWSCredentialIdentityResolver() },
+            { try ProfileAWSCredentialIdentityResolver() },
+            { try STSWebIdentityAWSCredentialIdentityResolver() },
+            { try ECSAWSCredentialIdentityResolver() },
+            { try IMDSAWSCredentialIdentityResolver() }
+        ]
+
+        let lastIndex = resolverFactories.count - 1
+        for index in 0..<lastIndex {
```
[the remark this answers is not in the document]
sichanyoo: It was intentional, it tries to resolve from 1st to (n - 1)th resolver, then at the end of the function tries to resolve with the last resolver.

Sources/Core/AWSSDKIdentity/Sources/AWSSDKIdentity/AWSCredentialIdentityResolvers/DefaultAWSCredentialIdentityResolverChain.swift
sichanyoo: Atm that info gets logged by CRT afaik; when they're implemented in SDK side, I think individual providers would log specific issue it faced while trying to resolve credentials.

Sources/Core/AWSSDKIdentity/Sources/AWSSDKIdentity/AWSCredentialIdentityResolvers/DefaultAWSCredentialIdentityResolverChain.swift
jbelkins: I tend to agree with changing this now.  The error should give some info about the specific providers in the chain that failed.  Returning an IMDS error just because that’s the last in the chain does not make it clear what has really happened

jbelkins (changes requested):
Let's add a "chain failed" error at the same time we add providers to the chain

jbelkins (approved):

dayaffe (approved):

sichanyoo:
After team discussion, decided to add more descriptive error throwing behavior _at the end of the default chain overhaul project_. The final breaking change PR for the project will re-order credential resolvers to match the SEP as well as throw more descriptive error at the end of the chain if every resolver failed to get credentials.

PR #1928, opened 2025-05-01
Revert "feat: Decompose CRT default chain into individual wrappers"

sichanyoo:
Reverts awslabs/aws-sdk-swift#1926

STS assume role credential provider was accidentally removed from the chain. Reverting this back temporarily, will put up a new PR.


sichanyoo:
Merging revert PR before CIs pass to resolve Sev2 and Sev2.5.

jbelkins:
Reviewed after merge, approved; rolling this PR back was the correct response to the problem.

PR #1936, opened 2025-05-12
feat: Decompose default chain re-do w/ caching behavior

sichanyoo:
## Issue \#

2380

## Description of changes

- Re-do of default chain decomposition, now with credential caching.
- Confirmed via dev stack that integration tests pass in internal builds.
- Adds an integration test that confirms credential caching happens as expected.

## Conventional Commits


Sources/Core/AWSSDKIdentity/Sources/AWSSDKIdentity/AWSCredentialIdentityResolvers/DefaultAWSCredentialIdentityResolverChain.swift
```diff
@@ -24,21 +26,57 @@ import protocol SmithyIdentity.AWSCredentialIdentityResolvedByCRT
...
+        return credentials
+    }
+
+    private func shouldRefreshCredentials(expiration: Date?) -> Bool {
+        guard let expiration = expiration else {
+            return false // No expiration provided, assume no refresh is needed
+        }
+
+        let now = Date()
+        let fiveMinutesBeforeExpiration = expiration.addingTimeInterval(-300)
```
jbelkins, line 76: Does this calculation need to allow clock skew?

Sources/Core/AWSSDKIdentity/Sources/AWSSDKIdentity/AWSCredentialIdentityResolvers/DefaultAWSCredentialIdentityResolverChain.swift
```diff
@@ -24,21 +26,57 @@ import protocol SmithyIdentity.AWSCredentialIdentityResolvedByCRT
...
+
+        // The error thrown from the last resolver is not caught and instead gets thrown to caller.
+        let lastResolver = try resolverFactories[lastIndex]()
+        let credentials = try await lastResolver.getIdentity(identityProperties: identityProperties)
+        cachedCredentials = credentials
+        return credentials
+    }
+
+    private func shouldRefreshCredentials(expiration: Date?) -> Bool {
+        guard let expiration = expiration else {
```
jbelkins, line 71: nit: you can skip the assignment.
```suggestion
        guard let expiration else {
```

IntegrationTests/Services/AWSSTSIntegrationTests/DefaultAWSCredentialIdentityResolverChainTests.swift
```diff
@@ -0,0 +1,66 @@
...
+    }
+
+    func testCredentialsCaching() async throws {
+        let input = GetCallerIdentityInput()
+        for _ in 1...5 {
+            _ = try await stsClient.getCallerIdentity(input: input)
+        }
+    }
+
+    fileprivate class XCTestAssertionInterceptor<InputType, OutputType>: Interceptor {
```
jbelkins, line 34: I like this use of interceptors for test purposes.

Does the test fail when the caching is removed from the default chain?

Sources/Core/AWSSDKIdentity/Sources/AWSSDKIdentity/AWSCredentialIdentityResolvers/DefaultAWSCredentialIdentityResolverChain.swift
sichanyoo: Changed as suggested.

IntegrationTests/Services/AWSSTSIntegrationTests/DefaultAWSCredentialIdentityResolverChainTests.swift
sichanyoo: In GitHub CI it passes regardless bc we resolve credentials once and save it as environment variables. Regardless, your comment revealed a bug. Every API request gets its own copy of interceptor, so cached credential wasn't actually being compared properly. I changed the interceptor provider to instantiate interceptor once and return that for all subsequent `create` functions and I've confirmed that caching behavior on/off makes test pass/fail when using STS assume role in local.

sichanyoo: Hmm... the changed integ test doesn't work in iOS and tvOS bc using a parameterized protocol type (i.e., a protocol with associated types) in a type-erased or generic way at runtime isn't supported in older OS versions. Looking for workaround rn.

jbelkins: I hadn't thought of the state in the interceptors, glad you found it... a singleton should fix it, good call

jbelkins: There's no way to get around the force-cast to a parameterized protocol that causes the compile failure?

sichanyoo: There might be, but I figured testing it in mac and linux and confirming it behaves as expected (when run in Catapult) is adequate.

jbelkins (approved):

PR #1946, opened 2025-05-30
chore!: Remove AWSSDKIdentity dependency from AWSClientRuntime

sichanyoo:
## Issue \#

N/A

## Description of changes

- Removes AWSSDKIdentity dependency from AWSClientRuntime by deleting unnecessary method that simply constructs and returns default chain creds resolver, located in AWSClientRuntime's config defaults provider.
- This removal of dependency is a prerequisite for generating service clients for service-dependent credential resolvers in AWSSDKIdentity module without resulting in a circular dependency issue.

## Conventional Commits


Sources/Core/AWSClientRuntime/Sources/AWSClientRuntime/Plugins/DefaultAWSClientPlugin.swift
```diff
@@ -22,7 +22,6 @@ public class DefaultAWSClientPlugin: Plugin {
                 config.awsRetryMode,
                 config.maxAttempts
             )
-            config.awsCredentialIdentityResolver = try AWSClientConfigDefaultsProvider.awsCredentialIdentityResolver()
```
jbelkins, line 25: does removing this have any real effect?  We always set the `awsCredentialIdentityResolver` field directly anyway

jbelkins (commented):
Just a question

Sources/Core/AWSClientRuntime/Sources/AWSClientRuntime/Plugins/DefaultAWSClientPlugin.swift
sichanyoo: I don't think it does. For customers that use plugins, the service-specific plugins generated for each service sets the awsCredentialIdentityResolver anyways. See here for an example: https://github.com/awslabs/aws-sdk-swift/blob/5e7c7fd555ec638973233569b86ff2236b840c15/Sources/Services/AWSACM/Sources/AWSACM/Plugins.swift#L47

Note that in linked code it still uses the config defaults provider, but with changes in this PR it now just constructs default chain directly with initializer.

PR #1950, opened 2025-06-03
chore: add AWSQueryCompatible Test Cases

dayaffe:
## Conventional Commits


Sources/Core/AWSClientRuntime/Sources/AWSClientRuntime/Protocols/AWSJSON/AwsQueryCompatibleErrorDetails.swift
```diff
@@ -16,9 +16,9 @@ public class AwsQueryCompatibleErrorDetails {
         self.type = type
     }
 
-    public static func parse(_ value: String?) throws -> AwsQueryCompatibleErrorDetails {
+    public static func parse(_ value: String?) throws -> AwsQueryCompatibleErrorDetails? {
```
jbelkins, line 26: Why has the return value been made optional?

Having a throwing function return an optional value is kind of an irregular pattern

dayaffe: Because TC2 doesnt want us to throw an error if the response doesnt have query details even if the operation has the aws query compatible trait. Before this would throw an error

jbelkins: Please add a doc comment to this function that explains what a nil response value means & what a thrown error means, then I'll approve

jbelkins (approved):

PR #1980, opened 2025-07-09
feat: Checksum reuse between retries in chunked streaming requests

sichanyoo:
Companion PR:
- https://github.com/smithy-lang/smithy-swift/pull/946
## Issue \#

3350

## Conventional Commits

### DIFF
--- a/Sources/Core/AWSSDKChecksums/Sources/AWSSDKChecksums/AWSChunkedUtil.swift
+++ b/Sources/Core/AWSSDKChecksums/Sources/AWSSDKChecksums/AWSChunkedUtil.swift
@@ -13,6 +13,7 @@ import struct SmithyHTTPAPI.Headers
 import class SmithyHTTPAPI.HTTPRequestBuilder
 import enum SmithyChecksumsAPI.ChecksumAlgorithm
 import class SmithyChecksums.ChunkedStream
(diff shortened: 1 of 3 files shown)


Sources/Core/AWSSDKChecksums/Sources/AWSSDKChecksums/AWSChunkedUtil.swift
```diff
@@ -46,7 +48,8 @@ extension HTTPRequestBuilder {
                 signingConfig: signingConfig,
                 previousSignature: signature,
                 trailingHeaders: trailingHeaders,
-                checksumString: checksumString
+                checksumString: checksumString,
+                context: context
```
dayaffe, line 52: the only thing I am thinking about is whether passing the context around is bad, but it shouldnt really be too large

jbelkins: It's a reference type, it's easy to pass around.

The real concern is avoiding reference cycles since Context holds references to so many things...

jbelkins (approved):

jbelkins:
Tested in Xcode Instruments for memory leakage.  Confirmed that there are none, approving.

PR #1983, opened 2025-07-11
fix: bug with with authSchemePreference + example client codegen

dayaffe:
## Conventional Commits

### DIFF
--- a/Sources/Services/AWSBedrockRuntime/Sources/AWSBedrockRuntime/BedrockRuntimeClient.swift
+++ b/Sources/Services/AWSBedrockRuntime/Sources/AWSBedrockRuntime/BedrockRuntimeClient.swift
@@ -74,7 +74,7 @@ import typealias SmithyHTTPAuthAPI.AuthSchemes
 
 public class BedrockRuntimeClient: ClientRuntime.Client {
     public static let clientName = "BedrockRuntimeClient"
-    public static let version = "1.3.51"
+    public static let version = "1.3.52"
     let client: ClientRuntime.SdkHttpClient
     let config: BedrockRuntimeClient.BedrockRuntimeClientConfiguration
     let serviceName = "Bedrock Runtime"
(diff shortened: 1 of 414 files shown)


Sources/Services/AWSBedrockRuntime/Sources/AWSBedrockRuntime/BedrockRuntimeClient.swift
```diff
@@ -229,7 +229,7 @@ extension BedrockRuntimeClient {
                 httpClientConfiguration ?? AWSClientConfigDefaultsProvider.httpClientConfiguration(),
                 authSchemes ?? [SmithyHTTPAuth.BearerTokenAuthScheme(), AWSSDKHTTPAuth.SigV4AuthScheme()],
                 authSchemePreference ?? nil,
-                authSchemeResolver ?? DefaultBedrockRuntimeAuthSchemeResolver(),
+                authSchemeResolver ?? DefaultBedrockRuntimeAuthSchemeResolver(authSchemePreference: authSchemePreference ?? []),
```
jbelkins, line 232: Rather than pass auth scheme preference in at initialization, should we put `authSchemePreference` into Context and let auth scheme resolvers retrieve it from there?

dayaffe: Sure we could, what is the reason to prefer that method? Im open to it but wondering where your thought process is. 

I chose this approach for the cleanest diff / least amount of changes. If we plan to keep support for AuthSchemeResolver initializing authSchemePreference then this takes advantage of that being there. If we only want users to configure it on client config then I would def not want to keep this current approach.

dayaffe: Discussed offline: Will look to see if we can have this only on client config and pass via context if so.

PR #1992, opened 2025-07-23
feat: Add Bedrock API key support

jbelkins:
## Description of changes
Adds support for Bedrock API keys.  See https://docs.aws.amazon.com/bedrock/latest/userguide/api-keys.html for details about this Bedrock feature.
- Adds a `BedrockAPIKeyPlugin` that detects the API key in the process environment, then loads it into the client.
- Plugin is added to all service clients in the `bedrock` sigv4 signing domain.

### DIFF
--- a/codegen/smithy-aws-swift-codegen/src/main/kotlin/software/amazon/smithy/aws/swift/codegen/AWSClientConfigurationIntegration.kt
+++ b/codegen/smithy-aws-swift-codegen/src/main/kotlin/software/amazon/smithy/aws/swift/codegen/AWSClientConfigurationIntegration.kt
@@ -9,24 +9,32 @@ import software.amazon.smithy.aws.swift.codegen.config.AWSDefaultClientConfigura
 import software.amazon.smithy.aws.swift.codegen.config.AWSEndpointClientConfiguration
 import software.amazon.smithy.aws.swift.codegen.config.AWSRegionClientConfiguration
 import software.amazon.smithy.aws.swift.codegen.plugins.AuthSchemePlugin
+import software.amazon.smithy.aws.swift.codegen.plugins.BedrockAPIKeyPlugin
 import software.amazon.smithy.aws.swift.codegen.plugins.DefaultAWSAuthSchemePlugin
 import software.amazon.smithy.aws.swift.codegen.plugins.EndpointPlugin
+import software.amazon.smithy.aws.traits.auth.SigV4Trait
 import software.amazon.smithy.swift.codegen.config.ClientConfiguration
 import software.amazon.smithy.swift.codegen.integration.Plugin
 import software.amazon.smithy.swift.codegen.integration.ProtocolGenerator
 import software.amazon.smithy.swift.codegen.integration.ServiceConfig
 import software.amazon.smithy.swift.codegen.integration.SwiftIntegration
 import software.amazon.smithy.swift.codegen.integration.plugins.DefaultAWSClientPlugin
+import software.amazon.smithy.swift.codegen.model.getTrait
 
 class AWSClientConfigurationIntegration : SwiftIntegration {
     override fun clientConfigurations(ctx: ProtocolGenerator.GenerationContext): List<ClientConfiguration> =
         listOf(AWSDefaultClientConfiguration(), AWSRegionClientConfiguration(), AWSEndpointClientConfiguration(ctx))
 
-    override fun plugins(serviceConfig: ServiceConfig): List<Plugin> =
-        listOf(
+    override fun plugins(ctx: ProtocolGenerator.GenerationContext, serviceConfig: ServiceConfig): List<Plugin> {
+        val list = mutableListOf(
             DefaultAWSClientPlugin(),
             EndpointPlugin(serviceConfig),
             DefaultAWSAuthSchemePlugin(serviceConfig),
             AuthSchemePlugin(serviceConfig),
         )
+        if (ctx.service.getTrait<SigV4Trait>()?.name == "bedrock") {
+            list.add(BedrockAPIKeyPlugin())
+        }
+        return list
+    }
 }
--- a/codegen/smithy-aws-swift-codegen/src/main/kotlin/software/amazon/smithy/aws/swift/codegen/plugins/BedrockAPIKeyPlugin.kt
+++ b/codegen/smithy-aws-swift-codegen/src/main/kotlin/software/amazon/smithy/aws/swift/codegen/plugins/BedrockAPIKeyPlugin.kt
@@ -1,0 +1,10 @@
+package software.amazon.smithy.aws.swift.codegen.plugins
+
+import software.amazon.smithy.aws.swift.codegen.swiftmodules.AWSClientRuntimeTypes
+import software.amazon.smithy.codegen.core.Symbol
+import software.amazon.smithy.swift.codegen.integration.Plugin
+
+class BedrockAPIKeyPlugin(): Plugin {
+    override val className: Symbol = AWSClientRuntimeTypes.Customizations.BedrockAPIKeyPlugin
+    override val isDefault: Boolean = true
+}
--- a/codegen/smithy-aws-swift-codegen/src/main/kotlin/software/amazon/smithy/aws/swift/codegen/swiftmodules/AWSClientRuntimeTypes.kt
+++ b/codegen/smithy-aws-swift-codegen/src/main/kotlin/software/amazon/smithy/aws/swift/codegen/swiftmodules/AWSClientRuntimeTypes.kt
@@ -33,9 +33,8 @@ object AWSClientRuntimeTypes {
         }
     }
 
-    object RpcV2Cbor {
-        val RpcV2CborError = runtimeSymbol("RpcV2CborError", SwiftDeclaration.STRUCT, listOf("SmithyReadWrite"))
-        val CborValidateResponseHeaderMiddleware = runtimeSymbol("CborValidateResponseHeaderMiddleware", SwiftDeclaration.STRUCT)
+    object Customizations {
+        val BedrockAPIKeyPlugin = runtimeSymbol("BedrockAPIKeyPlugin", SwiftDeclaration.STRUCT)
     }
 
     object Core {
--- a/Sources/Core/AWSClientRuntime/Sources/AWSClientRuntime/Customizations/BedrockAPIKeyPlugin.swift
+++ b/Sources/Core/AWSClientRuntime/Sources/AWSClientRuntime/Customizations/BedrockAPIKeyPlugin.swift
@@ -1,0 +1,26 @@
+//
+// Copyright Amazon.com Inc. or its affiliates.
+// All Rights Reserved.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import class Foundation.ProcessInfo
+import protocol ClientRuntime.Plugin
+import protocol ClientRuntime.ClientConfiguration
+import protocol ClientRuntime.DefaultHttpClientConfiguration
+import struct SmithyIdentity.BearerTokenIdentity
+import struct SmithyIdentity.StaticBearerTokenIdentityResolver
+
+public struct BedrockAPIKeyPlugin: Plugin {
+
+    public init() {}
+
+    public func configureClient(clientConfiguration: any ClientRuntime.ClientConfiguration) async throws {
+        guard var config = clientConfiguration as? DefaultHttpClientConfiguration else { return }
+        guard let bearerToken = ProcessInfo.processInfo.environment["AWS_BEARER_TOKEN_BEDROCK"] else { return }
+        config.authSchemePreference = ["smithy.api#httpBearerAuth"] + (config.authSchemePreference ?? [])
+        let identity = BearerTokenIdentity(token: bearerToken)
+        config.bearerTokenIdentityResolver = StaticBearerTokenIdentityResolver(token: identity)
+    }
+}


Sources/Core/AWSClientRuntime/Tests/AWSClientRuntimeTests/Customizations/BedrockAPIKeyInterceptorTests.swift
```diff
@@ -0,0 +1,133 @@
...
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import XCTest
+import ClientRuntime
+import SmithyHTTPAPI
+@testable import AWSClientRuntime
+import Smithy
+import SmithyIdentity
```
jbelkins, line 14: Unit tests below verify that the interceptor makes the expected changes to the request context.  These test cases are based on the spec for the feature.

Sources/Core/AWSClientRuntime/Tests/AWSClientRuntimeTests/UserAgent/BusinessMetricsTests.swift
```diff
@@ -96,6 +96,27 @@ class BusinessMetricsTests: XCTestCase {
         XCTAssertEqual(userAgent.businessMetrics?.description, expectedString)
     }
```
jbelkins, line 98: Tests below verify that the correct business metrics get sent when the Bedrock API Key is used.

Sources/Core/AWSClientRuntime/Tests/AWSClientRuntimeTests/UserAgent/BusinessMetricsTests.swift
```diff
@@ -111,6 +132,5 @@ class BusinessMetricsTests: XCTestCase {
     private func configureContext(host: String, accountID: String) {
         let selectedAuthScheme = SelectedAuthScheme(schemeID: "aws.auth#sigv4", identity: AWSCredentialIdentity(accessKey: "abc", secret: "def", accountID: accountID), signingProperties: Attributes(), signer: nil)
         context.selectedAuthScheme = selectedAuthScheme
-        let uri = URIBuilder().withScheme(.https).withPath("/").withHost(host).build()
```
jbelkins, line 114: Unused line of code

codegen/smithy-aws-swift-codegen/src/main/kotlin/software/amazon/smithy/aws/swift/codegen/AWSClientConfigurationIntegration.kt
```diff
@@ -9,24 +9,36 @@ import software.amazon.smithy.aws.swift.codegen.config.AWSDefaultClientConfigura
...
+        val list =
+            mutableListOf(
+                DefaultAWSClientPlugin(),
+                EndpointPlugin(serviceConfig),
+                DefaultAWSAuthSchemePlugin(serviceConfig),
+                AuthSchemePlugin(serviceConfig),
+            )
+        if (ctx.service.getTrait<SigV4Trait>()?.name == "bedrock") {
+            list.add(BedrockAPIKeyPlugin())
+        }
```
jbelkins, line 41: The Bedrock plugin is added only for the appropriate services, i.e. ones with `bedrock` as the sigv4 signing name.

codegen/smithy-aws-swift-codegen/src/main/kotlin/software/amazon/smithy/aws/swift/codegen/plugins/BedrockAPIKeyPlugin.kt
```diff
@@ -0,0 +1,10 @@
+package software.amazon.smithy.aws.swift.codegen.plugins
+
+import software.amazon.smithy.aws.swift.codegen.swiftmodules.AWSClientRuntimeTypes
+import software.amazon.smithy.codegen.core.Symbol
+import software.amazon.smithy.swift.codegen.integration.Plugin
+
+class BedrockAPIKeyPlugin : Plugin {
+    override val className: Symbol = AWSClientRuntimeTypes.Customizations.BedrockAPIKeyPlugin
+    override val isDefault: Boolean = true
+}
```
jbelkins, line 10: Simple `Plugin` implementation that allows rendering a `BedrockAPIKeyPlugin` into Swift code.

codegen/smithy-aws-swift-codegen/src/main/kotlin/software/amazon/smithy/aws/swift/codegen/swiftmodules/AWSClientRuntimeTypes.kt
```diff
@@ -33,9 +33,8 @@ object AWSClientRuntimeTypes {
         }
     }
 
-    object RpcV2Cbor {
-        val RpcV2CborError = runtimeSymbol("RpcV2CborError", SwiftDeclaration.STRUCT, listOf("SmithyReadWrite"))
-        val CborValidateResponseHeaderMiddleware = runtimeSymbol("CborValidateResponseHeaderMiddleware", SwiftDeclaration.STRUCT)
+    object Customizations {
+        val BedrockAPIKeyPlugin = runtimeSymbol("BedrockAPIKeyPlugin", SwiftDeclaration.STRUCT)
```
jbelkins, line 37: Symbol for the new plugin type.  Deleted two unused symbols.

codegen/smithy-aws-swift-codegen/src/main/kotlin/software/amazon/smithy/aws/swift/codegen/AWSHTTPBindingProtocolGenerator.kt
```diff
@@ -81,6 +83,15 @@ abstract class AWSHTTPBindingProtocolGenerator(
...
     }
 
+    override fun addCustomizationMiddleware(
+        ctx: ProtocolGenerator.GenerationContext,
+        operation: OperationShape,
+    ) {
+        if (ctx.service.getTrait<SigV4Trait>()?.name == "bedrock") {
+            operationMiddleware.appendMiddleware(operation, BedrockAPIKeyMiddleware())
+        }
+    }
```
jbelkins, line 95: This customization point was added in https://github.com/smithy-lang/smithy-swift/pull/951.

Interceptor is installed in the stack when the signing name is `bedrock`, this is per the spec

IntegrationTests/Services/AWSBedrockRuntimeIntegrationTests/BedrockAPIKeyGenerator.swift
```diff
@@ -0,0 +1,106 @@
...
+import struct Foundation.TimeInterval
+import protocol SmithyIdentity.AWSCredentialIdentityResolver
+import protocol SmithyIdentityAPI.IdentityResolver
+import class AWSSDKIdentity.DefaultAWSCredentialIdentityResolverChain
+import class SmithyHTTPAPI.HTTPRequestBuilder
+import struct SmithyHTTPAPI.Headers
+import struct Smithy.URIQueryItem
+import struct Smithy.Attributes
+import enum SmithyHTTPAuthAPI.SigningPropertyKeys
+import class SmithyHTTPAuth.SigV4Signer
```
jbelkins, line 18: Below is a generator for Bedrock auth tokens.  This is used in integration tests to generate a token for use in the test.

A token is basically a presigned URL with some extra text processing applied to it, as you can see below.

This type is based on the Python token generator package (which despite being a package, is just a few lines of code that use Boto3 to construct the token.)
https://github.com/aws/aws-bedrock-token-generator-python/blob/main/aws_bedrock_token_generator/token_generator.py

IntegrationTests/AWSIntegrationTestUtils/BearerTokenWasUsedInterceptor.swift
```diff
@@ -0,0 +1,40 @@
...
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import XCTest
+import protocol ClientRuntime.AfterSerialization
+import protocol ClientRuntime.HttpInterceptorProvider
+import protocol ClientRuntime.Interceptor
+import class SmithyHTTPAPI.HTTPRequest
+import class SmithyHTTPAPI.HTTPResponse
+
```
jbelkins, line 14: Below is a simple interceptor that will fail a test if the auth used for the request is not bearer token auth.

IntegrationTests/Services/AWSBedrockRuntimeIntegrationTests/BedrockAPIKeyTests.swift
```diff
@@ -0,0 +1,36 @@
...
+
+import XCTest
+import AWSIntegrationTestUtils
+import AWSBedrockRuntime
+
+final class BedrockAPIKeyIntegrationTests: XCTestCase {
+    let region = "us-west-2"
+    let envVarName = "AWS_BEARER_TOKEN_BEDROCK"
+    let apiKeyDuration: TimeInterval = 600.0
+
```
jbelkins, line 16: Below is an integration test that:
- creates a Bedrock API token
- sets the token in the expected environment var
- calls a live Bedrock operation
- verifies that the request succeeded
- verifies that bearer token auth was indeed used

IntegrationTests/Services/AWSBedrockRuntimeIntegrationTests/BedrockAPIKeyGenerator.swift
```diff
@@ -0,0 +1,106 @@
...
+        CommonRuntimeKit.initialize()
+        let signedRequestBuilder = try await SigV4Signer().signRequest(
+            requestBuilder: inputBuilder,
+            identity: awsCredentialIdentity,
+            signingProperties: signingProperties
+        )
+
+        // Create a presigned URL
+        let signedRequest = signedRequestBuilder.build()
+        let url = signedRequest.endpoint.uri.url!
```
jbelkins, line 96: note: `!` won't ever fail here because the essential URL components were set on the `inputBuilder` above

IntegrationTests/AWSIntegrationTestUtils/BearerTokenWasUsedInterceptor.swift
```diff
@@ -0,0 +1,40 @@
...
+    public typealias RequestType = HTTPRequest
+    public typealias ResponseType = HTTPResponse
+
+    public func readAfterSigning(context: some AfterSerialization<InputType, HTTPRequest>) async throws {
+        let request = context.getRequest()
+        guard let authHeader = request.headers.value(for: "Authorization") else {
+            XCTFail("Authorization header was not present in signed response")
+            return
+        }
+        guard authHeader.hasPrefix("Bearer ") else {
```
dayaffe, line 35: Not a blocking comment but just wondering if there might be a better way to check this such as splitting on space. Checking for literal prefix "Bearer " with a space does seem like it would work but isnt so robust

jbelkins: https://datatracker.ietf.org/doc/html/rfc6750#section-2.1

"Bearer" plus one space is in the IETF standard.

codegen/smithy-aws-swift-codegen/src/main/kotlin/software/amazon/smithy/aws/swift/codegen/swiftmodules/AWSClientRuntimeTypes.kt
```diff
@@ -33,9 +33,8 @@ object AWSClientRuntimeTypes {
         }
     }
 
-    object RpcV2Cbor {
-        val RpcV2CborError = runtimeSymbol("RpcV2CborError", SwiftDeclaration.STRUCT, listOf("SmithyReadWrite"))
-        val CborValidateResponseHeaderMiddleware = runtimeSymbol("CborValidateResponseHeaderMiddleware", SwiftDeclaration.STRUCT)
+    object Customizations {
+        val BedrockAPIKeyInterceptor = runtimeSymbol("BedrockAPIKeyInterceptor", SwiftDeclaration.STRUCT)
```
jbelkins, line 37: The CBOr-related symbols above were unused

codegen/smithy-aws-swift-codegen/src/main/kotlin/software/amazon/smithy/aws/swift/codegen/AWSHTTPBindingProtocolGenerator.kt
jbelkins: Additional condition added late: the signing name must be `bedrock` AND the service must have the `httpBearerAuth` trait for the Bedrock API Keys customization to apply

Sources/Core/AWSClientRuntime/Sources/AWSClientRuntime/Customizations/BedrockAPIKeyInterceptor.swift
```diff
@@ -0,0 +1,63 @@
...
+        let attributes = context.getAttributes()
+
+        // Check if identityResolvers already contains a custom bearer token resolver,
+        // i.e. one that is not the default bearer token identity resolver chain.
+        // If so, return immediately & use that instead of the Bedrock API token.
+        let identityResolvers = attributes.getIdentityResolvers() ?? Attributes()
+        let key = AttributeKey<any BearerTokenIdentityResolver>(name: "smithy.api#httpBearerAuth")
+        guard !identityResolvers.contains(key: key) || identityResolvers.get(key: key) is
+                DefaultBearerTokenIdentityResolverChain else {
+            print("Custom bearer token resolver")
```
sichanyoo: Q: What's the point of the second condition `identityResolvers.get(key: key) is DefaultBearerTokenIdentityResolverChain`? Isn't checking whether custom bearer token provider is configured or not enough (the first condition)? Also, should the `print` statement here instead be a debug log sent to an instance of `SwiftLogger`?

jbelkins: The spec says that the Bedrock API token should not be used if the customer sets a custom bearer token identity resolver, because in that case the customer's token should be resolved instead.

However , the bearer token identity resolver is never unset because if none was provided, we set an instance of `DefaultBearerTokenIdentityResolverChain` as the bearer token identity resolver.  So to tell if no custom bearer token identity resolver was provided, I have to check either (1) that there really is none, or (2) that there is one but it was just the default one that we automatically set (`DefaultBearerTokenIdentityResolverChain`).  Anything other than those two options is presumed to have been provided by the customer and is used instead.

The print statement is left over from debugging and I'll remove it.

sichanyoo (approved):

PR #1999, opened 2025-08-06
chore: Add Swiftlint analyzer to lint CI job

jbelkins:
## Description of changes
Performs Swiftlint `analyze` to run extra rules (in the `analyzer_rules` section of `.swiftlint.yml`).  Swiftlint analyzer can do deeper inspection of code since it relies on an AST built by the compiler.

Specific changes:
- Normal Swiftlint still runs the way it does now, then `analyze` runs next.  Lint will "fail fast" if there are issues, and `analyze` will not run.
- Moves Swiftlint from Linux to Mac since `analyze` requires building the SDK with Xcode before analysis.
- Deletes AWS service clients from the project so only runtime remains, then builds the project with Xcode.
- Many analyzer lint violations have been fixed in this PR.

The entire swiftlint routine seems to take 7-8 minutes, so won't hold up the completion of CI.


sichanyoo (approved):
Nice !

PR #2043, opened 2025-10-31
fix: Use full UUIDs for random identifiers in integration tests

jbelkins:
## Description of changes
Recently it was noted that there were occasional name collisions between S3 buckets when the buckets were named using the first 8 digits of a UUID for randomness.  Switching to full UUIDs appears to have eliminated the problem.

This PR takes every other place that something less than a full UUID was used in a random name, and converts it to a full UUID.
- A helper method was created for generating random names that are easily identifiable to a specific test.
- One test was refactored to move its cleanup steps into a `tearDown` method.

### DIFF
--- a/IntegrationTests/Services/AWSS3IntegrationTests/S3ConcurrentTests.swift
+++ b/IntegrationTests/Services/AWSS3IntegrationTests/S3ConcurrentTests.swift
@@ -41,18 +41,18 @@ final class S3ConcurrentTests: S3XCTestCase, @unchecked Sendable {
     // Puts data to S3, gets the uploaded file, asserts retrieved data == original data, deletes S3 object
     @Sendable
     private func getObject(data: Data) async throws {
-       let objectKey = UUID().uuidString.split(separator: "-").first!.lowercased()
-       let putObjectInput = PutObjectInput(body: .data(data), bucket: bucketName, key: objectKey)
+        let objectKey = String.uniqueID(service: "s3")
+        let putObjectInput = PutObjectInput(body: .data(data), bucket: bucketName, key: objectKey)
 
-       _ = try await client.putObject(input: putObjectInput)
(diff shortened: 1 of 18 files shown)


IntegrationTests/Services/AWSS3IntegrationTests/S3ConcurrentTests.swift
```diff
@@ -41,18 +41,18 @@ final class S3ConcurrentTests: S3XCTestCase, @unchecked Sendable {
...
+            bucket: bucketName, key: objectKey
+        )).body?.readData()
 
-       XCTAssertEqual(data, retrievedData)
+        XCTAssertEqual(data, retrievedData)
 
-       let deleteObjectInput = DeleteObjectInput(bucket: bucketName, key: objectKey)
-       _ = try await client.deleteObject(input: deleteObjectInput)
+        let deleteObjectInput = DeleteObjectInput(bucket: bucketName, key: objectKey)
+        _ = try await client.deleteObject(input: deleteObjectInput)
```
jbelkins, line 56: Fixed formatting above

IntegrationTests/Services/AWSS3IntegrationTests/S3ExpressIntegrationTests.swift
```diff
@@ -48,15 +61,5 @@ final class S3ExpressIntegrationTests: S3ExpressXCTestCase {
...
-
-        // Delete the object from each bucket
-        for bucket in buckets {
-            try await deleteObject(bucket: bucket, key: key)
-        }
-
-        // Delete each directory bucket
-        for bucket in buckets {
-            try await deleteBucket(bucket: bucket)
-        }
```
jbelkins, line 60: Moved this test cleanup content from the test body to the `tearDown()` override above, to ensure it gets run after the test.

PR #2054, opened 2025-11-10
feat: add performance test for NIO HTTP Client against sts

dayaffe:
## Conventional Commits

### DIFF
--- a/PerformanceTests/Sources/PerformanceTestRunner/Tests/AWSSTSTests.swift
+++ b/PerformanceTests/Sources/PerformanceTestRunner/Tests/AWSSTSTests.swift
@@ -7,6 +7,7 @@
 
 import AWSSTS
 import Foundation
+import ClientRuntime
 
 struct AWSSTSGetCallerIdentity: PerformanceTest {
     let name = "sts.getcalleridentity.latency"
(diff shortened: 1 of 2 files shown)


PerformanceTests/Sources/PerformanceTestRunner/Tests/AWSSTSTests.swift
```diff
@@ -7,6 +7,7 @@
 
 import AWSSTS
 import Foundation
+import ClientRuntime
 
 struct AWSSTSGetCallerIdentity: PerformanceTest {
```
jbelkins, line 12: What client is this test testing?  Just whatever happens to be the default client?

Wondering if we should set up tests for all three, and selectively compile only the tests for supported clients on the current platform.

jbelkins (approved):

PerformanceTests/Sources/PerformanceTestRunner/Tests/AWSSTSTests.swift
dayaffe: Will look into setting this up. Right now I think we only run performance tests on codebuild which right now only supports LInux. So this would effectively test CRT + NIO but not URLSession

PR #2066, opened 2025-11-20
fix: API ref landing page

sichanyoo:
## Issue \#

N/A

## Description of changes

- Swift requires header of the index page to be the same as the actual module name for it to be rendered as the landing page for the module.
- Adds DisplayName metadata tag to doc index page so the landing page title is more descriptive for both old and new API reference docs.

## Conventional Commits

### DIFF
--- a/AWSSDKSwiftCLI/Sources/AWSSDKSwiftCLI/Resources/DocIndex.Base.md
+++ b/AWSSDKSwiftCLI/Sources/AWSSDKSwiftCLI/Resources/DocIndex.Base.md
@@ -1,5 +1,9 @@
 # ``AWSSDKForSwift``
 
+@Metadata {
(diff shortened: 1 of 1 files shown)


Sources/Core/SDKForSwift/Documentation.docc/SDKForSwift.md
```diff
@@ -1,4 +1,4 @@
-# ``AWSSDKForSwift``
+# ``SDKForSwift``
```
dayaffe, line 1: https://www.swift.org/documentation/docc/displayname can we use this to show the right name?

sichanyoo: Yep that worked flawlessly. Tested via building documentation in Xcode and confirmed it behaves as expected when using DisplayName metadata tag.

PR #2074, opened 2025-12-02
fix: Remove S3/200 cust from S3 Select op, re-enable int test

jbelkins:
## Description of changes
- Removes the S3/200 customization from the S3 `SelectObjectContent` operation.  Per the spec it is not to be applied to event streaming operations, and it was causing S3 Select to fail because it was attempting to parse the entire event stream as XML.
- Re-enables the S3 `SelectObjectContent` integration test.
- The code-generated change to the S3 client (removing the customization from `SelectObjectContent`) is included in the PR.

### DIFF
--- a/IntegrationTests/Services/AWSS3IntegrationTests/S3EventStreamTests.swift
+++ b/IntegrationTests/Services/AWSS3IntegrationTests/S3EventStreamTests.swift
@@ -26,7 +26,7 @@ class S3EventStreamTests: S3XCTestCase {
     }
 
     // Tests event stream output in restXml protocol using S3::SelectObjectContent.
-    func xtestEventStreamOutput() async throws {
+    func testEventStreamOutput() async throws {
         let result = try await client.selectObjectContent(input: SelectObjectContentInput(
             bucket: bucketName,
             // Gets the two ID objects from the S3 object content.
--- a/Sources/Services/AWSS3/Sources/AWSS3/S3Client.swift
+++ b/Sources/Services/AWSS3/Sources/AWSS3/S3Client.swift
@@ -9101,7 +9101,6 @@ extension S3Client {
         context.set(key: Smithy.AttributeKey<EndpointParams>(name: "EndpointParams"), value: endpointParamsBlock(context))
         builder.applyEndpoint(AWSClientRuntime.AWSEndpointResolverMiddleware<SelectObjectContentOutput, EndpointParams>(paramsBlock: endpointParamsBlock, resolverBlock: { [config] in try config.endpointResolver.resolve(params: $0) }))
         builder.selectAuthScheme(ClientRuntime.AuthSchemeMiddleware<SelectObjectContentOutput>())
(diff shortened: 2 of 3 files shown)


codegen/smithy-aws-swift-codegen/src/main/kotlin/software/amazon/smithy/aws/swift/codegen/customization/s3/S3ErrorWith200StatusIntegration.kt
```diff
@@ -40,15 +40,9 @@ class S3ErrorWith200StatusIntegration : SwiftIntegration {
...
         // which adds a small amount of overhead to response processing.
         val output = ctx.model.expectShape(operationShape.output.get())
-        val outputIsNotAStreamingBlobShape =
-            output.members().none {
-                val targetShape = ctx.model.expectShape(it.target)
-                val isBlob = it.isBlobShape || targetShape.isBlobShape
-                val isStreaming = it.isStreaming || targetShape.isStreaming
-                isBlob && isStreaming
-            }
+        val outputIsNotAStreamingShape = output.members().none { ctx.model.expectShape(it.target).isStreaming }
```
jbelkins, line 43: In the logic above:
- We no longer check if the targeted shape is a blob, since we want to keep the customization off of both data _and_ event streams.
- No need to check for the trait on the member shape, since `@streaming` is never allowed to be applied there, only on the target shape.

IntegrationTests/Services/AWSS3IntegrationTests/S3EventStreamTests.swift
```diff
@@ -26,7 +26,7 @@ class S3EventStreamTests: S3XCTestCase {
     }
 
     // Tests event stream output in restXml protocol using S3::SelectObjectContent.
-    func xtestEventStreamOutput() async throws {
+    func testEventStreamOutput() async throws {
```
jbelkins, line 29: Test re-enabled.

Sources/Services/AWSS3/Sources/AWSS3/S3Client.swift
```diff
@@ -9101,7 +9101,6 @@ extension S3Client {
         context.set(key: Smithy.AttributeKey<EndpointParams>(name: "EndpointParams"), value: endpointParamsBlock(context))
         builder.applyEndpoint(AWSClientRuntime.AWSEndpointResolverMiddleware<SelectObjectContentOutput, EndpointParams>(paramsBlock: endpointParamsBlock, resolverBlock: { [config] in try config.endpointResolver.resolve(params: $0) }))
         builder.selectAuthScheme(ClientRuntime.AuthSchemeMiddleware<SelectObjectContentOutput>())
-        builder.interceptors.add(AWSClientRuntime.AWSS3ErrorWith200StatusXMLMiddleware<SelectObjectContentInput, SelectObjectContentOutput>())
```
jbelkins, line 9104: This is the body of `SelectObjectContent`, where the customization is removed.

PR #2095, opened 2026-01-26
Feat: Adding a script to run SwiftLint locally

AntAmazonian:
## Issue \#

N/A

## Description of changes

The Swiftlint analyzer is configured to run on aws-sdk-swift CI. The analyzer provides better feedback than the main linter, but it also requires setup, including building the project prior to analysis.

A script is provided that developers can run on desktop to perform the same Swiftlint analysis that CI performs.

N/A

## Conventional Commits

"Add SwiftLint config and local execution script"


scripts/lint.sh
```diff
@@ -0,0 +1,23 @@
...
+    echo "Error: SwiftLint not found. Install it using 'brew install swiftlint'."
+    exit 1
+fi
+
+echo "Running SwiftLint..."
+
+SCRIPT_DIR="$(dirname "$0")"
+CONFIG_PATH="$SCRIPT_DIR/../.swiftlint.yml"
+
+swiftlint lint --config "$CONFIG_PATH" --strict
```
sichanyoo: This just runs lint, not analyzer.

sichanyoo (changes requested):
Script in this PR only runs lint, not analyzer.

scripts/lint.sh
AntAmazonian: Script now builds logger, runs analyzer, and lint

sichanyoo (changes requested):
Is there a particular reason SPM build log was used instead of Xcode build log like in CI? I think the main point of having a script for this is so that we can easily run the analyzer lint the same way it is done in CI, but this technically diverges from it.

Have you run this lint yourself? Not sure if it would take forever to run because it tries to build every service in the project. See https://github.com/awslabs/aws-sdk-swift/blob/4c3e4ae40492b2560740df79ea0ddc306c52dffd/.github/workflows/lint.yml#L44-L60 for how generated SDK is removed before running analyzer.

scripts/lint.sh
```diff
@@ -0,0 +1,53 @@
+#!/bin/bash
+#
+#  lint.sh
+#  aws-sdk-swift
+#
+#  Created by Felix, Anthony on 1/23/26.
+#
```
jbelkins: nit: Replace the header with the standard Amazon header. Example:
https://github.com/awslabs/aws-sdk-swift/blob/main/Sources/Core/AWSClientRuntime/Sources/AWSClientRuntime/AWSClientConfigDefaultsProvider.swift#L1-L6

Alternatively, you can simply delete the header entirely.

scripts/lint.sh
```diff
@@ -0,0 +1,53 @@
...
+cd AWSSDKSwiftCLI
+swift run AWSSDKSwiftCLI generate-package-manifest ..
+cd ..
+
+echo "Step 3: Building minimal project for analyzer..."
+if xcodebuild -scheme aws-sdk-swift-Package -destination platform=macOS > "$LOG_FILE" 2>&1; then
+    echo "✅ Build completed successfully"
+    
+    echo "Step 4: Running SwiftLint Analyze (this may take a few minutes)..."
+    timeout 300 swiftlint analyze --compiler-log-path "$LOG_FILE" || echo "⚠️  SwiftLint Analyze timed out after 5 minutes"
```
jbelkins: What is the reason for the 5 minute timeout? I generally disfavor setting timeouts unless they're there to address an operational problem, which I don't think we have here.

scripts/lint.sh
```diff
@@ -0,0 +1,53 @@
...
+#  aws-sdk-swift
+#
+#  Created by Felix, Anthony on 1/23/26.
+#
+
+# Standard brew path setup
+export PATH="$PATH:/opt/homebrew/bin:/usr/local/bin"
+
+# Unset AWS_SWIFT_SDK_USE_LOCAL_DEPS to use remote dependencies instead of local ones
+unset AWS_SWIFT_SDK_USE_LOCAL_DEPS
```
jbelkins: I don't think we want to do this for a locally-run development script, since SDK developers are typically using a local copy of smithy-swift with local SDK.  If you've made local SDK changes that have corresponding smithy-swift changes, the `xcodebuild` initiated below will fail.

.swiftlint.yml
```diff
@@ -21,17 +21,15 @@ disabled_rules:
   - cyclomatic_complexity
   - file_length
   - syntactic_sugar
```
jbelkins, line 23: Why the specific changes to linter rules in this file?

Also, for simplicity, we typically keep the same linter rules in SDK and in smithy-swift.  So if we do change rules here, there should be a 2nd PR to change them there as well.

jbelkins (changes requested):
See individual comments

scripts/lint.sh
AntAmazonian: Okay, I didn't think about developers making their own changes if they used a copy of the SDK locally, I'll change it so leaves the deps alone if they were set by the developer. That way it's a choice on their end.

scripts/lint.sh
AntAmazonian: The timeout I added as a just in case for local runs by the developer, but thinking twice about it now I think this should be more on the developer to set if they want to.

PR #2096, opened 2026-01-28
chore: Move test teardown for S3 Express int tests to a teardown method

jbelkins:
## Description of changes

Move test teardown to a `tearDown()` method to ensure that it gets run in all cases after the test.

Also:
- Refactored common S3 Express test logic into `S3ExpressTestCase` so it can be shared among tests
- Made bucket creation happen in parallel for `S3ExpressIntegrationTest` to reduce the time required for the test


dayaffe (approved):
LGTM assuming this was tested and both before and after the test there is no change in the number of buckets regardless of whether the test passes

PR #2146, opened 2026-04-15
chore: Improved retry, logging on EventBridge int tests

jbelkins:
## Description of changes
The EventBridge integration test is a frequent cause of spurious failures, due to not waiting long enough for the global endpoint created in test setup to resolve successfully in DNS.  To correct these problems:
- A retry loop (3x with 2, 4, 8 sec waits) is established around the `PutEvents` call, which is the one that uses the custom host created in setup.
- In several places where retry is performed, there is now logging to make clear when retry was exhausted.


IntegrationTests/Services/AWSEventBridgeIntegrationTests/EventBridgeSigV4ATests.swift
```diff
@@ -124,12 +128,28 @@ class EventBridgeSigV4ATests: XCTestCase {
...
+        var output: PutEventsOutput?
+        for attempt in 0..<4 {
+            do {
+                output = try await sigv4aEventBridgeClient.putEvents(input: input)
+                break
+            } catch {
+                print("Attempt \(attempt) failed: \(error)")
+                // Retry delays are 4, 8, 16, 32 secs for 1st, 2nd, 3rd, 4th retry respectively
+                let delay = UInt64(4 << attempt)
+                try await Task.sleep(nanoseconds: delay * UInt64(NSEC_PER_SEC))
```
dayaffe, line 146: nit: here and on 103 it looks like we sleep after the last attempt, ideally we shouldn't do that since it can add in this case 32 seconds in each instance we do it necessarily

IntegrationTests/Services/AWSEventBridgeIntegrationTests/EventBridgeSigV4ATests.swift
```diff
@@ -124,12 +128,28 @@ class EventBridgeSigV4ATests: XCTestCase {
...
+        // The host used by this call is a custom host, which sometimes takes a few seconds
+        // after creation for DNS to resolve it successfully.
+        let input = PutEventsInput(endpointId: unwrappedEndpointId, entries: [event])
+        var output: PutEventsOutput?
+        for attempt in 0..<4 {
+            do {
+                output = try await sigv4aEventBridgeClient.putEvents(input: input)
+                break
+            } catch {
+                print("Attempt \(attempt) failed: \(error)")
```
dayaffe, line 143: question: do we just want to print or should we keep track of the error and throw it?

dayaffe (approved):
1 question and 1 nit -- otherwise LGTM

PR #2173, opened 2026-05-26
fix: Remove flaky EventBridge SigV4A integration test

dayaffe:
This test is functionally a duplicate of existing test `testS3MRAPSigV4A`:
  1. Build a service client with authSchemes = [SigV4AAuthScheme()] (SigV4 stripped out, only SigV4A left)
  2. Make a real call to AWS
  3. Assert it succeeds
  
 Removing since its known to be very spotty.

## Conventional Commits


jbelkins (changes requested):
Remove EventBridge from here too:
https://github.com/awslabs/aws-sdk-swift/blob/main/scripts/integration-test-sdk.properties

(this controls what gets code generated for integration tests)

jbelkins (approved):

